PR #26, opened 2015-07-20
instanceof not working with strings.

SeoFood:
Hi

instanceof is not working when the class name is just a string. It will be always false.


freekmurze:
Thanks!

freekmurze:
This fix is included in the newly tagged version 2.1.5

PR #36, opened 2015-07-30
resolve #34, unset glideImage object to save memory on regenerate or bulkinsert

### LINKED ISSUE #34: php artisan medialibrary:regenerate runs out of memory
while regenerating many images it never releases memory so it runs out.
please release used memory after each images.

pamo85:


freekmurze:
Have you tested this yourself? In my tests this change doesn't solve the problem. There's still a memory leak.

pamo85:
yes i tested it. it solves memory leak from ram allocated by image object in glide. there's still a memory leak left somewhere. but regenerating 300 images only leaks about 10mb instead of 3gb memory.

freekmurze:
Mmm, on my machine the leak remained even after your fix. I

I've updated the underly Glide package to resolve the memory issue. Could you, in your project, run a composer update to get the latest version of the medialibrary and version 2.2.5 of spatie/laravel-glide (that one has been released a few minutes ago). Run your tests again and let me know if it is fixed.

pamo85:
i tried it. calling the gc does the job. even better than my solution.

freekmurze:
Nice! Thanks for testing!

PR #50, opened 2015-08-23
Add custom directory feature 

### LINKED ISSUE #45: Question: Any way to preserve or specifiy folder structure?
Hello,

Right now the folder structure is the path we specifiy in `media` drive plus the `id` of media object plus the  filename. Ex: `/public/images/1/test.jpg`

I want to modify the folder structure because i want to organize the folder. Now everything is inside one folder. For example i want to have path like: `/public/images/reports/daily/1.jpg` 

`/public/images` part comes from the filesystem config file. We have to specifiy `reports/daily` part, maybe with `toFolder` method. Or with `preserveFolderStructure` we can keep the original structure.

We can discuss this and i can make a PR if you think this is a good idea.

canvural:
closes #45

### DIFF
--- a/src/FileAdder/FileAdder.php
+++ b/src/FileAdder/FileAdder.php
@@ -269,6 +269,8 @@ class FileAdder
 
         $media->save();
 
+        $media->setDirectory($this->subject->getDirectory());
+
         $this->subject->media()->save($media);
         $this->fileSystem->add($this->pathToFile, $media, $this->fileName);
 
--- a/src/Filesystem.php
+++ b/src/Filesystem.php
@@ -113,7 +113,7 @@ class Filesystem
     {
         $this->filesystems->disk($media->disk)->put('.gitignore', Gitignore::getContents());
 
-        $directory = $media->id;
+        $directory = $media->directory;
         $this->filesystems->disk($media->disk)->makeDirectory($directory);
 
         return $directory;
--- a/src/HasMedia/HasMediaTrait.php
+++ b/src/HasMedia/HasMediaTrait.php
@@ -252,9 +252,22 @@ trait HasMediaTrait
         }
 
         $this->media()->get()->map(function (Media $media) {
+            $media->setDirectory($this->getDirectory());
             $media->delete();
         });
 
         return true;
+    }
+
+    /**
+     * This method is designed to be overridden by the model.
+     * It should return the path of the directory to be used for the media
+     * of this model. By default it will use the id of the media.
+     *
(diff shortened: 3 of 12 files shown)


freekmurze:
Thanks for all this work. I greatly appreciate that you added some tests. At first glance it technically seems ok. In the coming weeks I'll review it more thoroughly at let you know my thoughts on it.

freekmurze:
Hi, I'll be reviewing this PR soon. Could you rebase it to the current master branch?

canvural:
I think i rebased :) Gonna fix the CS issues.

sebastiandedeyne:
Sidenote: you can ignore the camel case warnings in in the tests.

freekmurze:
@canvural great thanks!

canvural:
@freekmurze All good now? 

I can't see 2 newline at `tests/TestModelWithCustomDirectory.php`. I don't understand that :)

freekmurze:
I've got two remarks on this PR:

1. The `getDirectory`-function in the trait is kinda weird. It's  not that intuitive. Specify the custom directory via the model also isn't very flexible. I think ik would be better to define a method to specify the directory on the `FileAdder` class. Such a method could be named `usingDirectory` (maybe you could come up with something better). In that way a custom directory can be specified per file.

2. As it stands now there is no protection against a media object saving it's files inside the directory of another media object. This could result in data loss when a media object with the parent folder gets deleted. I don't think this can be easily solved. I'm thinking of added a big warning in the documentation that this could happend. It's up to the user of the package to don't make any mistakes.

What are your thoughts on this?

canvural:
1) I think it like this. "Get the directory of this model". 
 And one problem with adding the method to `FileAdder` class is, when we are fetching media objects from database, we can't know its directory. In this PR i'm getting the directory from model, while fetching from database.

2) It still uses `id`s for the folder name. So unless the user gives a number for the custom directory name, which is also the id of another media, it's fine. Maybe we can fix it by restricting numbers in the custom directory.

freekmurze:
Hi,

Though your implementation is technically sound, I have decided not to merge this PR in.

I believe only a small amount of users will need to use custom directories. Supporting this feature adds a lot of code complexity that I'm not willing to maintain. The main thought behind this package is that a user should not worry about the directories that the media library uses.

That being said, I want to thank you for the work you've put into this. If you want to use custom directories in your own project you could use something like [Satis](https://getcomposer.org/doc/articles/handling-private-packages-with-satis.md) to use your own branched off version.

PR #55, opened 2015-09-02
Use built in morphs() function to generate polymorphic columns

uyab:
It also automatically create index for both column, which could improve performance.


freekmurze:
Nice, I didn't know that function. Thanks!

PR #64, opened 2015-09-29
Support for custom Media Eloquent model

pascalbaljet:


freekmurze:
Unless you present a convincing use case, I'm not going to merge this PR. Are you missing certains methods on Media?

pascalbaljet:
The primary reason I want this is to have relations between the Media model and other models. For example, I want to know who uploaded the media, so I now have a relationship with a User model.

freekmurze:
That's a good example. I've decided to accept your PR. Could you rebase it first? Then I'll merge your changes, do a little bit of housekeeping and tag a new version.

Thanks!

uyab:
HI @pascalbaljet, if you have relationship between Media and User model, did you add additional column to media table (something like user_id or uploader_id) ?

freekmurze:
@pascalbaljet I found a bit of time to review your changes and solved the merge conflicts myself. All changes are available in the newly tagged version 3.4.0. Thank you very much for your contribution.

@uyab If you're going to use a custom model and want to use extra field, you should write a migration to add fields (given that you're using the media table provided by this package as a base)

uyab:
Several type hint in HasMediaTrait are still pointing to Model. If I completely can use custom Model, there is a chance that I create the media model from scratch (not extend built in Media). So this will cause an error. What do you think @freekmurze ?

freekmurze:
Are you referring to the `Media` type hints in the closures?

uyab:
Yes

freekmurze:
I've removed the typehints. Unless you need this quickly, these change will get tagged the next time I'll update the package in a more meaningful way.

PR #69, opened 2015-10-03
Added support of classes morphMap

a-komarev:
There is an exception in case when trying to add media to class which exists in `Relation::$morphMap` array.

```php
class AppServiceProvider extends ServiceProvider
{
    public function boot()
    {
        Relation::morphMap([
            'page' => \App\Models\Page::class,
        ]);
    }
}
```

If model exists in morphMap array - this alias will be written in polymorphic tables as `*_type` column value.

I haven't found any clean way to fix it and just copied `\Illuminate\Database\Eloquent\Model::getActualClassNameForMorph()` method at this moment.

### DIFF
--- a/src/Conversion/ConversionCollection.php
+++ b/src/Conversion/ConversionCollection.php
@@ -2,7 +2,9 @@
 
 namespace Spatie\MediaLibrary\Conversion;
 
+use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
+use Illuminate\Database\Eloquent\Relations\Relation;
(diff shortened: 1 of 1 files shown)


freekmurze:
The tests are failing. Could you give me an example of the problem (or even better add a test)?

a-komarev:
I will recheck why tests are failed. It works for me with this code.

a-komarev:
I will close current PR and recreate a new one in separate feature branch with tests as soon as possible.

PR #70, opened 2015-10-04
Added ability to attach media to models within Relation::$morphMap array

a-komarev:
Parent issue: #69

There is 2 ways to define model morph class alias:
1. Add it to [Relation::$morphMap array](https://github.com/laravel/framework/pull/9891)
2. Define it in model's attribute $morphClass

This PR solving retrieving of concrete class name from alias only for 1st case. I'm not sure about the way to solve the 2nd one without rewriting code a lot. Maybe I missed an easier way...

### DIFF
--- a/src/Conversion/ConversionCollection.php
+++ b/src/Conversion/ConversionCollection.php
@@ -2,7 +2,9 @@
 
 namespace Spatie\MediaLibrary\Conversion;
 
+use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
+use Illuminate\Database\Eloquent\Relations\Relation;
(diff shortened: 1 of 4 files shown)


a-komarev:
I don't understand why build is failing on Relation dependency, because phpunit isn't dropping any alerts.

freekmurze:
Morphmap was introduced in Laravel 5.1.16. Using this would break compatibility with older versions.

a-komarev:
Yes... morphMap was introduced recently, but 2nd use case was implemented much earlier, but I don't know how to add support for it in smooth way.

PR #74, opened 2015-10-12
use Laravel built in bootTraits to perform deletion callback

### LINKED ISSUE #67: Trait delete method conflict
>Trait method delete has not been applied, because there are collisions with other trait methods

Conflicting with `EntityTrait` in https://cartalyst.com/manual/attributes/1.0 package.

uyab:
This also fix #67

### DIFF
--- a/src/HasMedia/HasMediaTrait.php
+++ b/src/HasMedia/HasMediaTrait.php
@@ -245,21 +245,16 @@ trait HasMediaTrait
     }
 
     /**
-     * Delete the model. The extra logic isn't handled in a model event since the boot function is unreliable
-     * for currently unknown reasons.
+     * Boot the has media trait for a model.
      *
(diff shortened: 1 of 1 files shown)


freekmurze:
The tests are failing. Could you fix that?

uyab:
Test failed when using --prefer-lowest, caused by following issue: https://github.com/danielstjules/Stringy/issues/69

freekmurze:
I'm going to leave this for now. Upping the dependancies is considered a breaking change and I'm not ready for a new major version. I'll fix this when tagging a new major version in future.

uyab:
Just curious, why use --prefer-lowest ?

freekmurze:
Here is some background info: http://evertpot.com/testing-composer-prefer-lowest/

Issue #19, opened 2015-07-14
Functionality to add remote media links


coding-sunshine:
Hi, yours is a great library and i am loving to use it.
I haws wondering how we can add remote urls as a media so it doesn't give file not found exception?
I am working on an app where you can upload your files or just add links to Youtube/vemeo or any direct images to the section and it will fetch from there.
I wanted to know if that is possible with your package?

freekmurze:
This package is built around adding files, there are currently no plans to add url's. But I'll take a look in the next weeks if this can be elegantly added to the package.

coding-sunshine:
thanks very much :)

freekmurze:
As we do not need this functionality ourselves adding it is low priority. If you need this, feel free to submit a pull request.

coding-sunshine:
Thanks, will submit a PR soon.

cviebrock:
+1 on this.  Something like `$model->addMediaFromUrl($remoteUrl)->toMediaLibrary()` would be nice to have.

freekmurze:
I think it would be nice to have this. It's still low priority for us, but we'd happily accept a good PR that adds this to the package.

freekmurze:
Thanks to @cviebrock this feature is now available in v.3.8.0

Sisifo-Developer:
+1 to this request.

Unfortunately, I have not the knowledge for a PR...

PR #96, opened 2015-11-07
Feature/add from url

cviebrock:
This is a solution for #19 and lets you add remote files to the media library:

```php
$model->addMediaFromUrl('http://medialibrary.spatie.be/assets/images/mountain.jpg')
    ->toCollection();
```

### DIFF
--- a/src/Exceptions/UrlCouldNotBeOpened.php
+++ b/src/Exceptions/UrlCouldNotBeOpened.php
@@ -1,0 +1,9 @@
+<?php
+
+namespace Spatie\MediaLibrary\Exceptions;
+
+use Exception;
+
+class UrlCouldNotBeOpened extends Exception
+{
+}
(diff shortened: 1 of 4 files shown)


freekmurze:
Thank you very much. I'll tag a new version containing these changes soon.

Sisifo-Developer:
Hi,

Firstly, really thanks for your great plugin! I love it.

I arrived here from #19, I understand that the idea is to save the links to youtube/vimeo  without storing the whole video. So, to have images, docs, or whatever filestorage and video links. In that way, it will be possible to have a "complete media library" of a model.

It is suppose that addMediaFromUrl solve #19, however I can not store a link; and I don't see in the official documentation any commet about this functionality. I try something similar to this:

$y->addMediaFromURL('https://www.youtube.com/watch?v=xgSYcrH36Pk')->toMediaCollection('video',media);

Could you clarify me if it is possible to storage "youtube links" in spatie-MediaLibrary? If yes, could you told me where it is documented?

Really thanks

Sisifo

PR #101, opened 2015-11-10
Fix bug when using a custom class (defined via string in config)

cviebrock:
When you simplified the big if statement with 2fe6fae, you forgot to instantiate the custom class, if it was specified.  This fixes that.


freekmurze:
Whoops, sorry about that

PR #116, opened 2015-12-03
Accept Symfony\Component\HttpFoundation\File\File objects

cviebrock:
This allows the `addMedia()` method to allow regular Symfony _File_ objects, in addition to the current _UploadedFile_ objects (or strings).


freekmurze:
Great, thanks!

PR #124, opened 2015-12-21
Use Database Path for Publishing Migrations.

irazasyed:
The current method always publishes the migrations to a hardcoded path (`<basePath>/database/migrations`). But in cases where the database path is different from the standard and default, the vendor publish command ends up always publishing this migrations file.

Using `database_path` method would publish to the right directory as per the config (default or custom path).


freekmurze:
Thanks!

irazasyed:
:+1:

freekmurze:
Tagged!

irazasyed:
Thanks!

Issue #132, opened 2016-01-06
Setting Cache Control


michaelstivala:
Hello Spatie.

I'm trying to solve a particular problem when it comes to setting the browser cache for resources stored on external disks - to elaborate, I upload my media to S3 and then use cloudfront as a CDN to deliver the media as quickly as possible.

The above works as expected using your *excellent* package. The issue was brought to my attention through Google's PageSpeed Insight tool which complains that media hosted on cloudfront doesn't leverage browser caching:

![image](https://cloud.githubusercontent.com/assets/4493561/12145600/776ab8ee-b48f-11e5-86f9-798ecb235aaf.png)

Below are some possible solutions.

*Manually change headers*
Manually log into S3 and set the Cache-Control property, this solves the issue - but is obviously not a viable solution. 

*Pragramatically add headers*
Add the ability to set the Cache-Control header (or other headers) via this package. I tested this quickly by modifying Spatie\MediaLibrary\Filesystem::copyToMediaLibrary to read as follows:

```
public function copyToMediaLibrary($file, Media $media, $conversions = false, $targetFileName = '')
    {
        $destination = $this->getMediaDirectory($media, $conversions).
            ($targetFileName == '' ? pathinfo($file, PATHINFO_BASENAME) : $targetFileName);

        if ($media->getDiskDriverName() === 'local') {
            $this->filesystem
                ->disk($media->disk)
                ->put($destination, fopen($file, 'r+'));
        } else {
            $this->filesystem
                ->disk($media->disk)
                ->getDriver()
                ->put($destination, fopen($file, 'r+'), ['ContentType' => File::getMimeType($file), 'CacheControl' => 'max-age=604800']);
        }
    }
```
(Note the new 'CacheControl' key in the array)

The ability to add the Cache-Control header could be moved to a config to give the developer more control, or potentially be expanded to allow any additional headers to be added for each file.

*Create a custom S3 Adaptor*
I haven't tried this yet, but referencing this [stackoverflow question](http://stackoverflow.com/a/30269860/2426387), the developer could write a custom adaptor that includes default settings for file upload, specifying and defaults that the project requires.

Do you have any thoughts and opinions on the above? I'd love to hear feedback from the creators.

Thanks for listening, and for providing this package.

freekmurze:
Hi Maltese Rockstar,

thank you very much for this quality explanation of the problem. I also appreciate that you suggest some quality solutions. Good job. :+1: 

[...]

PR #134, opened 2016-01-08
Add configurable headers when uploading media to a remote disk

michaelstivala:
This commit resolves issue #132 and allows the developer to specify any additional headers to include when uploading media to a remote disk.

The Spatie\MediaLibrary\Filesystem::copyToMediaLibrary method has been updated to merge any headers set in the updated laravel-medialibrary config file with the default 'Content-Type' header.

---

Your opinion regarding tests would be appreciated - I've tested this manually, but couldn't find automated tests for this class to contribute to. Would you say that this change is simple enough to accept as is?

Correct me if I'm wrong but the only way I can foresee to test this change would be through a mock object, that might end up in a brittle test.

Let me know!

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,20 @@ return [
          */
         'domain' => 'https://xxxxxxx.s3.amazonaws.com',
     ],
+
+    'remote' => [
+        /**
+         * Any extra headers that should be included when uploading media to
+         * a remote disk. Even though supported headers may vary between
+         * different drivers, a sensible default has been provided.
+         *
+         * Supported by S3: CacheControl, Expires, StorageClass,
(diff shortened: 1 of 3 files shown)


freekmurze:
This PR seems ok. 

Yeah, there currently are no tests for those extra headers. I'll test it manually as well. When I have a little them I'll figure out the best way to automate this.

Thank you very much for your contribution.

freekmurze:
I did a small refactor on your changes that made it possible to write a little pragmatic test.

michaelstivala:
Excellent - your refactor makes a lot of sense, my instinct would have been to make the `getRemoteHeadersForFile` method private, but I see the value in making it public to allow a test to be written - I'll keep it in mind in the future.

Thanks Freek!

PR #137, opened 2016-01-09
When deleting, also delete conversions in custom paths

kduma:
Now, when deleting, it is performed check whether the conversions are located in a subdirectory of the original files or outside. If they are in another place they are also deleted.

### DIFF
--- a/src/Filesystem.php
+++ b/src/Filesystem.php
@@ -119,7 +119,13 @@ class Filesystem
      */
     public function removeFiles(Media $media)
     {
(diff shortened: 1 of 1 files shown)


freekmurze:
Change seems good. Could you also add a test [here](https://github.com/spatie/laravel-medialibrary/blob/master/tests/PathGenerator/BasePathGeneratorTest.php) to make sure your change works?

freekmurze:
Still waiting on that test :smile:

kduma:
Unfortunately I don't have idea, how to test this. :/ 
You don't have tests which checks actual creation or removing of files.

PR #142, opened 2016-01-17
Support for Relation::morpMap()

henrikmartinsson:
The media model uses polymorphic relationship. When morphMap is used the addConversionsFromRelatedModel method on the ConversionCollection fails. By specifying a morphMap you can use whatever type name you wish. 
```php
Relation::morphMap([
    'news' => \App\News::class
]);
```
When using this the model_type would be "news" instead of "App\News", hence the error. Getting the FQCN from the morphMap fixes that.

This guy explains morphMap: https://nicolaswidart.com/blog/laravel-52-morph-map

### DIFF
--- a/src/Conversion/ConversionCollection.php
+++ b/src/Conversion/ConversionCollection.php
@@ -2,6 +2,8 @@
 
 namespace Spatie\MediaLibrary\Conversion;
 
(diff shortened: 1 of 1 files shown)


freekmurze:
On first thought, this will make the package incompatible with previous L5 versions. Reopen if I'm wrong.

PR #151, opened 2016-01-25
Use config_path helper.

irazasyed:
This lets people use this package in Lumen if they use it with [larasupport](https://github.com/irazasyed/larasupport) or manually add the helper. 

The current method prevents people from using this package in Lumen due to Lumen not having that method by default. This PR fixes that.


freekmurze:
Thanks!

irazasyed:
No problem. Thanks for merging :+1:

Issue #152, opened 2016-01-25
Configurable deletion of media for models?


cviebrock:
Currently, the `delete()` method on the `HasMediaTrait` automatically deletes all of a model's media when the model is deleted.

It would be nice if this were configurable via a configuration value or a model property (model property would be better IMO because you might want different behaviour for different models).

Alternatively (or maybe in addition), if your Eloquent model that uses the trait also uses `SoftDeletes`,then you probably _don't_ want to automatically delete all the associated media.

Thoughts on how you'd like to handle this? I can then put together a PR for you.

cviebrock:
Another option would be to rename `delete()` to `deleteIncludingMedia()` or something, so that you have the explicit option of doing either.

freekmurze:
Hi, excellent suggestion!

I prefer a function over adding extra configuration.  `delete()` should not be changed. That would be a breaking change.  I think it's better to introduce function like `deletePreservingFiles()` (maybe the name could be a bit better).

Yet another option would be to introduce an event called `MediaWillBeDeleted`. It could be consumed like this (when hooked up in the Kernel):

```php
namespace App\Listeners;

use Log;
use Spatie\MediaLibrary\Events\MediaAddedEvent;

class MediaLogger
{
    /**
     * @param \Spatie\MediaLibrary\Events\MediaWillBeDeleted $event
     */
    public function handle(MediaWillBeDeleted $event)
    {
        $media = $event->getMedia();
       //do whatever you want here
    }
}
```php

This would allow a user to add custom logic, such as copying/moving files that will be deleted  to some archive directory.

I'd accept a PR that adds such a function and/or event.

cviebrock:
In my opinion, the user should explicitly state what happens to attached media when a model is deleted ... which is why I was pushing for the configuration option.  I agree, though, that it makes things more bit complicated.

The `deletePreservingMedia()` method -- which I submitted a PR for -- is my second choice.  Only because it forces users to (possibly) choose between two different method calls in several places in their code.  To me, it would be cleaner to just call `$model->delete()` and (similar to soft deleting) just know that It Will Work™.

(I suppose another option, although again a breaking change, would be two new traits which can be applied to models: `DeletesMediaOnDelete` and `PreservesMediaOnDelete`.)

I sort of get what you mean with the event, but the listener would be tied to the Eloquent model, not the Media model ... so I was a little confused about how best to set that up.

So .. hopefully my PR gives us both what we need. :)

[...]

PR #153, opened 2016-01-26
Delete preserving media

cviebrock:
Adds a `deletePreservingMedia()` method to the `HasMediaTrait`, so that deleting a model doesn't necessarily mean that associate media are deleted as well (see #152).


freekmurze:
Thanks again!

PR #165, opened 2016-02-16
Add command "medialibrary:clear"

Maxeee09:
Add a command to delete "medias"

### DIFF
--- a/src/Commands/ClearCommand.php
+++ b/src/Commands/ClearCommand.php
@@ -1,0 +1,81 @@
+<?php
+
+namespace Spatie\MediaLibrary\Commands;
+
+use Illuminate\Console\Command;
+use Illuminate\Console\ConfirmableTrait;
+use Spatie\MediaLibrary\FileManipulator;
+use Spatie\MediaLibrary\Media;
+use Spatie\MediaLibrary\MediaRepository;
+
+class ClearCommand extends Command
+{
+    use ConfirmableTrait;
+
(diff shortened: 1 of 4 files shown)


freekmurze:
Thanks for contributing!

I like the new functionality and will review the code in detail in the coming days.

freekmurze:
I only got one big remark: The `delete_model_and_collection_medias` should also prove that media in other collections will still exist after calling the command.

Maxeee09:
Yes @freekmurze, this is what I've done. In the `delete_model_and_collection_medias` test, first I check that `media` still exists after calling `medialibrary:clear` command with the `collection1` collection name (which is not the associated collection name). And then, I check that `media` does not exists after calling `medialibrary:clear` command with the `collection` collection name (which is the associated collection name). I agree that used names are disturbing :)

freekmurze:
Ah I see, didn't read it right the first time. All ok then!

freekmurze:
I'll fix some very small nitpicks later tonight (at the latest tomorrow) and tag a new version.

Maxeee09:
@freekmurze No problems :)

freekmurze:
I refactored the command and the tests a bit an tagged a new version `3.15`. 🎉

In your changes there was a bug present that could lead to data loss. If only `collectionName` was specified when running the command, all media would be deleted.

Maxeee09:
🎉

PR #173, opened 2016-02-25
Add mediaIds option to regenerate command

Rhincodon:
A added the ability to pass media ids array to `medialibrary:regenerate` command. When this can be needed? Let's say we uploaded our images and added them to medialibrary, and then we wanted to add them to the queue for compressing/processing with some remote API later. We compress only main image, and after that we need ability to regenerate conversions for our new replaced compressed/processed image/images. For now we can only regenerate all media images or by the model type.

### DIFF
--- a/src/Commands/RegenerateCommand.php
+++ b/src/Commands/RegenerateCommand.php
@@ -14,7 +14,7 @@ class RegenerateCommand extends Command
      *
      * @var string
      */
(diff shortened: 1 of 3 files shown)


freekmurze:
Thank you very much for this. I got a few nitpicks with the naming of some stuff, but I'll fix that myself real quick.

PR #182, opened 2016-03-19
Add ability to subscribe for the model deletion

a-komarev:
Hello, guys! I've added some functionality in my application and wanted to share it.

This will allow to easily add one line of code to `$subscribe` array in `EventServiceProviders` and perform cascade media deletion in case of owner model being deleted. More details [about Event Subscribers](https://laravel.com/docs/5.2/events#event-subscribers) could be found in documentation.

Code will look like this:

**app/Providers/EventServiceProvider.php**
```php
<?php

namespace App\Providers;

class EventServiceProvider extends ServiceProvider
{
    protected $subscribe = [
        \Spatie\MediaLibrary\Listeners\MediaListener::class,
    ];
}
```

This solution is good because it's optional and could be attached only if needed.

### DIFF
--- a/src/Listeners/MediaListener.php
+++ b/src/Listeners/MediaListener.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace Spatie\MediaLibrary\Listeners;
+
+use Spatie\MediaLibrary\HasMedia\Interfaces\HasMediaConversions;
+use Spatie\MediaLibrary\Media;
+
+class MediaListener
+{
+    /**
+     * Handle entity deleting events.
+     * @param $entity
+     */
+    public function onEntityDeleting($entity)
+    {
+        if ($entity instanceof HasMediaConversions) {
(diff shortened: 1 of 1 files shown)


freekmurze:
Thank you for taking the time to put this together.

Currently there's a delete-method on `hasMediaTrait` that handles the deletion of the media models. It works most of the time, but is incompatible with some other packages. I'll make this better when working on a new major version of the medialibrary. ETA: somewhere this year.

Meanwhile, I'm going to leave things as they are.

a-komarev:
Maybe we can just make hasMediaTrait usable by adding it's own boot method? This will fix an issue, I can provide a PR for it.

freekmurze:
Adding a boot method an listening to model events there seems like a good solutions indeed. I'd probably  accept a PR that adds that.

a-komarev:
Already testing it. Will be pushed soon.

PR #183, opened 2016-03-19
Moved deletion method to events to avoid other packages conflicts

a-komarev:
This will solve #67 issue.

### DIFF
--- a/src/HasMedia/HasMediaTrait.php
+++ b/src/HasMedia/HasMediaTrait.php
@@ -21,6 +21,22 @@ trait HasMediaTrait
     public $mediaConversions = [];
 
     /**
+     * @var bool
+     */
+    private $deletePreservingMedia = false;
+
+    public static function bootHasMediaTrait()
+    {
+        static::deleted(function ($entity) {
+            if (!$entity->deletePreservingMedia) {
+                $entity->media()->get()->map(function (Media $media) {
+                    $media->delete();
+                });
+            }
+        });
+    }
+
+    /**
      * Set the polymorphic relation.
      *
      * @return mixed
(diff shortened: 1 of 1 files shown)


a-komarev:
The last try. I don't know any more ways to solve it :}

src/HasMedia/HasMediaTrait.php
```diff
@@ -282,31 +298,14 @@ public function addMediaConversion($name)
...
-    }
-
-    /**
      * Delete the model, but preserve all the associated media.
      *
      * @return bool
      */
     public function deletePreservingMedia()
     {
+        $this->deletePreservingMedia = true;
```
sebastiandedeyne: In edge cases where `delete` would return false, you'd now have a model with `deletePreservingMedia` true, even if you set it false before. Feels slightly off to me, but I don't know if I prefer this or added complexity in this method by keeping the old value. Pinging @freekmurze for this one.

src/HasMedia/HasMediaTrait.php
```diff
@@ -282,31 +298,14 @@ public function addMediaConversion($name)
...
+
         return parent::delete();
```
sebastiandedeyne: This line could be replaced by `$this->delete()` now

freekmurze: Yup

src/HasMedia/HasMediaTrait.php
freekmurze: @sebastiandedeyne Good point and you're certainly right that it isn't 100% clean. 

On the other hand I don't think there will be many users testing the return value and calling the `delete` method after they've just called `deletePreservingMedia`.

I do feel that when a `deletePreservingMedia` gets called on the object the user doesn't intend to use it any further. So that'll probably be just the last call that object will get.

So I'm ok with this code. Sounds logical?

src/HasMedia/HasMediaTrait.php
a-komarev: No it could not. Because `deleted` event will be fired in this case and you will not be able to prevent media deletion. I've done so because unit tests were red.

freekmurze: `$this->deletePreservingMedia = true;` is indeed necessary.

But `parent::delete()` could be replace by `$this->delete()`.

src/HasMedia/HasMediaTrait.php
a-komarev: @sebastiandedeyne what is the problem that `deletePreservingMedia` is setted to true? It's private and shouldn't be used outside of the class.

src/HasMedia/HasMediaTrait.php
a-komarev: Ah... sure I will change it.

a-komarev:
Changes are done. Anything else needed for this PR?

freekmurze:
Looks good to me! Thanks for solving this issue!

a-komarev:
@freekmurze my pleasure to support such useful package.

freekmurze:
This fix included in v3.16.1

PR #184, opened 2016-03-20
Added support of morphMap method

a-komarev:
@freekmurze Sorry for resurrecting this question again, but I really want to solve an issue with `morphMap` described in PR #70 and stop forking and maintaining this package in my personal namespace.

Because `morphMap` was introduced only in Laravel 5.1.16 we should use some conditionals to keep  maintaining previous versions and don't limit usage of package for all newer Laravel versions too. Our luck PHP has `method_exists(Relation::class, 'morphMap')` that will do the job!

### DIFF
--- a/src/Conversion/ConversionCollection.php
+++ b/src/Conversion/ConversionCollection.php
@@ -2,7 +2,9 @@
 
 namespace Spatie\MediaLibrary\Conversion;
 
+use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
+use Illuminate\Database\Eloquent\Relations\Relation;
 use Spatie\MediaLibrary\Exceptions\UnknownConversion;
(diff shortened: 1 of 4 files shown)


freekmurze:
I'm going to change my stance on this. 

After talking it through with @sebastiandedeyne there's really nothing wrong with upping the required versions of the illuminate components to 5.1.16. According to the semver rules this is not a breaking change so we can do this without creating a new major version of the medialibrary. 

So, in hindsight #70 was ok.  Could you adapt this PR so it doesn't check on the existance of `MorphMap` and up the requirements of the illuminate components to 5.1.16. If Travis confirms that all is ok, I'll pull. 

Sorry for the extra hassle, your work on this is greatly appreciated.

a-komarev:
Will do it from this one soon.

a-komarev:
Code cleaned up & squashed.

freekmurze:
Nice, thanks!

PR #211, opened 2016-04-26
Feature/remove individual media event

ceesvanegmond:
This pull request enables developers to hook into an extra event, when an individual (single) media is deleted. For example, if you use an third party system that has something to do with the Media model, developers can now hook in to is.

Basically, it's a new `IndividualMediaRemoved` event. 

@freekmurze If this could be merged (and if it's a good idea), I will convert this to use at master branch.

### DIFF
--- a/src/HasMedia/HasMediaTrait.php
+++ b/src/HasMedia/HasMediaTrait.php
@@ -5,6 +5,7 @@ namespace Spatie\MediaLibrary\HasMedia;
 use Illuminate\Contracts\Events\Dispatcher;
 use Spatie\MediaLibrary\Conversion\Conversion;
 use Spatie\MediaLibrary\Events\CollectionHasBeenCleared;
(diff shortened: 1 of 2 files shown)


freekmurze:
Having a media deleted event is a good idea. But as it stand your code will only raise events when clearing a collection. Users will expect the event to fire any time a media object is deleted. I think the `MediaObserver` would be a better place to raise this event. In that way calling `$media->delete()` would also raise that event.

Name of the event could be `MediaHasBeenDeleted` if called after `$media->delete()` or  `MediaWillBeDeleted` if called before `$media->delete()`. 

Thoughts?

ceesvanegmond:
@freekmurze Sounds like a great idea. I'll will implement this very soon. Thanks for the feedback

freekmurze:
Closing this one temporarly. Reopen if you have some time to work on the PR.

PR #215, opened 2016-04-29
get media by id

mohd-isa:


freekmurze:
To get a media object by it's id, just use the model.

`Spatie\MediaLibrary\Media::find($id);`

Issue #225, opened 2016-05-12
Conversions should keep source image format (png, gif only)


marsd:
For all conversions, I have noticed that the default output format is jpg. For png and gif formats, some images might have transparency which will be lost during the conversion process. 

I have experimented with adding `'fm' => 'src'` and while this kept the correct source format, the filename is saved with `_filename_.src`. 
```
public function registerMediaConversions()
{
    $this->addMediaConversion('thumb_square_big')
        ->setManipulations([
            'w'   => 400,
            'h'   => 400,
            'fit' => 'crop',
            'fm'  => 'src',
        ])
        ->performOnCollections('upload');
}
```

Is there a "correct" way to achieve this?

freekmurze:
Currently there is no option to keep the source image format. By default it'll convert images to `jpg` and `fm` can be used to convert all images of that profile to a specific format.

I'll add an option to keep to original format in the future. If you need this quick, I'd accept a PR.

hugofabricio:
@freekmurze This pr will be accepted?

bstrahija:
This would really be great if you could merge it in.

freekmurze:
I'll try to make some time for this soon. If you need it quick, create a PR and be sure to add some tests.

bstrahija:
There's already a PR in this issue that works perfectly.

freekmurze:
I see that @marsd committed some code to his own repo. There currently is no PR.

bstrahija:
Yeah I though it was a PR. Added it now: https://github.com/spatie/laravel-medialibrary/pull/237

freekmurze:
Cool, thanks!

freekmurze:
The PR has been merged. Changes are present in v4.2

elephantux:
"spatie/laravel-medialibrary": "^4.10"

All image formats stored with .src extension :(

elephantux:
'''
UploadedFile {#605 ▼
  -test: false
  -originalName: "cover.jpg"
  -mimeType: "image/jpeg"
  -size: 940996
  -error: 0
  path: "/tmp"
  filename: "phpdvrddw"
  basename: "phpdvrddw"
  pathname: "/tmp/phpdvrddw"
  extension: ""
  realPath: "/tmp/phpdvrddw"
  aTime: 2017-01-04 19:42:24
  mTime: 2017-01-04 19:42:24
  cTime: 2017-01-04 19:42:24
  inode: 654100
  size: 940996
[...]

PR #226, opened 2016-05-12
Conversions return src format extensions #225

marsd:
getResultExtension checks if `'fm' => 'src'` has been defined, and if the image formats are png or gif to return the source's extension.


freekmurze:
Can you add tests to this PR to make sure that these changes do work? Make sure that you verify if url generation is ok for profiles with `'fm' => 'src'`.

hugofabricio:
For me this is also happening.

PR #232, opened 2016-05-26
Easily allow pseudo softdelete configuration from the model

kblais:
Actually, because you can't override a trait attribute, you need to specify that you want to delete the model without deleting the associated media using the `deletePreservingMedia()` method.

The solution I offer allows you specify it directly in your model, by overriding the `getDeletePreservingMedia()` method. This way, you can use the basic `delete()` method on your model, and still preserve your media.


freekmurze:
Thanks!

PR #246, opened 2016-06-22
Lumen support

rjbaker:
Hi there,

I've added support for Lumen 5.2. There wasn't too much requiring change, mainly figuring out how to get everything to load and configured correctly.

Key changes are removal of `Illuminate\Foundation\Bus\DispatchesJobs` trait from the FileManipulator class as this isn't present in Lumen. I have instead dispatched the job via the app helper, which is compatible with both Laravel and Lumen.

I have also updated the `MediaLibraryServiceProvider::boot()` method with a check to establish whether if it's running under Laravel or Lumen and configure accordingly.

Since there are very few functional changes to the package, I have not added any meaningful tests. However, it would be interesting to see how the test suite could be updated to run with `laravel/lumen-framework` as a dependency after running with `laravel/framework` mandated by the `orchestra\testbench` package.

Many thanks. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,8 +22,8 @@ Want to store some large files on another filesystem? No problem:
 $newsItem->addMedia($smallFile)->toCollectionOnDisk('downloads', 'local');
 $newsItem->addMedia($bigFile)->toCollectionOnDisk('downloads', 's3');
 ```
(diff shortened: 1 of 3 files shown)


freekmurze:
Thank you very much for this! Excellent job.

Issue #245, opened 2016-06-22
media:regenerate doesn't delete conversions that are no longer defined


jamiehd:
I would expect it to delete them - that way if you are no longer using a conversion you don't have orphaned files just sat there on the disc.

freekmurze:
A `medialibrary:clean` command has been added in version `4.5.0`

PR #254, opened 2016-07-04
Add clean command to delete files created for removed conversions

nicolasbeauvais:
Add a `--purge` option to the `medialibrary:regenerate` command.

Using the options will check all current conversions on disk and check if they are still defined in the current Model conversions.

if not the conversions is deleted.

Requested in #245

### DIFF
--- a/src/Commands/CleanCommand.php
+++ b/src/Commands/CleanCommand.php
@@ -1,0 +1,172 @@
+<?php
+
+namespace Spatie\MediaLibrary\Commands;
+
+use Illuminate\Console\Command;
+use Illuminate\Console\ConfirmableTrait;
+use Illuminate\Contracts\Filesystem\Factory;
+use Illuminate\Database\Eloquent\Collection;
+use Spatie\MediaLibrary\Conversion\ConversionCollection;
+use Spatie\MediaLibrary\Exceptions\FileCannotBeAdded;
+use Spatie\MediaLibrary\FileManipulator;
+use Spatie\MediaLibrary\Media;
+use Spatie\MediaLibrary\MediaRepository;
+use Spatie\MediaLibrary\PathGenerator\BasePathGenerator;
+
+class CleanCommand extends Command
+{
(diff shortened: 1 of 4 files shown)


freekmurze:
Hi, this PR seems great. Thanks for take the time to put this together.

The only thing I don't like it that `continue 2;` statement. Could you rewrite this so using `continue` isn't needed. It may be better to use a couple of functions and an early return.

nicolasbeauvais:
Will do it later today, thx :)

freekmurze:
After thinking this a bit through I'd rather have this functionality as a separate command called `medialibrary:clean`. Running that will delete any files that do not have conversions defined. In addition it would also delete files of which the `Media` model has been deleted. 

An handy option would be `--dry-run` which would just report which files would be deleted if the command is run without that option.

The benefit of a having a separate command is that the medialibrary can be cleaned without regenerating every file. The cleaning proces will be a lot faster and code probably a lot cleaner.

Could you modify the code of this PR? If you have any thoughts or comments on this approach, I'd be glad to hear them.

nicolasbeauvais:
Will do !

nicolasbeauvais:
Done :)

freekmurze:
Looking very good. I'll fix some nitpicks of mine later today and tag a new release.

Thank you again!

freekmurze:
A new version `4.5.0` containing the changes by this PR has been tagged.

For future PR's: I generally prefer using [`filter` and `each` on a collection](https://github.com/spatie/laravel-medialibrary/blob/master/src/Commands/CleanCommand.php#L153-L162) over a [`foreach` loop with a nested `if ` statement](https://github.com/spatie/laravel-medialibrary/blob/87fc1c006d680528b77a794961d6fa41019bb60f/src/Commands/CleanCommand.php#L156-L170).

nicolasbeauvais:
I've looked at your refactor commits, i will keep the change in mind for next time

PR #255, opened 2016-07-04
Simplify the conversion path logic.

Garbee:
Since this test is being linked to from the docs as an example, it is nice to simplify the test class down. This way developers don't spend time duplicating logic if they don't realize they can call the other method first.


freekmurze:
Agreed! Thanks!

PR #256, opened 2016-07-04
Support for svg conversion

nicolasbeauvais:
This is a first version to add support to svg images conversions.

It handle transparency when converted to png, and by default use a white background when converted to jpg.

The conversion could be moved to it's own package "svg-to-image" to had more feature like handling density etc..

### DIFF
--- a/src/FileManipulator.php
+++ b/src/FileManipulator.php
@@ -26,7 +26,7 @@ class FileManipulator
             return;
         }
 
-        if (in_array($media->type, [Media::TYPE_PDF, Media::TYPE_PDF]) && !class_exists('Imagick')) {
+        if (in_array($media->type, [Media::TYPE_PDF, Media::TYPE_SVG]) && !class_exists('Imagick')) {
             return;
         }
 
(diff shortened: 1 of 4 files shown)


src/FileManipulator.php
```diff
@@ -25,7 +26,7 @@ public function createDerivedFiles(Media $media)
             return;
         }
 
-        if ($media->type === Media::TYPE_PDF && !class_exists('Imagick')) {
+        if (in_array($media->type, [Media::TYPE_PDF, Media::TYPE_PDF]) && !class_exists('Imagick')) {
```
freekmurze: one of the two TYPE_PDF should be TYPE_SVG 😄

src/FileManipulator.php
```diff
@@ -109,7 +114,7 @@ public function createTempDirectory() : string
         return $tempDirectory;
     }
 
-    protected function convertToImage(string $pdfFile) : string
+    protected function convertPDFToImage(string $pdfFile) : string
```
freekmurze: Nitpick: I don't like having extensions as capitals in function names. Could you make the function name camel case?

src/FileManipulator.php
```diff
@@ -118,6 +123,20 @@ protected function convertToImage(string $pdfFile) : string
         return $imageFile;
     }
 
+    protected function convertSVGToImage(string $svgFile) : string
```
freekmurze: Same here.

freekmurze:
This looks like a great PR, thank you!

I've added some notes in the code. Could you address them?

nicolasbeauvais:
Done :)

freekmurze:
👍

freekmurze:
Changes have tagged as `4.4.0`

PR #258, opened 2016-07-05
Convert multiple if/return to switch statement.

Garbee:
This just keeps the code easier to maintain in case more things are added later. Instead of adding more if's it is just a new case/break clause. You also don't have an array for the image types that can get unruly  as more support is added.

### DIFF
--- a/src/Media.php
+++ b/src/Media.php
@@ -100,17 +100,19 @@ class Media extends Model
      */
     public function getTypeFromExtensionAttribute()
     {
-        $extension = strtolower($this->extension);
-
-        if (in_array($extension, ['png', 'jpg', 'jpeg', 'gif'])) {
-            return static::TYPE_IMAGE;
+        switch (strtolower($this->extension)) {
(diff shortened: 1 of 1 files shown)


nicolasbeauvais:
This will conflict with #256

freekmurze:
Thanks for your contribution, but I'm not going to merge this.

In general a `switch` is certainly a good option to structure this kind of code. But in this situation I prefer if's and early returns.

Garbee:
@nicolasbeauvais Depending no whichever is merged first, resolving that is simple and hardly a problem.

@freekmurze Why in this situation are if statements preferred? Knowing this will help me, and possibly others, know whether trying to clean up other parts of the codebase would be worth putting the time into.

freekmurze:
There's no real right on wrong here. To my eyes early returns just "feels" lighter.  Imo a switch is slightly more error prone, you could forget that "break" keyword. 

It just comes down to programmer preference I think.

PR #259, opened 2016-07-05
Initial checkExistence command.

### LINKED ISSUE #257: Method to list missing images.
There should be a method to look over all of the registered media files and check for their existence. Further a command should be provided to output the list for easy verification.

For example:
```sh
php artisan medialibrary:checkExistance

# If all exist
All media files exist in storage.

# If some are missing
ID: 66 missing
ID: 67 missing
ID: 89 missing
Some media files weren't found in storage. Please verify that the listed media files exist.
```

There would be an internal class/method that could be called for doing the check. That would return a collection of the media objects which a developer could then do whatever they want with. In the case of the command, it just pulls the ID out. However, one could program an administration view using those objects and filtering them to help upkeep files that may get missing.

The command would have signature like so:

`medialibrary:checkExistence {--only= : A comma separated list of models to check for.} {--except= : A comma separated list of models to exclude from the check.}`

This will allow developers to specify targets against associated models to check for. Making it easy for them if they are either only looking for one area that may have been corrupted or trying to exclude an area they know may have fluctuation.

Sound acceptable?

Garbee:
Fixes #257 

Adds a command to check for file existence. The new `Spatie\MediaLibrary\Services\CheckExistence` class is what developers can inject into their applications to take advantage of the generator as well. The artisan command introduced here is just one example of a user interface to it.

Some work still needs doing, but initial code-review can at least begin now.

To be done:
* [x] Unit tests
    * [x] Basic
    * [x] Exemption
    * [x] Inclusive
* [x] Documentation - https://github.com/spatie/docs.spatie.be/pull/36
* [x] Add argument to do a full check instantly `--all` for example. This is helpful when automating command execution.


Garbee:
This PR is ready for complete review along with the documentation. All functionality is complete and tested.

freekmurze:
Hi, 

Thank you for the work you've put into this.

I'm probably not going to merge this in. I don't feel that this functionality is needed for most users of this package.

Derived images should always exists. If not, because a new conversion is defined on a model later on, users can use the `medialibrary:regenerate` command. Currently @nicolasbeauvais is working on a `medialibrary:clean` command that will remove any files that were produced for a conversion that has been removed (#254). 

So I think users have enough options to keep their medialibrary healthy.

Garbee:
`medialibrary:regenerate` is **exactly** why I need this upstream. It *fails* outright if you are missing a primary image. So you have to go remove the entry, then try again and hope another isn't missing.

I do plan to add a flag to that for fixing that problem. However, this command and service layer are still extremely useful. Since it allows you to easily keep *constant check* on your media.

This also has nothing to do with the clean command. That is an entirely different kind of issue.

I guess if this service isn't accepted upstream I'll need to fork the package outright to improve it for large-scale media sites. Currently it can't gracefully handle any issues that occur where a media item ends up missing. And when you have tens-of-thousands of media files, that is crucial to be able to handle.

freekmurze:
You're of course free to fork the package to mold it to your needs.

I've made some changes in the master branch so that the `regenerate` command won't fail if there are any missing files. I'll tag these changes soon.

Garbee:
> Derived images should always exists. If not, because a new conversion is defined on a model later on, users can use the `medialibrary:regenerate` command.

Also on this note there is still no verification that the *primary files* still exist which are a problem. So it completely dodges the point of "do my main files still exist."

How are applications supposed to know en-mass if the **primary** media files still exist? Or is every app using your package supposed to custom-code their own solution to this problem?

This is *massive* issue with this package for sites that have large-scale media.

freekmurze:
For the projects I've used the package on the package is fully responsible for the directories and files it is managing. There are no external processes / users messing with the files. 

If you are not in that situation you certainly fork this package to customize it to fit your needs.

Garbee:
I happen to have built a system automating trying to pull in 30 thousand images initially and several hundred more every few months. Some of these fail creating files for who-knows-why, yet the DB is still updated.

At scale, this package doesn't handle helping people with large datasets like this out. I'm also very confused as to why the package is open-source, but only things you need within your apps are approved for merge. Open-source is about helping everyone out. This package could greatly benefit from some thought-process as to how it works at scale, and everyone can benefit. Yet, it is being turned aside with only "we've never needed it."

I'm simply upset that other users of your package can't benefit from this addition if they are having problems doing the same type of thing. Nothing here affects how any existing apps operate or how future things would operate. It is extremely puzzling.

freekmurze:
I don't agree with your assertion that open source is about helping everybody out. 

Whenever my colleagues or myself encounter a problem in one of our projects we take the time to create a package. The primary reason for doing so is so we can reuse the code ourselves. For everything we put out we write tests, follow semver, write docs and follow most best practices. Those packages show of our expertise to external parties so there's certainly a commercial aspect to it as well.

Like I've [stated](https://murze.be/2016/01/why-we-are-requiring-php-7-for-our-new-packages/) multiple times in the past the popularity of our packages is not our main concern. Others are free to use our solutions and we're grateful for all the feedback we're getting. If you take a look at [the PR submitted to our organisation](https://github.com/pulls?q=is%3Apr+user%3Aspatie+is%3Aclosed) you'll see that we generally are very willing to accept PRs.

Think about this: if we accept your PR we are responsible for maintaining your code [forever](https://twitter.com/solomonstre/status/715277134978113536). We rather would maintain things we need ourselves than things we don't need. [Simple things we don't really need ourselves generally will get merged](https://github.com/spatie/laravel-activitylog/pull/19), but big complex pieces of code (like the one in this PR) would just take too much time to maintain.

You are entirely free to fork our package (hurray open-source 😄) and build a solution for your needs on top of our package in which we already invested countless hours. You're getting al of our current work for free! You only need to add your custom bits. If you don't want to do this, you're of course free to look for another solution or to built something from scratch yourself.

If you want us to work on your specific use case and your specific code, we're available for hire.

Garbee:
So, there is a technical reason, it is too complex. Do you mind pointing out the complex areas? I could possibly reduce the solution to be simpler. The service layer is the important piece here and it is quite trivial. The artisan command is certainly slightly-complex, however that is to provide the best user-experience.

freekmurze:
The functionality just isn't needed for us and most users of this package.

"Some of these fail creating files for who-knows-why, yet the DB is still updated." --> I'd look more into this and try to assert the media is ok when creating it.

Like stated before I think you're better off forking this package and adjusting it to your needs. Problem solved for you? No?

If more people report the same issue, I'll reconsider this decision.

Issue #271, opened 2016-07-14
Deleting custom property


iCodem:
How can I delete a custom property from a mediaItem?

freekmurze:
There currently isn't. You could set the property to `null`.

I'd accept a PR that adds a `removeCustomProperty` method to the `Media` class.

iCodem:
Thanks guys. so I'd suggest adding this feature as well: ability to change/update CustomProperty after adding new media.

freekmurze:
Added `removeCustomProperty` in version `4.6.0`

You can update a custom property by just setting it again.

PR #272, opened 2016-07-14
Adds removeCustomProperty() to the Media class

legshooter:
For #271

### DIFF
--- a/src/Media.php
+++ b/src/Media.php
@@ -188,6 +188,14 @@ class Media extends Model
     {
         $this->custom_properties = array_merge($this->custom_properties, [$name => $value]);
     }
+    
+    /**
+     * @param string $name
+     */
+    public function removeCustomProperty(string $name)
+    {
+        unset($this->custom_properties[$key]);
+    }
 
     /*
      * Get all the names of the registered media conversions.


freekmurze:
Hi, thanks for contributing.

Could you also add a test to make sure your change works?

legshooter:
Done. Is that ok?

freekmurze:
Yup, thanks. For future reference you can push extra commits to a PR so you don't have to create a second PR.

legshooter:
Right, I thought I did, sorry :/

freekmurze:
After reviewing the PR in more detail it turned out that it did not work. `$name` was passed in, but `$key` was used in the function. Also `$this->custom_properties` cannot be modified directly.

When you're submitting a PR it always good to test the changes yourself, either by using it manually in a project or an automatic test.

Because it might come across as I'm being mad at you, please rest assured that I am not. We're all learning. I just want to give you some friendly advice for a future PR you're making either at this repo or some else's.

legshooter:
argh. Nah man, it's ok, this is the exact reason why I avoid PRing, I don't want to f:boom:ck things up and waste people's time.

Still got a way to go with OS/Git/GitHub/PRing/testing ($key/$name part has nothing to do with it of course, that was just a careless mix up)... I think I told you that once in the PJAX Middleware context. Which reminds me I made a nice function (I think) to bust the cache automatically, but I don't wanna embarrass myself with another PR, haha :)

I do have a couple of questions about stuff I am still not sure about, if you're willing.

You wrote this check:
```php
if ($this->hasCustomProperty($name)) {
    // ...
}
```
This way if I pass a non-existing `$name`, I'll never know that it wasn't in `$this->custom_properties` to begin with. The end result is indeed the same, `$this->custom_properties` will not contain a `$name` key, but I thought there's value to erroring out on passing non-existing keys. Was I supposed to know on my own that a check is preferred to let non-existing names "slide"?

Also, are you saying you can only overwrite `$this->custom_properties`? This is why you're overwriting it with a merged (with itself) array in `$this->setCustomProperty()`? Does this have anything to do with the fact `$this->custom_properties` is not explicitly declared anywhere?

PR #277, opened 2016-07-16
Add get mime method

nicolasbeauvais:
Add a `getMimeAttribute` method to the Media model.

It can be useful to access the raw mime type easily from a media instance, like:
```html
<video controls>
  <source src="{{ $media->getUrl() }}" type="{{ $media->mime }}" />
</video>
```


freekmurze:
Thanks!

PR #278, opened 2016-07-20
Use `putStream` instead of `put` for cloud drivers

hosmelq:
I'm using this adapter for Google Cloud Storage https://github.com/Superbalist/flysystem-google-storage, but i'm getting this error.

```
ErrorException in MediaFileUpload.php line 88: strlen() expects parameter 1 to be string, resource given
```


hosmelq:
I pushed this to v3 too. Is the version that i'm using. 😕

freekmurze:
see #279

PR #279, opened 2016-07-20
Use `putStream` instead of `put` for cloud drivers

hosmelq:


freekmurze:
Not sure if `putStream` is available on all drivers. I'll check when I have a bit of time on my hands.

hosmelq:
All the base adapters have `putStream ` available.

local: https://github.com/thephpleague/flysystem/blob/master/src/Adapter/Local.php#L147

ftp: https://github.com/thephpleague/flysystem/blob/master/src/Adapter/Ftp.php#L222

s3: https://github.com/thephpleague/flysystem-aws-s3-v3/blob/master/src/AwsS3Adapter.php#L352

rackspace: https://github.com/thephpleague/flysystem-rackspace/blob/master/src/RackspaceAdapter.php#L190

freekmurze:
Turns out that `putStream` isn't available for all drivers.

PR #281, opened 2016-07-24
Get mime type of file, from all driver

nicolasbeauvais:
Once a file is stored on a non-local server

```php
$media->getTypeFromMimeAttribute();
$media->getMimeAttribute();
```

will not work, as `File::getMimetype($this->getPath());` only work on the local driver.

I see two workarounds to always have the possibility to get a file mime:

1. use the `File::getMimetype` method before storing the file and store the value in the media table.
2. use a table of correspondence that return the file mime by using the file extension.

 What do you think about the feature and which one would be the best to implement ?


freekmurze:
Option 1: modifying the db at this point. So I'd rather not do that now.
Option 2: I don't want to maintain such a table of correspondence.

Maybe we should leave things as they are for now. I'll address this is the next major version of the package. 

If you want to store mime types, you can use custom properties for that in your own project.

PR #283, opened 2016-07-29
Add force option to commands

nicolasbeauvais:
I've added the `--force` option to the three tmedialibrary commands so we can use them in production environment:

before:
capture d ecran de 2016-07-29 18 56 27

after:
capture d ecran de 2016-07-29 19 05 16

I've chosen to put optional args in the `getOptions` method instead of the command signature to keep them clean and easily readable.

+ I re indented [this line](https://github.com/spatie/laravel-medialibrary/blob/master/src/Commands/RegenerateCommand.php#L69)

Now that you now that I use the package in production I must send you a postcard :smile:

### DIFF
--- a/src/Commands/CleanCommand.php
+++ b/src/Commands/CleanCommand.php
@@ -12,6 +12,7 @@ use Spatie\MediaLibrary\FileManipulator;
 use Spatie\MediaLibrary\Media;
 use Spatie\MediaLibrary\MediaRepository;
 use Spatie\MediaLibrary\PathGenerator\BasePathGenerator;
+use Symfony\Component\Console\Input\InputOption;
 
 class CleanCommand extends Command
 {
@@ -22,7 +23,7 @@ class CleanCommand extends Command
      *
      * @var string
      */
-    protected $signature = 'medialibrary:clean {modelType?} {collectionName?} {disk?} {--dry-run}';
+    protected $signature = 'medialibrary:clean {modelType?} {collectionName?} {disk?}';
 
     /**
      * The console command description.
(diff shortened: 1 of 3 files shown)


freekmurze:
Hi,

thanks for contribution again to the medialibrary 👍

most importantly: yes, you are required to send a postcard 📮  

I like the addition of the `--force` flag, but I see a few issues.

- the tests are failing due to the fact that `dry-run` is removed. Is this an error on your part or do you have a specific reason?
- the regenerate does not use the `confirmableTrait` yet. Could you add that?

nicolasbeauvais:
My bad, apparently `getOptions` is ignored if we define `$signature`, this is why `dry-run` disappeared.

I've had the confirmation on the regenerate command :smile:

src/Commands/CleanCommand.php
```diff
@@ -22,7 +22,9 @@ class CleanCommand extends Command
      *
      * @var string
      */
-    protected $signature = 'medialibrary:clean {modelType?} {collectionName?} {disk?} {--dry-run}';
+    protected $signature = 'medialibrary:clean {modelType?} {collectionName?} {disk?} 
+    {--dry-run : List files that will be removed without removing them},
+    {-- force : Force the compiled class file to be written}';
```
freekmurze: The description of `--force` seems to have been copy pasted 😄 
Could you improve the description?

freekmurze:
Added a little comment in the code. Will pull after that's fixed.

freekmurze:
Thank you very much for this!

nicolasbeauvais:
You're welcome :) It's my pleasure to contribute to this really great package :sunglasses:

freekmurze:
Changes have been tagged: `4.8.0`

PR #284, opened 2016-08-02
Add missing extensions to keep when converting

bstrahija:
The .jpg and .jpeg extensions where set to .src when using 'fm': 'src' in the conversion. This should fix this. If you could merge this in as soon as possible it would be greatly appreciated since we have a project thats going live very soon ;)


sebastiandedeyne:
Thanks, tagged in v4.7.1!

https://github.com/spatie/laravel-medialibrary/releases/tag/4.7.1

bstrahija:
Awesome, thanks

PR #309, opened 2016-08-25
add getLastMedia* methods into HasMediaTrait

dedenhabibi:
It is just a small adding. I need a shortcut to get the latest image from a collection in many cases such as getting the latest uploaded profile picture of a user, while preserving the old ones for reuse.

I don't know about the performance consideration, but I think it is complementing for the _getFirst*_ pair. I am using this additional snippet now, hopefully it could be official :)

### DIFF
--- a/src/HasMedia/HasMediaTrait.php
+++ b/src/HasMedia/HasMediaTrait.php
@@ -172,6 +172,53 @@ trait HasMediaTrait
 
         return $media->getPath($conversionName);
     }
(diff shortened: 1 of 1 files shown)


freekmurze:
Thank you for this, I'll probably pull your PR if you provide some tests to make sure it works.

freekmurze:
Closing this for now. Reopen when you're adding some tests.

PR #310, opened 2016-08-25
Add unit test for video svg and pdf conversion

nicolasbeauvais:
Really small cleanup + new unit tests for video / svg / pdf conversions


freekmurze:
Perfect, thanks yet again for making medialibrary better.

PR #315, opened 2016-08-29
Checking path before Finfo

taliptako:
Sometimes i can add media without file

### DIFF
--- a/src/Helpers/File.php
+++ b/src/Helpers/File.php
@@ -38,8 +38,13 @@ class File
      */
     public static function getMimetype(string $path) : string
     {
-        $finfo = new Finfo(FILEINFO_MIME_TYPE);
-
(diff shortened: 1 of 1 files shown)


taliptako:
I can fix styleci error if you accept this pull patch

freekmurze:
What's your use case for this change?

taliptako:
I'm trying to create media  and my media also have title and desc fields.When user create media without file just with title and desc i'm creating media like this $post->media()->create($fields); and i'm just getting error from Finfo i dont understand why because i'm not adding it from $post->addMedia();

freekmurze:
Imo a Media object should always have a valid file.

To avoid the usage `Finfo` in your project you could use a custom `Media` model overriding any function that are causing your troubles.

Issue #326, opened 2016-09-12
Media's URLs are not URL encoded, when preserving original file names.


hootlex:
There is a breaking issue with filenames not being URL encoded.
For example, if there is an image named _'This is my awesome image.png'_, media's `getUrl()` method returns `http://mydomain.com/This is my awesome image.png` **instead of** `http://mydomain.com/This%20is%20my%20awesome%20image.png` or `http://mydomain.com/This+is+my+awesome+image.png`.

The resulting URL works fine in some situations (such as within an `img` tag) but it causes issues when is parsed by scraper's, such as Facebook.

![](http://image.prntscr.com/image/0b7df1f44c194789b6585862793ab066.png)

GaspariLab:
It looks like this bug still exists.

Commit 48ed0b4 removed the rawurlencode function, so filenames with spaces are still broken.

PR #327, opened 2016-09-12
urlencode file name in media's URL

hootlex:
Solves https://github.com/spatie/laravel-medialibrary/issues/326


freekmurze:
Thanks

hootlex:
@freekmurze can you apply this to v3.* too? :)

GaspariLab:
It looks like this bug still exists.

Commit 48ed0b42fb1f3a0aa54000f64a16ac4fd68e2ed5 removed the `rawurlencode` function, so filenames with spaces are still broken.

Issue #325, opened 2016-09-11
File Conversions Not Created with S3


bakerstreetsystems:
I'm using Spatie/Laravel-Medialibrary to handle file attachments to my Product model. I'd like to have all product images upload to Amazon S3 and then have each image converted to different sizes automatically (i.e. "thumb", "small", "medium", "large").

Everything works great when working with the local filesystem. The package creates the conversions and I'm able to access them with my app. The problem comes when I try to change the package's settings to use Amazon S3 instead of the local file storage.

So in my `ImageController.php`, I have the following line:

`$product->addMedia(public_path() . '/temp/products/'.$product->id)->toCollection('images', 's3');
`
S3 config in `config/filesystems.php`

```
    's3' => [
            'driver' => 's3',
            'key' => env('AMAZON_S3_ID'),
            'secret' => env('AMAZON_S3_KEY'),
            'region' => env('AMAZON_S3_REGION'),
            'bucket' => env('AMAZON_S3_BUCKET'),
            'visibility' => 'public'
        ],
```

and the conversion settings located in my `Product.php` class:

```
    public function registerMediaConversions()
    {
        $this->addMediaConversion('thumb')
             ->setManipulations(['w' => 160, 'h' => 120])
             ->performOnCollections('images');

        $this->addMediaConversion('small')
             ->setManipulations(['w' => 280, 'h' => 210])
             ->performOnCollections('images');

        $this->addMediaConversion('medium')
             ->setManipulations(['w' => 400, 'h' => 300])
             ->performOnCollections('images');

        $this->addMediaConversion('large')
             ->setManipulations(['w' => 640, 'h' => 480])
             ->performOnCollections('images');
    }
```

As I mentioned above, everything works great if I use the local file system, but as soon as I change it to use Amazon S3, the file conversions are not created. The original file is successfully uploaded to Amazon, but the conversions are not there. Any advice?

nicolasbeauvais:
Hi @bakerstreetsystems,

can you please check this recent similar issue #321, and make all the checks mentioned there ?

Does `php artisan medialibrary:regenerate` create the conversions ?
Are you using a queue ?
...

Thanks !

bakerstreetsystems:
[...]

PR #329, opened 2016-09-12
Fix for conversions not being generated on S3 filesystem

bakerstreetsystems:
This is a proposed fix for issue #325 . Basically, the Spatie\MediaLibrary\FileManipulator::createDerivedFiles() method was returning early without creating any conversions because the package could not determine the mime type for a file located on S3.


freekmurze:
Could you rebase this against the current master?

src/Media.php
```diff
@@ -7,6 +7,7 @@
 use Spatie\MediaLibrary\Conversion\ConversionCollection;
 use Spatie\MediaLibrary\Helpers\File;
 use Spatie\MediaLibrary\UrlGenerator\UrlGeneratorFactory;
+use Storage;
```
sebastiandedeyne: The full namespace should be imported here, not the alias

PR #361, opened 2016-09-28
Fix s3 bug on 4.9.1

### LINKED ISSUE #359: Upgrading from 4.8.4 to 4.9.1, Call to undefined method Spatie\MediaLibrary\UrlGenerator\S3UrlGenerator::getPath()
After upgrading this package, the error now occurs when I upload from file to S3. None of my upload code was modified between the upgrade.

Current solution is to specify in composer `"spatie/laravel-medialibrary": "^4.8.4"`

Here is a snippet of what my code is doing:
```
$filename = 'random-generated-filename';
$uploaded = $request->file('media');
$return = $owner->addMedia($uploaded);
$extension = $uploaded->getClientOriginalExtension();
$filename = $filename . '.' . $extension;

$return = $return->preservingOriginal()
    ->usingName($filename)
    ->usingFileName($filename)
    ->toCollection('uploaded');
```

```
Symfony\Component\Debug\Exception\FatalThrowableError: Call to undefined method Spatie\MediaLibrary\UrlGenerator\S3UrlGenerator::getPath()
1 File "/var/www/myapp/vendor/spatie/laravel-medialibrary/src/Media.php" line 86 in getPath
2 File "/var/www/myapp/vendor/spatie/laravel-medialibrary/src/FileManipulator.php" line 138 in determineImageGenerator
3 File "/var/www/myapp/vendor/spatie/laravel-medialibrary/src/FileManipulator.php" line 44 in performConversions
4 File "/var/www/myapp/vendor/spatie/laravel-medialibrary/src/FileManipulator.php" line 25 in createDerivedFiles
5 File "/var/www/myapp/vendor/spatie/laravel-medialibrary/src/Filesystem.php" line 42 in add
6 File "/var/www/myapp/vendor/spatie/laravel-medialibrary/src/FileAdder/FileAdder.php" line 328 in toCollectionOnDisk
7 File "/var/www/myapp/vendor/spatie/laravel-medialibrary/src/FileAdder/FileAdder.php" line 285 in toCollection
```

nicolasbeauvais:
Should fix #359 

as @bakrpx mentioned in the issue the media->getPath()` method didn't existed for s3.

What I did to fix the bug is simply generating the original file copy on disk before checking if we can convert it, so we could always give a local path to `determineImageGenerator`.

It isn't the most efficient as a temp file will be created even if we can't convert the file, but it should work with all remote filesystem

### DIFF
--- a/src/FileManipulator.php
+++ b/src/FileManipulator.php
@@ -129,10 +129,8 @@ class FileManipulator
      */
     public function determineImageGenerator(Media $media)
     {
-        $isLocal = $media->getDiskDriverName() === 'local';
-
         $imageGenerators = $media->getImageGenerators()
-            ->map(function (string $imageGeneratorClassName) use ($isLocal) {
+            ->map(function (string $imageGeneratorClassName) {
                 return app($imageGeneratorClassName);
             });
 
--- a/src/FileManipulator.php
+++ b/src/FileManipulator.php
@@ -133,7 +133,7 @@ class FileManipulator
 
         $imageGenerators = $media->getImageGenerators()
             ->map(function (string $imageGeneratorClassName) use ($isLocal) {
-                return app($imageGeneratorClassName)->shouldCheckMime($isLocal);
+                return app($imageGeneratorClassName);
             });
 
         foreach ($imageGenerators as $imageGenerator) {
--- a/src/FileManipulator.php
+++ b/src/FileManipulator.php
@@ -129,13 +129,15 @@ class FileManipulator
      */
     public function determineImageGenerator(Media $media)
     {
+        $isLocal = $media->getDiskDriverName() === 'local';
+
         $imageGenerators = $media->getImageGenerators()
-            ->map(function (string $imageGeneratorClassName) {
-                return app($imageGeneratorClassName);
+            ->map(function (string $imageGeneratorClassName) use ($isLocal) {
+                return app($imageGeneratorClassName)->shouldCheckMime($isLocal);
             });
 
         foreach ($imageGenerators as $imageGenerator) {
-            if ($imageGenerator->canConvert($media->getPath())) {
+            if ($imageGenerator->canConvert($media)) {
                 return $imageGenerator;
             }
         }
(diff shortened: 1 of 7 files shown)


freekmurze:
Thank you! I'll review this soon!

freekmurze:
I'm not a big fan of copying the file back over to the local filesystem. Image you're processing a 500 MB file, ... that'll be slowwwww. 

For external filesystem we could just use the `file_name` (stored in the db) to determine the image generator. We could add a method `shouldCheckMime` on the `BaseGenerator`.

In `FileManipulator` we could check if we're working with a local filesystem. If we are, we pass `false` to `shouldCheckMime`

Thoughts?

nicolasbeauvais:
It looks better now :)

src/ImageGenerators/BaseGenerator.php
```diff
@@ -4,24 +4,24 @@
...
+    protected $shouldCheckMime = true;
 
+    public function canConvert(Media $media): bool
+    {
         if (! $this->requirementsAreInstalled()) {
             return false;
         }
 
-        if ($this->supportedExtensions()->contains(pathinfo($path, PATHINFO_EXTENSION))) {
+        if ($this->supportedExtensions()->contains($media->getExtensionAttribute())) {
```
freekmurze: This can just be `$media->extension`. Laravel magic 😄

freekmurze:
Much better. But after seeing your solution (which would work) and giving it some thought I think we can even go for a more simpler solution.

Instead of adding `shouldCheckMime` all over the place we could just replace this check:

```php
$this->shouldCheckMime && file_exists($media->getPath()
````

by 

`method_exists($media, 'getPath') && file_exists($media->getPath())`

If I'm right the right condition won't be executed if the left one is false. 

Could you make these changes?

**Comment by freekmurze** [comment]
fyi: changing the type of what gets passed in `canConvert` is a breaking change. But I'll allow it this time, because probably not many users will have built a custom converter.

**Comment by nicolasbeauvais** [comment]
Nice catch, indeed the right condition will only be executed if the left one is true !

File: src/FileManipulator.php [hunk 2]
```diff
@@ -129,13 +129,15 @@ protected function dispatchQueuedConversions(Media $media, ConversionCollection
      */
     public function determineImageGenerator(Media $media)
     {
+        $isLocal = $media->getDiskDriverName() === 'local';
```
freekmurze: `isLocal` can be removed

nicolasbeauvais: done

freekmurze: Thanks!

PR #363, opened 2016-10-08
Fix typo and sync readme content with config stub

drbyte:


freekmurze:
Thank you very much! 👍

PR #366, opened 2016-10-09
Remove unused supportedTypes method from image generators

nicolasbeauvais:
I was writing the doc for creating custom generators and noticed the `supportedTypes` isn't used anywhere on the project.


freekmurze:
Nice! Thank you! 
Looking forward to those docs.

PR #370, opened 2016-10-12
Case insensitive extension and mime check

nicolasbeauvais:
Fix suggested by @boofman15 in #325


freekmurze:
Thank you again! 👍 

Because you're doing so much I've added your name in the credits section of the readme.

nicolasbeauvais:
Thanks a lot :smile:

PR #371, opened 2016-10-13
Move image generators in configuration

nicolasbeauvais:
I think it will be easier to configure ImageGenrators in the medialibrary configuration, probably a change for v5 ?


freekmurze:
Good idea, but we'll indeed hold this off until the next major version.

freekmurze:
I started work on `v5`, so breaking changes are allowed. I foresee no major new features, just a good clean up of the code and we'll drop support for anything lower than Laravel 5.4.

If you have any more suggestions for the package, let me know.

PR #377, opened 2016-10-20
Allow custom job classes.

Garbee:
If someone is using an alternate bus package that allows non-selfhandling jobs, the current job will fail in these situations. This modification allows developers to override the job class itself to support alternate busses by extending the internal job.

The change is non-breaking since it defaults to the internal job even if the config key is not defined respecting existing rollouts.

### DIFF
--- a/config/laravel-medialibrary.php
+++ b/config/laravel-medialibrary.php
@@ -21,6 +21,11 @@ return [
     'queue_name' => '',
 
     /*
+     * The class to use for the queue job to convert media.
+     */
+    'queue_job_class' => \Spatie\MediaLibrary\Jobs\PerformConversions::class,
+
+    /*
      * The class name of the media model to be used.
      */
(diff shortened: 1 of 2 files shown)


freekmurze:
I don't think there are a lot of users that need this. Holding off on this for now.

Plytas:
Can this be looked at again? I'm using https://github.com/hyn/multi-tenant package and for the jobs to work, they need to be tanant aware. That package provides a trait that extends `SerializesModel` trait to attach some extra attributes required for tenant identification. Currently there's no way to extend `PerformConversion` job to add this functionality.

freekmurze:
Merge conflicts with this PR, I'll make the necessary changes for the current version myself. You can expect a new version later today.

luceos:
Thanks @freekmurze much appreciated! 🙇

freekmurze:
No problem @luceos, changes are now available in `v7.4.0`.

PR #378, opened 2016-10-20
Fix s3 conversion (undefined getPath method)

nicolasbeauvais:
I reworked #329, I tested it on a new project against a real s3 bucket.

Is there a way to perform some basic unit test with files on s3 (at least upload, conversion & url generation) ? I know that we should not test "s3" itself but part of the medialibrary code is different for local and remote files, which caused some major bug on the lasts versions.

What is your thought on this ?

### DIFF
--- a/src/ImageGenerators/BaseGenerator.php
+++ b/src/ImageGenerators/BaseGenerator.php
@@ -5,6 +5,7 @@ namespace Spatie\MediaLibrary\ImageGenerators;
 use Illuminate\Support\Collection;
 use Spatie\MediaLibrary\Helpers\File;
 use Spatie\MediaLibrary\Media;
+use Spatie\MediaLibrary\UrlGenerator\UrlGeneratorFactory;
 
 abstract class BaseGenerator implements ImageGenerator
 {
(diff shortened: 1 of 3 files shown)


freekmurze:
The users of this package and myself thank you for this PR. 👍 

Sadly, I haven't good a good way to automatically test S3 functionality. Most of the time I do it the same way as you did: just try it in a real project. I'll give this some thought. I'll let you know if I come up with a good idea

nicolasbeauvais:
I see two solutions:

- using something like https://github.com/jubos/fake-s3
- using a real s3 bucket, and we can put the AWS keys in Travis-ci env (on the Travis-ci website so they remain private)

And we make the test run only if an s3 connection is available so it doesn't prevent other developers from running the test if they don't have a bucket. (Just like we did with specific image generators tests).

freekmurze:
Putting AWS keys in Travis ci seems like a good solution. 👍 

 I'll take care of this is the coming days.

nicolasbeauvais:
I can write the test class to save you some time if its ok :)

freekmurze:
That certainly is ok! 😄

freekmurze:
We'll handle this further in #381

Issue #381, opened 2016-10-22
Make S3 connection testable


freekmurze:
@nicolasbeauvais 

I've setup an S3 bucket and added it's credentials to travis:

![screen shot 2016-10-22 at 14 36 07](https://cloud.githubusercontent.com/assets/483853/19619350/475e883c-9865-11e6-8265-40624aaeb748.jpg)

With those credentials you can only read write to the bucket specified in `S3_BUCKET_NAME`.

If I understand Travis' documentation on the subject correctly the variables will only be available when running tests on commits and merges originating from our own repo. So if somebody else sends a PR the variable isn't available during the tests of that PR (and that's a good thing, because otherwise you could just echo out the variable and see it in the travis log).

Let me know when you're ready with the S3 test, so I can test this out. If it works correctly I will investigate how your PRs can get access to these variables. Maybe I'll just add you as a collaborator on this repo.

nicolasbeauvais:
What is the region of the bucket ? `eu-west-1` ?

nicolasbeauvais:
To perform the tests I need `flysystem-aws-s3-v3` which require `flysystem-aws-s3-v3` which require `guzzlehttp/guzzle  ^5.3 || ^6.0.1` but `scrutinizer/ocular` already require `guzzlehttp/guzzle 3`, and there is no more recent version of `scrutinizer/ocular`  :/

I saw on https://scrutinizer-ci.com/g/spatie/laravel-medialibrary/ that the code coverage is not enabled, but there is this line in travis.yml:

```yaml
after_script:
  - php vendor/bin/ocular code-coverage:upload --format=php-clover coverage.clover
```

Some light on this ?

freekmurze:
I've added `S3_BUCKET_REGION` with the right region value.

You may just remove `scrutinizer/ocular`. I believe it isn't needed anymore.

freekmurze:
I've added you as a collaborator on this repo. I hope in this way your PRs will have access to the travis environment variables as well.

Because you're now a collaborator (you're the first one ever on any of our public repo's, congrats 😄 ) you can now approve PRs and tag releases as well. Unless it's a simple bugfix could you always consult me or @sebastiandedeyne  before merging changes or tagging releases?

nicolasbeauvais:
test doesn't pass  :/ `ErrorException: file_get_contents(): php_network_getaddresses: getaddrinfo failed: Name or service not known`

[...]

PR #384, opened 2016-10-22
S3 tests

nicolasbeauvais:
As discussed in #381 

I tough it would be a good thing to test the file integrity in https://github.com/spatie/laravel-medialibrary/compare/master...nicolasbeauvais:s3-tests?expand=1#diff-bae027168159c1221823e2acf336b0b0R93

but the s3 bucket needs an open policy so that Travis-ci can read the file with `file_get_content()` so it's commented for now.


nicolasbeauvais:
The teardown method that cleaned all the files uploaded to s3 doesn't work :/ I will investigate tomorrow night !

freekmurze:
The open policy has been set, all stored objects are now publicly accessible.

tests/TestCase.php
```diff
@@ -160,4 +167,31 @@ private function setUpMorphMap()
...
+        $this->canTestS3 = ! (bool) array_search(false, $s3Configuration);
+
+        $app['config']->set('filesystems.disks.s3', $s3Configuration);
+        $app['config']->set(
+            'laravel-medialibrary.s3.domain',
+            'https://'.$s3Configuration['region'].'.amazonaws.com/laravel-medialibrary'
+        );
+
+        if ($this->canTestS3) {
+            register_shutdown_function(function () {
```
nicolasbeauvais: Couldn't find a way to call `Storage::` or `app('filesystem.disk')` in a ̀tearDownAfterCLass` method, using `register_shutdown_function` is the only working way that I found to clean the S3 bucket

nicolasbeauvais:
Everything should work now :+1: But as you said my PR Travis build didn't have access to the S3 env vars

freekmurze:
I'll try out the test by merging your PR!

PR #387, opened 2016-10-25
v3 - Laravel 5.3 + PHP 5

OpusDevTeam:
Very simple fixes to allow v3 branch to be used with Laravel 5.3 + PHP 5 (we cant quite move to php 7 yet).

### DIFF
--- a/src/Conversion/ConversionCollection.php
+++ b/src/Conversion/ConversionCollection.php
@@ -38,7 +38,7 @@ class ConversionCollection extends Collection
      */
     public function getByName($name)
     {
-        $conversion = $this->first(function ($key, Conversion $conversion) use ($name) {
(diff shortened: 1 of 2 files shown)


freekmurze:
Normally we don't offer support for an old major version. But as the code changes are easy I'm willing to make an exception.

Unfortunately this PR breaks compatiblity with L5.1 and L5.2. Please make it so that the package can run in all versions from Laravel 5.1 up.

I think the `first` collection call should be replaced by a regular loop. I'm a bit unsure what the effect of removing `SelfHandling` is in L5.1 and L5.2.

OpusDevTeam:
Yeah its a little tricky. I'm going to cancel the request as I found some additional problems in some of the other files.

Issue #406, opened 2016-12-04
Add custom headers on runtime


arubacao:
Hi,

I'd like to add custom headers depending on file, collection and disks.
In the [config file one can add the custom headers](https://github.com/spatie/laravel-medialibrary/blob/master/config/laravel-medialibrary.php#L56), however they get [appended every time](https://github.com/spatie/laravel-medialibrary/blob/master/src/Filesystem.php#L75-L77).

Simple example were adding custom headers on runtime could be relevant:

- one S3 bucket
- one collection `profile_images` --> should be **publicly** available 
- one collection `private_images` --> should be **private**

One could edit the [config value during runtime](https://laravel.com/docs/5.3/configuration#accessing-configuration-values), but in my opinion that's not an elegant way.

Cheers

freekmurze:
Hi, thanks for that PR. On first sight it looks good, I'll review it later this week.

PR #407, opened 2016-12-04
Add custom remote headers on runtime logic

arubacao:
This is a quick hack that corresponds to #406.

I feel like `customRemoteHeaders` belongs to the `Filesystem` since it gets used there.
However, it should (optionally) get set on each file and the user never gets to set params on the `Filesystem`. Therefore the setter is located within the `FileAdder` which in turn sets the values in `$filesystem`

Example: 
```
$user->addMediaFromUrl('https://docs.spatie.be/images/medialibrary/header.jpg')
                ->setCustomRemoteHeaders([
                    'ACL' => 'public-read',
                ])
                ->toCollectionOnDisk('profile_images', 's3-media');
```

If this approach makes any sense, I'd also add some tests.  
Cheers

### DIFF
--- a/src/FileAdder/FileAdder.php
+++ b/src/FileAdder/FileAdder.php
@@ -235,15 +235,15 @@ class FileAdder
     }
 
     /**
-     * Set additional custom remote headers.
+     * Add additional custom remote headers.
      *
      * @param array $customRemoteHeaders
      *
      * @return $this
      */
-    public function setCustomRemoteHeaders(array $customRemoteHeaders)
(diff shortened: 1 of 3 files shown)


freekmurze:
Hi, thanks for your work on this.

Seems ok, I'll probably pull this in when you add some tests.

Could you rename the method to `addCustomRemoteHeaders`. That sound a bit friendlier. What do you think?

arubacao:
Sure and thank you :)

freekmurze:
Thank you very much for your work on this.

arubacao:
No need to thank me! Keep up the great work ✌️

Issue #405, opened 2016-12-04
Probably a change to increase performance for disks with s3 drivers


arubacao:
According to the [directory structure](https://docs.spatie.be/laravel-medialibrary/v4/advanced-usage/using-a-custom-directory-structure) _"files will be stored inside a **directory** that uses the id of its Media-object as a name."_

I've noticed that the media-library could save up to 50% HTTP calls to S3 (and I guess `rackspace` would also work) when associating a new file to a model.

During the upload of a file the method `getMediaDirectory` gets called which in turn [makes a call to create a directory](https://github.com/spatie/laravel-medialibrary/blob/master/src/Filesystem.php#L131).
This is redundant for s3 because you can upload the file to a non-existing folder and s3 takes care of it.
Supposedly, it's also bad practice to treat object storages like POSIX - but I didn't research that (However, I guess that this would be a tweak for [league/flysystem-aws-s3-v3](https://github.com/thephpleague/flysystem-aws-s3-v3) ). 

Just a hacky snipplet to illustrate my point:
```
    /*
     * Return the directory where all files of the given media are stored.
     */
    public function getMediaDirectory(Media $media, bool $conversion = false) : string
    {
        $pathGenerator = PathGeneratorFactory::create();

        $directory = $conversion ?
            $pathGenerator->getPathForConversions($media) :
            $pathGenerator->getPath($media);

        if(config("filesystems.disks.{$media->disk}.driver") !== 's3') {
            $this->filesystem->disk($media->disk)->makeDirectory($directory);
        }

        return $directory;
    }
```

I'm pretty sure that there is a much better solution to incorporate the idea, but I wanted to hear your thoughts on this first.

Cheers

arubacao:
The [`copyToMediaLibrary`](https://github.com/spatie/laravel-medialibrary/blob/master/src/Filesystem.php#L48) method could also be a good place for this.

freekmurze:
I'd accept a PR to improves S3 performance. `copyToMediaLibrary` seems like a good spot indeed.

arubacao:
PR was merged, so I'm going to close this here.

PR #409, opened 2016-12-05
Feature/reduce s3 http calls

arubacao:
For #405 

After I dug a little bit more, I ended up putting this in the `getMediaDirectory` method after all.

Quick wrap-up:

On various occasions the method `getMediaDirectory` gets called.
In my opinion its main task is to determine the directory path name.
However, it also calls `$this->filesystem->disk($media->disk)->makeDirectory($directory);`. Therefor it guarantees that the directory name returned will physically be a folder on the disk.
This makes sense for the `local` filesystem, but is redundant/non-essential for `s3` disks.
Media/files can get pushed to non-existing "folders" (_i assume they aren't real folders on s3_) on s3 - AWS takes care of everything.

## Pros:

1. Skipping `$this->filesystem->disk($media->disk)->makeDirectory($directory);` for s3 disks reduces up-to 50% HTTP calls to S3 for a file association and thus improves performance.

2. Allegedly making empty directories in S3 or treating object storages as POSIX is bad practice.  (I've heard - cannot prove)

3. I couldn't come up with a scenario where this might be a breaking change. 

## Cons:

1. If the file upload is corrupt, but the preceding `makeDirectory` succeeds, there won't be any empty folders on S3 (I don't even know if this is a con. I was just looking for something 😄)

### Extra Info:

As I mentioned in the issue, I think that this performance improvement also works for `rackspace`. This is the reason why I'm using `in_array`. But I didn't add `rackspace` to the array because I don't have an account an therefore cannot test it.

I've tested it personally and everything worked fine for me.
Each test works on his own, but fail when running in series:

```
1) Spatie\MediaLibrary\Test\FileAdder\S3IntegrationTest::it_store_a_file_on_s3
 ErrorException: unlink(/var/www/html/tests/temp/testfiles/test.jpg): Text file busy

 /var/www/html/src/FileAdder/FileAdder.php:347
 /var/www/html/src/FileAdder/FileAdder.php:265
 /var/www/html/tests/S3Integration/S3IntegrationTest.php:36

 2) Spatie\MediaLibrary\Test\FileAdder\S3IntegrationTest::it_store_a_file_and_its_conversion_on_s3
 ErrorException: mkdir(): File exists

 /var/www/html/vendor/laravel/framework/src/Illuminate/Filesystem/Filesystem.php:337
 /var/www/html/vendor/illuminate/support/Facades/Facade.php:215
 /var/www/html/tests/TestCase.php:120
 /var/www/html/tests/TestCase.php:63
 /var/www/html/vendor/orchestra/testbench/src/Traits/ApplicationTrait.php:101
 /var/www/html/vendor/laravel/framework/src/Illuminate/Foundation/Testing/ApplicationTrait.php:34
 /var/www/html/vendor/orchestra/testbench/src/TestCase.php:43
 /var/www/html/tests/TestCase.php:34
 /var/www/html/tests/S3Integration/S3IntegrationTest.php:13
```

Cheers

### DIFF
--- a/src/Filesystem.php
+++ b/src/Filesystem.php
@@ -128,7 +128,9 @@ class Filesystem
             $pathGenerator->getPathForConversions($media) :
             $pathGenerator->getPath($media);
 
-        $this->filesystem->disk($media->disk)->makeDirectory($directory);
+        if(! in_array($media->getDiskDriverName(), ['s3'], true)) {
+            $this->filesystem->disk($media->disk)->makeDirectory($directory);
+        }
 
         return $directory;
     }
--- a/src/Media.php
+++ b/src/Media.php
@@ -170,7 +170,7 @@ class Media extends Model
 
     public function getDiskDriverName() : string
     {
-        return config("filesystems.disks.{$this->disk}.driver");
+        return strtolower(config("filesystems.disks.{$this->disk}.driver"));
     }
 
     /*


freekmurze:
@nicolasbeauvais Seems fine to me, is this ok for you too?

nicolasbeauvais:
@freekmurze  Sorry for not being very active, I'm completely swamped with work :/

The PR look fine, great find @arubacao !

I do not have the time to test it for real now, maybe at the end of the week :)

freekmurze:
@nicolasbeauvais no worries, though I appreciate all your work on this package, you have zero obligations here whatsoever.

If I find a little bit of time, I'll test it too.

freekmurze:
Thanks!

arubacao:
Thanks @freekmurze and @nicolasbeauvais 👍

PR #411, opened 2016-12-07
Remove realpath() from getStoragePath

Melcus:
Remove realpath() from the return in LocalUrlGenerator::getStoragePath() to avoid creating custom URL's for linked storages


freekmurze:
Going to leave this as it is for now.

PR #419, opened 2016-12-16
add findMedia method to find by id

bryanrscott:


nicolasbeauvais:
That looks like overengineering to me. Aren't
```php
$item->media->find($id);
// and
$media->fresh();
```
simple enough ? 

Is there a special use case for this ?

bryanrscott:
Agree, simpler

PR #423, opened 2016-12-23
Fix HasMediaTrait inconsistency on hasMedia and getMedia

MarceauKa:
Hi guys!

I'm using your awesome library but there's some inconsistencies.

Consider this basic scenario:
```php
$model->addMedia($request->file('picture'))->toCollection();
```

By default, `toCollection` appends the file to the collection named `default`.

If I want to check if the file was added I do:
```php
$model->hasMedia();
```

But, methods `hasMedia` and `getMedia` are checking by default a collection named ` ` and return false.

My pull request fix that for more consistency.

### DIFF
--- a/src/HasMedia/HasMediaTrait.php
+++ b/src/HasMedia/HasMediaTrait.php
@@ -108,7 +108,7 @@ trait HasMediaTrait
     /*
      * Determine if there is media in the given collection.
      */
-    public function hasMedia(string $collectionName = '') : bool
+    public function hasMedia(string $collectionName = 'default') : bool
     {
         return count($this->getMedia($collectionName)) ? true : false;
     }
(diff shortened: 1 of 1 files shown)


freekmurze:
Good catch, could you also update the tests?

freekmurze:
`GetMediaTest` and `HasMediaTest` are still failing.

MarceauKa:
Just tried to update them but it seems there's others problems to solve.
I'm trying to investigate on that...

freekmurze:
Ignore, the S3 tests, I need to fix those myself.

freekmurze:
I've decided to hold this one off for now, as it will break existing installations. I'll revisit this when creating a new major version.

freekmurze:
Thanks, I will take care of the tests.

PR #438, opened 2017-01-10
Nested custom properties

sebastiandedeyne:


src/Media.php
```diff
@@ -204,11 +236,34 @@ public function setCustomProperty(string $name, $value)
...
+
+        $this->custom_properties = $customProperties;
+    }
+
+    /**
      * @param string $name
      *
      * @return $this
+     *
+     * @deprecated Will be renamed to `forgetCustomProperty` in the next
```
freekmurze: This notice may be removed.

src/Media.php
```diff
@@ -221,10 +276,43 @@ public function removeCustomProperty(string $name)
...
+    public function removeCustomProperty(string $name)
+    {
+        return $this->forgetCustomProperty($name);
+    }
+
+    /**
+     * Forget a custom property using dot notation.
+     *
+     * @param string $name
+     * @param mixed $value
```
freekmurze: This line may be removed.

PR #439, opened 2017-01-12
Update LocalUrlGenerator.php

MahmoudAljabary:


freekmurze:
I don't understand why you want this change. Add a failing test to reproduce your bug.

Issue #443, opened 2017-01-14
Why fopen in Filesystem.php require "write" permission?


ethaizone:
https://github.com/spatie/laravel-medialibrary/blob/master/src/Filesystem.php#L61

I use `addMedia()` to add original file from samba and it trigger error "Permission denied". I mount it as folder and have permission as 755 so I think it must be something bad in code then I look at them.

I don't know why in `fopen()` in Filesystem.php need "r+".  It needs to read file only. Can we change it to "r"?

freekmurze:
We'll discuss this further at #444

PR #444, opened 2017-01-14
Fix fopen mode parameter to be read only.

ethaizone:
Ref: https://github.com/spatie/laravel-medialibrary/issues/443


freekmurze:
Have you verified that this change does fix your problem?

freekmurze:
Thanks

ethaizone:
I'm sorry that I don't answer your question. I tested it on my staging server and it work after I patch this PR. It fix my problem. ^^

Thank you for this merge. :D

PR #449, opened 2017-01-18
Avoid creating / deleting temp dir if no conversions should be performed

nicolasbeauvais:
Skip the creation of the temp dir, media copy to temp and temp dir delete if no conversions are found for a media.


freekmurze:
Thank you!

PR #466, opened 2017-01-29
Allow the Filesystem class to be overridden

barryvanveen:
Like some other folks mentioned in the issues, I would like to bypass the restriction of placing a single media object in a single directory. Although placing multiple objects in a single directory is no problem, deleting an object *is* since the whole directory is deleted.

My workaround is quite simple: define a FilesystemInterface and bind that to the service container. Now users can bind their own implementation of the interface to extend or alter the behavior of the default Filesystem class. As far as I can see this has no real downsides or compatibility problems while increasing flexibility.

The Filesystem implementation that should be used can also be configurable, like it is for the Media model already. I have not included this in my PR but I would happily add it.

I'm curious how you look at this @freekmurze. 

Cheers,

Barry

### DIFF
--- a/src/FileAdder/FileAdder.php
+++ b/src/FileAdder/FileAdder.php
@@ -3,9 +3,9 @@
 namespace Spatie\MediaLibrary\FileAdder;
 
 use Spatie\MediaLibrary\Media;
-use Spatie\MediaLibrary\Filesystem;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Contracts\Cache\Repository;
+use Spatie\MediaLibrary\FilesystemInterface;
 use Symfony\Component\HttpFoundation\File\File;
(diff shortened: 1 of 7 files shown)


freekmurze:
Thanks for your work on this.

Although I wouldn't necessarily recommend to the users of the package to override the `Filesystem` methods, this seems like a good way to make the package more flexible for power users.

Thoughts @nicolasbeauvais

nicolasbeauvais:
Great change!

@barryvanveen could you make a PR to https://github.com/spatie/docs.spatie.be to make a short explanation on how one could override the Filesystem ?

barryvanveen:
Glad you like it :)

I've added some documentation and created a PR at https://github.com/spatie/docs.spatie.be/pull/109.

freekmurze:
Awesome PR. Thanks again!

Issue #467, opened 2017-01-30
Rename config file


freekmurze:
We've decided to drop `laravel-` in the name of the config file for all our packages.

I'd accept a PR targeting the `v5` branch that renames the config file from `laravel-medialibrary.php` to `medialibrary.php`.

Please update the tests as well.

LKDevelopment:
I'll make it, just one moment :)

freekmurze:
Thank @LKDevelopment 👍

PR #468, opened 2017-01-30
Remove "laravel-" from all config files and access

LKDevelopment:
As said from @freekmurze in https://github.com/spatie/laravel-medialibrary/issues/467 it removes all laravel- Prefixes from laravel-medialibary.

Just a Question to @freekmurze what would you thing about "auto rename" if the media library runs the first time under v5?


LKDevelopment:
arg wrong branch sorry!

PR #469, opened 2017-01-30
Remove "laravel-" Prefix on Config Files 

LKDevelopment:
As said from @freekmurze in #467 it removes all laravel- Prefixes from laravel-medialibary.

Just a Question to @freekmurze what would you thing about "auto rename" if the media library runs the first time under v5?

(Sorry for the wrong branch in #468 )


freekmurze:
Thank you for this 👍 

And automated rename isn't necessary, there will be lots of other breaking changes when going from `v4` to `v5`.

LKDevelopment:
Okay, glad to help you with your projects. if there are something more just say a word :)

Issue #492, opened 2017-02-02
v5 discussion thread


freekmurze:
@nicolasbeauvais @sebastiandedeyne 

`v5` of the package is nearing completation and I'm planning to tag it early next week. Do you have any feature requests or suggestions? Breaking changes obviously allowed.

nicolasbeauvais:
Here's my Trello column about media library potential features:

#### Possibility to use the same media multiple amounts of time but with only one upload/conversion.
=> use case: When a user is created I attach a default avatar image to it, when my website will get 1000 users I will have 1000 time the same image + conversions. I can hard link the avatar to avoid that but it's easier if all my code (custom avatar or not) use the same API.

#### New "URL" media type
=> use case: A user can have a media gallery with images/videos, a video can be a .mp4 or a youtube / Vimeo link, always to use a single API instead of having to write some custom code to handle URLs. + It can be a good way to deal with larger files that are already hosted, I don't my users to upload a 2gb video, but I don't mind them using one in their gallery if they have hosted it. We would be able to create custom image generator, for youtube it would be an API call that gets the thumbnail of the video and convert it.

#### Vuejs image loader
=> use case: When a user change it's avatar, we can't display it for like 3-5 seconds until it get converted. The vuejs module could operate in two mode, one that request the image every second until it works or one that wait for a web socket event triggered by a `ConversionHasBeenCompleted`. This is clearly outside of this package scope but it was in my list ^^

freekmurze:
Thanks for your suggestions!

> Possibility to use the same media multiple amounts of time but with only one upload/conversion.
I think this should be handled in application logic. What we do in our project is to just use a presenter that tries to get avatar of a user. If the use has none an url to a generic image is returned. Imho there's no need to store a generic image for each user.

> New "URL" media type
I very much like this idea. Have you any thoughts on how to implement it best? If the implentation is simple, I'd like to add it.

> Vuejs image loader
Yea, like you already mentioned, clearly something that should be handled outside of this package.

nicolasbeauvais:
Also, we can remove image types (except the `TYPE_OTHER` maybe ?):
https://github.com/spatie/laravel-medialibrary/blob/master/src/Media.php#L21-L24

As all type corresponding to an image generator no longer use them:
https://github.com/spatie/laravel-medialibrary/blob/master/src/ImageGenerators/BaseGenerator.php#L44

For the URL feature:
[...]

PR #507, opened 2017-02-03
Remove unused static media types

nicolasbeauvais:
As explained in #492 

I miserably failed and pushed this commit to the repo v5 branch instead of my fork, so the git tree as been rewritten to remove it (sorry).


freekmurze:
Thanks!

PR #510, opened 2017-02-06
Review

sebastiandedeyne:


sebastiandedeyne:
One more thing. I disagree with this line:

https://github.com/spatie/laravel-medialibrary/blob/master/src/Conversion/ConversionCollection.php#L74

I'd expect it to be okay for me to conditionally add conversions to models since it's an instance method, and not a static method.

freekmurze:
Thanks, pulling this in, so I can fix the tests quickly myself.

PR #516, opened 2017-02-06
Fix: getFirstMediaUrl() only returns first conversion

bradgenereux:
This fixes ```getFirstMediaUrl()``` from always returning the first defined media conversion.

Here is what was happening:
```
<img src="{{ $item->getFirstMediaUrl('product', 'sm') }}">
// conversions/sm.jpg

<img src="{{ $item->getFirstMediaUrl('product', 'md') }}">
// conversions/sm.jpg

<img src="{{ $item->getFirstMediaUrl('product', 'lg') }}">
// conversions/sm.jpg
```

Also, my first ever pull request. Hope I did okay!

### DIFF
--- a/src/Conversion/ConversionCollection.php
+++ b/src/Conversion/ConversionCollection.php
@@ -49,8 +49,8 @@ class ConversionCollection extends Collection
      */
     public function getByName(string $name)
     {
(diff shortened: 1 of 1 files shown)


freekmurze:
Thank you very much!

Issue #521, opened 2017-02-08
Postgres?


patrickguevara:
Docs say that because this package uses `json` column types, it's only compatible with MySQL 5.7+. 

Looking at the Laravel Schema Grammars, Postgres makes the distinction between `json` and `jsonb` as column types whereas MySQL uses the `json` column type for both `json()` and `jsonb()` functions from migration files. This is because MySQL stores JSON internally as BSON where Postgres allows both JSON and BSON.

Is it safe to assume that publishing the migrations file and making the following change...

```
$table->json('manipulations');
$table->json('custom_properties');
```

```
$table->jsonb('manipulations');
$table->jsonb('custom_properties');
```

... can make this package compatible with Postgres?

If so, is it worth making a note of this and/or creating a separate migrations file to publish if you use Postgres?

freekmurze:
I don't use postgres, so I haven't needed or tried compatibility with postgres.

I'm not 100 % sure if your changes will work. To get an indication wether the package works with `jsonb` fields, you could try running the testsuite on a postgres db. You can change the db configuration in `TestCase.php` and run the tests with `vendor/bin/phpunit run`.

If it all works for you, I'd accept a PR that adds a note in the readme.

patrickguevara:
Looks like it works with Postgres using either `json` or `jsonb`. This is of course, without the S3 tests.

One note, however, is the assumptions made about column data types and query results order for a Sqlite database. **Because of this, I had to modify the tests in order for them to pass when connected to a Postgres database instance**. 

For instance, `HasMediaTrait/GetMediaTest.php` asserts that `it_will_return_preloaded_media_sorting_on_order_column`. This assumes that `order_column` value returned from the database will be wrapped in single quotes. This is true for Sqlite but not true for Postgres where it comes out as an integer (as PHP understands it). This makes the test fail when using Postgres. I fixed it with a `->map()` that typecasts the results to an `int`. 

In the `HasMedia/UpdateMediaTest.php` the assertion for `it_removes_a_media_item_if_its_not_in_the_update_array` fails because the wrong media item was removed in the test setup. This is based on the assumption that the insert order is preserved when fetching. This is true for a Sqlite database, but not for Postgres. I modified the test setup to use the `scopeOrdered` trait for ensure that we're removing the intended item.

[...]

PR #522, opened 2017-02-09
Test Postgres compatibility

patrickguevara:
#521

### DIFF
--- a/tests/HasMediaTrait/GetMediaTest.php
+++ b/tests/HasMediaTrait/GetMediaTest.php
@@ -211,9 +211,11 @@ class GetMediaTest extends TestCase
             ->first();
 
         $this->assertSame([
-            1 => '1',
-            2 => '2',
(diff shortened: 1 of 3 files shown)


freekmurze:
Thank you for this, but I'm not going to pull it in.

Reasons:
- I don't want commented code (not even in the tests)
- all test should be able to run on Travis
- I myself don't need Postgresql compatibility

If you do want to add tests for Postgresql, maybe put some dedictated tests in a seperate file. Make sure they can run on Travis (I don't know how to do this properly in conjunction with installing Mysql, so you'll have to research that on your own if you want this).

patrickguevara:
Thanks! 

But just to be clear, the commented code was only to show the settings I used for the postgres database. That can easily be removed. 

The tests *do* run on Travis. The only modifications were so that the assertions would play nicely with a postgres database.

I understand you don't need postgres compatibility, but these changes were only meant to harden the tests to make the assertions more reliable.

Anyway, I appreciate you reviewing it! and I get why you aren't merging it in 👍

Issue #525, opened 2017-02-17
Uploading PDFs


patrickguevara:
I'm having trouble uploading PDFs, and I've searched through some issues but haven't found anything that solves the problem.

I'm uploading to an S3 bucket and in my test I can upload a jpeg with no issue, but when I change the file to a PDF, I get...

```
Tests\Features\MediaControllerTest::test_can_add_media_item_to_default_collection_for_team
Symfony\Component\Debug\Exception\FatalThrowableError: Call to undefined method Spatie\MediaLibrary\UrlGenerator\S3UrlGenerator::getPath()
```

I'm not using a custom URL or Path generator, just what comes with the package. 

From my understanding, I can trace it down to a particular issue within FileManipulator.

Here's the trace, as I follow it.

```
HasMediaTrait : addMedia / addMediaFromRequest
FileAdderFactory : create / createFromRequest
FileAdder : setFile
FileAdder : toMediaLibrary
DefaultFilesystem : add
FileManipulator : createDerivedFiles
ConversionCollection : getNonQueuedConversions
FileManipulator : performConversions
FileManipulator : determineImageGenerator
ImageGenerator : canConvert
Media/HasMediaTrait : getPath
```

The call to `FileManipulator::performConversions` is what caused the issue. I'm not defining any sort of conversion for the PDFs and yet we're still diving into that function. 

Making this change solved the problem for me.
```php
// FileManipulator.php

public function createDerivedFiles(Media $media)
    {
        $profileCollection = ConversionCollection::createForMedia($media);

        $nonQueuedConversions = $profileCollection->getNonQueuedConversions($media->collection_name);

        if ($nonQueuedConversions->isNotEmpty()) {
            $this->performConversions($nonQueuedConversions, $media);
        }

        $queuedConversions = $profileCollection->getQueuedConversions($media->collection_name);

        if ($queuedConversions->isNotEmpty()) {
            $this->dispatchQueuedConversions($media, $queuedConversions);
        }
    }
```

This at least prevents the call to `$media->getPath()` which allows me to move forward with the tests, but I still haven't found anything that addresses the S3UrlGenerator's lack of `getPath`.

If the version numbers help...
```
spatie/image 1.0.0
spatie/laravel-medialibrary 5.0.2
spatie/pdf-to-image 1.2.2
spatie/string 2.2.1
spatie/temporary-directory 1.1.2
```

freekmurze:
Hi,

[...]

PR #526, opened 2017-02-17
Check for NonQueued conversions first

patrickguevara:
Check for NonQueued conversions before calling `performConversions`

#525


freekmurze:
An alternative fix is present in `5.1.0`. Let me know if your problem is solved by this.

PR #527, opened 2017-02-17
If curl is available, prefer using curl over fopen

anorgan:
As fopen is usually a bit slower and might even be disabled if [allow_url_fopen](http://php.net/manual/en/filesystem.configuration.php#ini.allow-url-fopen) is set to off, use curl by default. If not available, fallback to fopen

### DIFF
--- a/src/HasMedia/HasMediaTrait.php
+++ b/src/HasMedia/HasMediaTrait.php
@@ -78,12 +78,33 @@ trait HasMediaTrait
      */
     public function addMediaFromUrl(string $url)
     {
-        if (! $stream = @fopen($url, 'r')) {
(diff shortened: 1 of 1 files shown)


freekmurze:
Thank you for your PR.

No users have reported problems with the current implementation and it works pretty good for me as well.  As the current implementation is the most simple one, I'm going to leave it as it is for now.

gotroch:
I'm behind a proxy and fopen doesn't work for me. I need to use curl with proxy options CURLOPT_PROXY, CURLOPT_PROXYPORT  and @anorgan helps to sort it out. This is usefull for me. Thanks.

PR #528, opened 2017-02-18
Add getFullUrl method

faustbrian:
Just a small helper method which could be used instead of having to do `{{ url($post->images[0]->getUrl()) }}` you could do `{{ $post->images[0]->getFullUrl() }}`.

I am currently building an API which returns full urls to media objects and see myself constantly doing `url($image->getUrl())` in Transformers and Services. Think it would be more consistent to have a `getUrl` and `getFullUrl` method to handle both cases where you want the relative or full url.

### DIFF
--- a/src/Media.php
+++ b/src/Media.php
@@ -38,6 +38,20 @@ class Media extends Model
     }
 
     /**
+     * Get the full url to a original media file.
+     *
+     * @param string $conversionName
+     *
+     * @return string
+     *
+     * @throws \Spatie\MediaLibrary\Exceptions\InvalidConversion
+     */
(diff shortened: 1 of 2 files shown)


freekmurze:
Thanks! Could you add a test too?

faustbrian:
I quickly added a test which I suppose should work, just closed my mcbook so had to add the test in the browser and couldn't test it on my local machine.

If it fails I will send a fix for the test tomorrow.

freekmurze:
Thanks, I'll polish it a bit myself

freekmurze:
Released as `5.1.0`

Issue #238, opened 2016-06-05
Add base64 support for `addMedia`


taliptako:

freekmurze:
Please provide some info on what you'd expect the medialibrary to do. If possible add some code that you'd like to work and state why you need this.

taliptako:
I'm using cropit in my project [](https://github.com/scottcheng/cropit) and with this plugin users pick a photo and crop it.
This plugin create base64 code for this image also other popular similiar tools do the same.
So i should save this image to my media library easily but i cant, only i can do is decode this code and after save it with file_put_contents function

a-komarev:
Here is my quick solution for this.

```php
/**
 * Get file content from base64 string.
 *
 * @param $data
 * @return string
 */
private function getBase64Content($data)
{
    list($type, $data) = explode(';', $data);
    list(, $data) = explode(',', $data);
    $data = base64_decode($data);

    return $data;
}

/**
 * Saves temporary file from base64 string.
 *
 * @param $filename
 * @param $data
 * @return File
 */
private function saveFileFromBase64($filename, $data)
{
    $data = $this->getBase64Content($data);
    file_put_contents($filename, $data);
    $file = new File($filename);

    return $file;
}
```

freekmurze:
I'd accept a PR that adds a `addMediaFromBase64` to the package. You can some cues from the [`addMediaFromUrl`](https://github.com/spatie/laravel-medialibrary/blob/b62a50f6c629a4c00ef558d6fc459b001639a38b/src/HasMedia/HasMediaTrait.php#L80) method.

a-komarev:
it should be in `addMediaFromUrl` because v3 don't have `addMediaFromRequest` method or there are any other reasons to place it there? Base64 is not an URL and will come from request.

taliptako:
I have problem with picture extensions because before sending my picture to AddMedia i should save it to somewhere and i think i should use temporary folder for this.But i cant save temporary file with extension(not sure).

> $tmpFile = tempnam(sys_get_temp_dir(), 'media-library');
        file_put_contents($tmpFile, $stream);

        $filename = basename(parse_url($url, PHP_URL_PATH));

This code snippet from addMediaFromUrl function in this example you get extension from url but if i save my picture to temporary folder i cant send you a image path with extension for this media library save pictures without extensions.

We should find best way to do this.
And i thought maybe we can use this code in AddMedia i find this code from this package.
[https://github.com/Intervention/image](url)

>     /**
[...]

PR #532, opened 2017-02-19
implement adding a base64 encoded file

fgreinus:
* also supports adding uri data scheme prefixed files

This is my first attempt to fix the issue #238.

Feedback very welcome as it is my first contribution to this project :).

### DIFF
--- a/src/HasMedia/HasMediaTrait.php
+++ b/src/HasMedia/HasMediaTrait.php
@@ -13,6 +13,7 @@ use Spatie\MediaLibrary\Events\CollectionHasBeenCleared;
 use Spatie\MediaLibrary\Exceptions\MediaCannotBeDeleted;
 use Spatie\MediaLibrary\Exceptions\MediaCannotBeUpdated;
 use Spatie\MediaLibrary\Exceptions\FileCannotBeAdded\UnreachableUrl;
+use Spatie\MediaLibrary\Exceptions\FileCannotBeAdded\InvalidBase64Data;
 
 trait HasMediaTrait
 {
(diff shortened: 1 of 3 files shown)


src/HasMedia/HasMediaTrait.php
```diff
@@ -94,6 +95,50 @@ public function addMediaFromUrl(string $url)
...
+     * @param string $data
+     *
+     * @throws InvalidBase64Data
+     * @throws \Spatie\MediaLibrary\Exceptions\FileCannotBeAdded
+     *
+     * @return \Spatie\MediaLibrary\FileAdder\FileAdder
+     */
+    public function addMediaFromBase64(string $data)
+    {
+        if (! is_string($data)) {
```
freekmurze: The scalar type hint on this function will ensure that `$data` will always be a string. Remove this `if` block.

src/HasMedia/HasMediaTrait.php
```diff
@@ -94,6 +95,50 @@ public function addMediaFromUrl(string $url)
...
+     * Add a base64 encoded file to the medialibrary.
+     *
+     * @param string $data
+     *
+     * @throws InvalidBase64Data
+     * @throws \Spatie\MediaLibrary\Exceptions\FileCannotBeAdded
+     *
+     * @return \Spatie\MediaLibrary\FileAdder\FileAdder
+     */
+    public function addMediaFromBase64(string $data)
```
freekmurze: Rename `$data` to `$base64data`

src/HasMedia/HasMediaTrait.php
```diff
@@ -94,6 +95,50 @@ public function addMediaFromUrl(string $url)
...
+
+        // decoding and then reeconding should not change the data
+        if (base64_encode(base64_decode($data)) !== $data) {
+            throw InvalidBase64Data::create();
+        }
+
+        $data = base64_decode($data);
+
+        // temporarily store the decoded data on the filesystem to be able to pass it to the fileAdder
+        $tmpFile = tempnam(sys_get_temp_dir(), 'media-library');
```
freekmurze: Rename `media-library` to `medialibrary` (that's a bit more consistent with the reset of the code)

tests/FileAdder/IntegrationTest.php
```diff
@@ -336,4 +337,43 @@ public function it_will_throw_an_exception_when_adding_a_file_to_a_non_existing_
...
+            ->addMediaFromBase64($testBase64Data)
+            ->toMediaLibrary();
+
+        $this->assertFileExists($this->getMediaDirectory($media->id.'/'.$media->file_name));
+    }
+
+    /** @test */
+    public function it_will_throw_an_exception_when_adding_invalid_base64_data()
+    {
+        $this->expectException(InvalidBase64Data::class);
```
freekmurze: Move this line to right before the line starting with ` $this->testModel`

src/HasMedia/HasMediaTrait.php
fgreinus: $base64data or $base64Data (the second seems to be more correct in favour of camel-case)

src/HasMedia/HasMediaTrait.php
fgreinus: First time actually using scalar type hints.. so I did not think about this neat behaviour.. thanks!

src/HasMedia/HasMediaTrait.php
freekmurze: Let's go for `$base64data`. That feels slightly more natural to me.

fgreinus:
I fixed the things you noted in your comments. The failed travis builds seem to be related to an issue within the provisioning.

freekmurze:
Yup, I just restarted the build and now it works fine.

freekmurze:
Thank you very much for this PR.

nelson6e65:
How can I get the URL? Files are stored without extensions when used `addMediaFromBase64()` :sweat_smile:

saqueib:
@nelson6e65 did you manage to get the file extension, I am having same issue, image is uploaded but it missing the extension.

nelson6e65:
I don't remember, @saqueib. :sweat_smile: 

I was able to display the image anyway directly, so I forget about it. Maybe later I'll check that project.

But... Maybe you can set the extension by using the ```->usingFileName(string)``` method. You can check the string for `'data:image/{jpeg|png|svg|etc};base64'` to get the extension and pass it to the method with a name.

saqueib:
@nelson6e65 thanks, I also used `->usingFileName(string)` to get around this. thanks

reppair:
Nice one!

PR #537, opened 2017-02-22
Update HasMediaTrait.php

iraklisg:
Update `HasMediaTrait` to handle the case where `getMedia` is called without arguments (`collectionName = 'default'`)

Related to https://github.com/spatie/docs.spatie.be/pull/122


freekmurze:
The tests are failing for this PR.

Issue #534, opened 2017-02-21
Call to undefined method Spatie\MediaLibrary\UrlGenerator\S3UrlGenerator::getPath


sebestenyb:
Having this issue on v5:

```
[Symfony\Component\Debug\Exception\FatalThrowableError]
Call to undefined method Spatie\MediaLibrary\UrlGenerator\S3UrlGenerator::getPath()

Exception trace:
 () at /vendor/spatie/laravel-medialibrary/src/Media.php:81
 Spatie\MediaLibrary\Media->getPath() at /vendor/spatie/laravel-medialibrary/src/ImageGenerators/BaseGenerator.php:24
 Spatie\MediaLibrary\ImageGenerators\BaseGenerator->canConvert() at /vendor/spatie/laravel-medialibrary/src/FileManipulator.php:120
```

A blank new Laravel 5.4 project.

```    
// composer.json
"require": {
        "php": ">=5.6.4",
        "laravel/framework": "5.4.*",
        "laravel/tinker": "~1.0",
        "league/flysystem-aws-s3-v3": "^1.0",
        "spatie/laravel-medialibrary": "^5.0"
    },
```

```
<?php
// User.php
namespace App;

use Illuminate\Notifications\Notifiable;
use Illuminate\Foundation\Auth\User as Authenticatable;
use Spatie\MediaLibrary\HasMedia\HasMediaTrait;
use Spatie\MediaLibrary\HasMedia\Interfaces\HasMedia;

class User extends Authenticatable implements HasMedia
{
    use Notifiable, HasMediaTrait;
```

```
// filesystem.php
        'media' => [
            'driver' => 's3',
            'key' => env('AWS_KEY'),
            'secret' => env('AWS_SECRET'),
            'region' => env('AWS_REGION'),
            'bucket' => env('AWS_BUCKET'),
        ],    
```

```
// tinker

Psy Shell v0.8.1 (PHP 7.0.8 — cli) by Justin Hileman
>>> use App\User
=> null
>>> use Faker\Factory
=> null
>>> $faker = Factory::create()
=> Faker\Generator {#678}
>>> User::first()->addMediaFromUrl($faker->imageUrl())->toMediaLibrary()
PHP error:  Call to undefined method Spatie\MediaLibrary\UrlGenerator\S3UrlGenerator::getPath() in /Users/balazssebesteny/Sites/medialibrary-test/vendor/spatie/laravel-medialibrary/src/Media.php on line 81
```

[...]

PR #543, opened 2017-03-01
Cleaned up canConvert in BaseGenerator

AlexVanderbist:
Since the MIME type is stored on the `Media` model in the database we no longer have to get the MIME type from the file path when determining which `ImageGenerator` to use.

This solves issues with S3 storage since it doesn't have a `S3UrlGenerator::getPath`.

I've also added `fclose` to `DefaultFileSystem::copyToMediaLibrary` to avoid 'text file busy' errors when testing. (Same issue as https://github.com/spatie/laravel-backup/issues/373#issuecomment-281633489)

### DIFF
--- a/src/Filesystem/DefaultFilesystem.php
+++ b/src/Filesystem/DefaultFilesystem.php
@@ -42,10 +42,14 @@ class DefaultFilesystem implements Filesystem
         $destination = $this->getMediaDirectory($media, $conversions).
             ($targetFileName == '' ? pathinfo($file, PATHINFO_BASENAME) : $targetFileName);
 
+        $filePointer = fopen($file, 'r');
+
         if ($media->getDiskDriverName() === 'local') {
             $this->filesystem
                 ->disk($media->disk)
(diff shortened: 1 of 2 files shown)


AlexVanderbist:
This should fix #534

freekmurze:
Cool thanks! Will pull and tag this as soon as Travis has finished.

JoelGross:
I am still having the exact same issue:

`vagrant@homestead:~/Code/recruiting$ php artisan backup:run
Starting backup...
Dumping database recruiting7...
Determining files to backup...
Zipping 6305 files...
Created zip containing 6305 files. Size is 196.01 MB
Copying zip to disk named s3...
Successfully copied zip to disk named s3.
Backup failed because: unlink(/home/vagrant/Code/recruiting/storage/app/backup-t
emp/temp/2017-10-15-10-53-01.zip): Text file busy.
`

Running PHP 7, Laravel 5.5 and Spatie backup:  "name": "spatie/laravel-backup",
            "version": "5.0.5",

PR #544, opened 2017-03-01
Fixes running multiple workers processing image manipulations

emielmolenaar:
We are testing this (pretty awesome) library in an environment where we have multiple queue workers running to process tasks, and also to process image manipulations.

In the current situation, the whole temporary directory used by FileManipulator is being trashed after all manipulations on one file are processed. This causes trouble (files not found, directories not found, etc.) when the same temporary directory is currently in use by one or more other workers.

This pull request covers two problems:

1. Allow the temporary image conversion directory to be configurable. Our workers are low-spaced boxes and we would like to make sure there is enough space in the temporary directory to work with; so we would like to point the temporary directory to one of our storage boxes.
2. Do not trash the temporary directory completely, as this causes problem when one is running multiple queue workers that process image manipulations. Instead, nicely clean the generated files.

If there are any tests required, I will look into them later.

### DIFF
--- a/config/medialibrary.php
+++ b/config/medialibrary.php
@@ -9,6 +9,12 @@ return [
     'defaultFilesystem' => 'media',
 
     /*
+     * The path where to store temporary files while performing image conversions.
+     * If set to null, storage_path('medialibrary/temp') will be used.
+     */
+    'temp_file_path' => null,
+
+    /*
      * The maximum file size of an item in bytes. Adding a file
      * that is larger will result in an exception.
      */
--- a/src/FileManipulator.php
+++ b/src/FileManipulator.php
@@ -55,7 +55,7 @@ class FileManipulator
             return;
         }
 
-        $temporaryDirectory = new TemporaryDirectory(storage_path('medialibrary/temp'));
+        $temporaryDirectory = new TemporaryDirectory($this->getTemporaryDirectoryPath());
 
         $copiedOriginalFile = app(Filesystem::class)->copyFromMediaLibrary(
             $media,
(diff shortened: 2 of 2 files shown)


freekmurze:
I think the only change that needs to be done here is making the path of the temporary directory configurable.  You could then fill `temp_file_path` with an environment variable that's different for each box.

Thoughts?

emielmolenaar:
That would work with one worker process per box, but not with multiple workers per box as they all use the same environment variables.

freekmurze:
What about suffixing the value of `temp_file_path` with the output of a call to `microtime`?

emielmolenaar:
That could work. But still, in a very rare situation, workers are doing stuff in the same directory and they will be trashing that directory. I think one must prevent that at all times. 

A worker cleaning up it's own files guarantees that it does not trash other workers' files and / or directories. And to be honest, microtime() calls in config files sound a bit hacky to me :wink:

freekmurze:
True, using `microtime` is a bit hacky/crappy 😄 

Is there any other way that you can differentiate between different workings? Is there some other environment variable being set? Something like `working_run_id` ?

emielmolenaar:
Not really. And only if I would have such an ID, that does not guarantee that other people don't have the same problem. 

What keeps you from changing the file-clearing?

Trashing the whole directory is also a solution to the problem, as long as it creates a uniquely-named directory on initialization; something like

`
do {
  $directory = $this->directory . '/' . sha1(uniqid(true));
} while(!is_directory($directory));
`
Or something like that. But this should happen somewhere in the application / library logic and not in config files.

freekmurze:
Ok, I'll pull in this PR and work a little on it myself

freekmurze:
The newly tagged `v5.4.0` should solve your problem.

emielmolenaar:
Thank you!

PR #548, opened 2017-03-06
set configured model class explicitly

peerrabe:
Hello Freek,

thanks for your awesome libraries, i'm having not only this in use. Yes, my postcard is on the way.

With laravel-medialibrary i got a problem with a custom class and the regenerate command. Because it uses always Media instead of configured model class. I didn't find any about in the documentation. Also i want to avoid to override MediaRepository in my own AppServiceProvider.

I derived from tag 4.13.0 because i'm using laravel 5.3...

### DIFF
--- a/src/MediaLibraryServiceProvider.php
+++ b/src/MediaLibraryServiceProvider.php
@@ -36,7 +36,12 @@ class MediaLibraryServiceProvider extends ServiceProvider
     {
         $this->mergeConfigFrom(__DIR__.'/../config/medialibrary.php', 'medialibrary');
 
-        $this->app->singleton(MediaRepository::class);
+        $this->app->singleton(MediaRepository::class, function () {
+            // set class explicitly, otherwise in case of a custom class with custom table name commands will fail
+            $mediaClass = $this->app['config']['medialibrary']['media_model'];
+
+            return new MediaRepository(new $mediaClass);
+        });
 
         $this->app->bind('command.medialibrary:regenerate', RegenerateCommand::class);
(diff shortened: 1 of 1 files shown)


freekmurze:
The tests seem to be failing for this one.

Could you also add a test to make sure that the added functionality works?

peerrabe:
I take a look to the log, it seems there is a default for media class missing. Sorry for that, i run tests only locally, there were only skipped tests. Also a test for the Repository should be possible, to check if it uses the correct injected class.

peerrabe:
Changes i made are working for a derived version of tag 4.13.0. Tests also failing without my changes on master.
It seems a class is missing `Error: Class 'Spatie\Image\Manipulations' not found` 
but indeed, also my changes raising an error because of using config helper returns at this point `NULL` now. Direct access in way of `$mediaClass = $this->app['config']['medialibrary']['media_model'];` will work.

peerrabe:
The PR target was wrong, now i corrected it and split it into two PRs because of renaming config file. A problem still exits, the missing `Spatie\Image\Manipulations` class. Do you know what's happens here?

peerrabe:
Ah sorry for confusion, a composer update did the trick and also tests running now without errors (skipped tests only) locally, hopefully also in travis-ci.

freekmurze:
I already pulled the other PR, this one isn't needed anymore right?

peerrabe:
This change is a little bit different, due to renaming of config file laravel-medialibrary -> medialibrary...

PR #550, opened 2017-03-07
set configured model class explicitly

peerrabe:
This fix is version dependent, because of renaming config file...


peerrabe:
Sorry for PR without tests again, but i didn't found any tests to MediaRepository, so i decided to go the same way. I tested with my current version in use and there it works (only with fix).

freekmurze:
Thank you for this, we'll add a test for this ourselves.

PR #551, opened 2017-03-07
Add addAllMediaFromRequest and addMultipleMediaFromRequest

AlexVanderbist:
This PR adds a `addAllMediaFromRequest` and `addMultipleMediaFromRequest` method to add multiple/all files from a request to the medialibrary.


src/HasMedia/HasMediaTrait.php
```diff
@@ -69,6 +69,28 @@ public function addMediaFromRequest(string $key)
     }
 
     /**
+     * Add multiple files from a request by keys.
+     *
+     * @param string[] $keys
+     *
+     * @return \Spatie\MediaLibrary\FileAdder\FileAdder[]
+     */
+    public function addMultipleMediaFromRequest(array $keys)
```
sebastiandedeyne: Can't we make `addMediaFromRequest` variadic and allow it to accept arrays instead? Or not allowed for semver?

AlexVanderbist: This was done mainly to have a clear difference between `addMediaFromRequest` returning a `FileAdder` instance and `addMultipleMediaFromRequest` returning an array of `FileAdder`s. 

Personally I'd prefer a variadic method as well but I'm not sure if that's considered a breaking change?

sebastiandedeyne: Didn't notice the return type was different, so yeah makes sense to have a different method. And it would indeed be a breaking change since we'd have to change the signature.

src/FileAdder/FileAdderFactory.php
```diff
@@ -30,10 +30,41 @@ public static function create(Model $subject, $file)
      */
     public static function createFromRequest(Model $subject, string $key)
     {
-        if (! request()->hasFile($key)) {
-            throw RequestDoesNotHaveFile::create($key);
+        return array_values(static::createMultipleFromRequest($subject, [$key]))[0];
```
freekmurze: call to `array_values` can be removed here

freekmurze:
Thanks, will clean it up a little and release it. Could you also send a PR to the docs?

Issue #549, opened 2017-03-07
Ability to hold media in storage folder


a-komarev:
To allow zero down-time deployment [storage:link](https://laravel.com/docs/5.3/filesystem#the-public-disk) command could be used. But after it medialibrary stop working.

There is an error:

> The storage path is not part of the public path

That issue raised a lot earlier. The problem is kinda trivial:

```php
// LocalUrlGenerator.php
/*
 * Get the path where the whole medialibrary is stored.
 */
protected function getStoragePath() : string
{
    $diskRootPath = $this->config->get('filesystems.disks.'.$this->media->disk.'.root');

    return realpath($diskRootPath);
}
```

`realpath` method converts `./public` path to `./storage/public`. Is there any reason to call `realpath`?

I suppose that's pretty common use case and if `realpath` is really required maybe we can just have a clone of the `LocalUrlGenerator` without `realpath` out of the box? For example: `LocalSymlinkUrlGenerator`. Or have a hack in current one :}

AlexVanderbist:
Hi @a-komarev!

Thanks for your input. I've created a PR (#552) that uses the `url` property on the public disk in the `getUrl` method. This should fix your issue without having to change `getStoragePath`!

Feel free to fetch the branch from the PR and try it out. You'll need to make sure that the `url` property on your media disk is set in `config/filesystem.php`. If you're using the default `public` disk it should already be set as follows:
```php
        'public' => [
            'driver' => 'local',
            'root' => storage_path('app/public'),
            'url' => env('APP_URL').'/storage',
            'visibility' => 'public',
        ],
```

a-komarev:
@AlexVanderbist Maybe better to write it this way?

```php
        'public' => [
            'driver' => 'local',
            'root' => storage_path('app/public'),
            'url' => public_path('storage'),
            'visibility' => 'public',
        ],
```

Or it wouldn't work?

AlexVanderbist:
@a-komarev The `public_path()` method would return the directory's path (for example `/Users/alex/Webdev/laravel-medialibrary-demo/public/storage`) instead of an url.

The url parameter is set to `env('APP_URL').'/storage'` by default in Laravel anyways: https://github.com/laravel/laravel/blob/master/config/filesystems.php#L54

a-komarev:
Yes, you are right.

a-komarev:
I will be able check your PR changes today evening.

a-komarev:
Fixed in #552

PR #552, opened 2017-03-07
Use URL property of disk as baseUrl when available

AlexVanderbist:
Use the `url` property of the selected media disk as a base url when available. 

This way symlinked public disks use the correct url instead of the non-public `root` path.


AlexVanderbist:
Should fix #549

a-komarev:
@AlexVanderbist It's working. Thanks!

freekmurze:
Great solution, thanks @AlexVanderbist

PR #555, opened 2017-03-08
code to stop duplicate create_media_table migration

coding-sunshine:
Added code to check if create_media_table migration exist, if so stop publishing duplicate file.


freekmurze:
Thank you!

PR #565, opened 2017-03-12
Update README to v5

qrazi:
- Update examples to be v5 compatible
- Update links pointing to Laravel websites to use Laravel version 5.4


freekmurze:
Thanks!

PR #567, opened 2017-03-15
Use put method on FileSystemAdapter (v4)

### LINKED ISSUE #453: FTP disk, fwrite() expects parameter 2 to be string, resource given
I am trying to store files externally via a ftp disk.
I managed to get through to this error:
```
ErrorException in Ftp.php line 204:
fwrite() expects parameter 2 to be string, resource given
```
Not sure what is passing the resource...

AlexVanderbist:
The `put` method on the `FileSystemAdapter` determines whether to use `putStream` or a regular `put` on the FileSystem driver. This way we don't have to worry about it.

This should fix #453

### DIFF
--- a/src/Filesystem.php
+++ b/src/Filesystem.php
@@ -65,7 +65,6 @@ class Filesystem implements FilesystemInterface
 
         $this->filesystem
             ->disk($media->disk)
-            ->getDriver()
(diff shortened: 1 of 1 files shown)


freekmurze:
Perfect!

PR #572, opened 2017-03-20
Revert #567

nicolasbeauvais:
The #567 PR introduced a bug with laravel 5.3:

https://github.com/laravel/framework/blob/5.3/src/Illuminate/Filesystem/FilesystemAdapter.php#L76

The third parameter of the `put` method should be a string, but `$this->getRemoteHeadersForFile($file)` return an array resulting in a `Array to string conversion` error

### DIFF
--- a/src/Filesystem.php
+++ b/src/Filesystem.php
@@ -65,6 +65,7 @@ class Filesystem implements FilesystemInterface
 
         $this->filesystem
             ->disk($media->disk)
+            ->getDriver()
             ->put($destination, fopen($file, 'r'), $this->getRemoteHeadersForFile($file));
     }
 


freekmurze:
Thanks! Tagged that revert commit

What would be a good solution for #567 @AlexVanderbist ?

nicolasbeauvais:
Maybe just removing the third parameter ?

AlexVanderbist:
My bad, must've missed this when testing...

The third parameter is necessary to set some headers when uploading the file to a remote driver (like S3) so removing that is not an option.

Now I'm looking at it again I'm thinking `putStream()` should be used instead of `put()` since `fopen()` will return a resource (on this line https://github.com/spatie/laravel-medialibrary/blob/v4/src/Filesystem.php#L69). I think this should fix the original #453 issue as well. Any thoughts @nicolasbeauvais ?

nicolasbeauvais:
@AlexVanderbist Looks good to me :)

AlexVanderbist:
I'll test this in a minute, then send a PR 👌

PR #577, opened 2017-03-22
Feature/rename to media library

### LINKED ISSUE #575: Rename `toMediaLibrary`
`toMedialibrary` doesn't feel that good when you pass it a collection.

```php
$this->user->addMediaFromRequest('file')->toMediaLibrary('avatar');
```

We should add a method `toMediaLibraryCollection`  and mark `toMediaLibrary` as deprecated.

fetzi:
Rename `toMediaLibrary` to `toMediaLibraryCollection`.

Fixes #575


freekmurze:
Thank you very much!

PR #584, opened 2017-03-22
Update LocalUrlGenerator

martintern:
when config('medialibrary.local_url_generator_realpath') is set to false, just return $diskRootPath.
this way projects won't need to extend BaseUrlGenerator when using public symlink to files outside public folder.


freekmurze:
Thank you for your PR but not a fan of this. If you need this behaviour just create your own custom UrlGenerator.

robinvalk:
This can be achieved by setting the `url` option for the media setting in the filesystem config. 

See https://github.com/spatie/laravel-medialibrary/issues/549#issuecomment-284697690 for more information.

PR #585, opened 2017-03-23
keepOriginalImageFormat manipulation

AlexVanderbist:
This PR adds a `keepOriginalImageFormat` manipulation which can be used on a `Conversion` to keep the original file type and image format.

For #225


freekmurze:
Great work, thanks!

Issue #576, opened 2017-03-22
Add a method to clear a collection except certain media


sebastiandedeyne:
If a collection needs to be overwritten by one or more `Media` instances, you wouldn't want to clear the collection before adding the media, because if the add would fail, the collection would be empty.

It would be nicer to add one or more media, then clear the collection of all items except the new ones.

Example:

```php
$avatar = $user->addMediaFromRequest('file')->toMediaLibrary('avatar');
$user->clearMediaCollection('avatar', [$avatar]);
```

Don't think we can change the signature because that would be breaking, gonna need a new method for this.

AlexVanderbist:
Maybe something like `clearMediaCollectionExcept()` or `clearMediaCollectionPreserving()`?

Is the change considered breaking if the `$excluded` argument would be optional and null by default?

sebastiandedeyne:
Yeah PHP doesn't seem to complain about that. https://3v4l.org/VZAog

If we'd go for a different method, `clearMediaCollectionExcept` sounds better since we already use the term `preserving` for "don't delete the files".

freekmurze:
I'd go for a different method instead of tacking on extra arguments. `clearMediaCollectionExcept` sounds fine to me. 👍

AlexVanderbist:
Should the `CollectionHasBeenCleared` event still be fired if the collection was only partly cleared?

AlexVanderbist:
Also; doesn't the `removeMediaItemsNotPresentInArray` method do exactly what the issue describes?

sebastiandedeyne:
> Should the CollectionHasBeenCleared event still be fired if the collection was only partly cleared?

My gut feeling says no /cc @freekmurze 

> Also; doesn't the removeMediaItemsNotPresentInArray method do exactly what the issue describes?

This method is protected and tailored for `updateMedia`, which updates media based on a raw array, not an array of media items.

Sounds like `clearMediaCollectionExcept` is the way to go!

freekmurze:
> Should the CollectionHasBeenCleared event still be fired if the collection was only partly cleared?

Indeed no

PR #586, opened 2017-03-23
Add clearMediaCollectionExcept method

AlexVanderbist:
For #576

### DIFF
--- a/src/HasMedia/HasMediaTrait.php
+++ b/src/HasMedia/HasMediaTrait.php
@@ -315,6 +315,38 @@ trait HasMediaTrait
     }
 
     /**
+     * Remove all media in the given collection except some.
+     *
+     * @param string $collectionName
+     * @param \Spatie\MediaLibrary\Media[]|\Illuminate\Support\Collection $excludedMedia
+     *
+     * @return $this
+     */
+    public function clearMediaCollectionExcept(string $collectionName = 'default', $excludedMedia = [])
+    {
+        $excludedMedia = collect($excludedMedia);
+
+        if ($excludedMedia->isEmpty()) {
+            return $this->clearMediaCollection($collectionName);
+        }
+
+        $this->getMedia($collectionName)
+            ->reject(function (Media $media) use ($excludedMedia) {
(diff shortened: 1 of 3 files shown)


src/HasMedia/HasMediaTrait.php
```diff
@@ -315,6 +315,38 @@ public function clearMediaCollection(string $collectionName = 'default')
...
+        if ($excludedMedia->isEmpty()) {
+            return $this->clearMediaCollection($collectionName);
+        }
+
+        $this->getMedia($collectionName)
+            ->reject(function (Media $media) use ($excludedMedia) {
+                return $excludedMedia->where('id', $media->id)->count();
+            })
+            ->each(function (Media $media) {
+                app(Filesystem::class)->removeFiles($media);
```
freekmurze: Is it necessary to remove files manually here? I'm think the `MediaObserver` should take care of that.

tests/HasMediaConversionsTrait/DeleteMediaTest.php
```diff
@@ -39,6 +39,21 @@ public function it_can_clear_a_collection()
     }
 
     /** @test */
+    public function it_can_clear_a_collection_excluding_some_media()
```
freekmurze: Test feels a bit brittle. It only asserts that the number of items is correct. It would be better to check if the ids of excluded items are still present after using `clearMediaCollectionExcept`

AlexVanderbist:
Updated the test and removed the unnecessary `removeFiles()`

src/HasMedia/HasMediaTrait.php
```diff
@@ -334,10 +331,7 @@ public function clearMediaCollectionExcept(string $collectionName = 'default', $
             ->reject(function (Media $media) use ($excludedMedia) {
                 return $excludedMedia->where('id', $media->id)->count();
             })
-            ->each(function (Media $media) {
-                app(Filesystem::class)->removeFiles($media);
-                $media->delete();
-            });
+            ->each->delete();
```
freekmurze: nice 👍

freekmurze:
Well done, thank you!

PR #593, opened 2017-03-29
Update Filesystem.php

Butochnikov:


freekmurze:
Does this currently give an error?

Butochnikov:
yes, if I use `CustomPathGenerator`:
```php
public function getPath(Media $media) :string
    {       
        return $media->collection_name . '/' . $media->id .'/';
    }
    public function getPathForConversions(Media $media) :string
    {      
        return $media->collection_name . '/' . $media->id .'/c/';
    }
```
`Path` will be removed the first and `PathForConversions` doesn't exist any more.

PR #595, opened 2017-03-30
Fix bug where `getResultExtension` would ignore `keepOriginalImageFormat`

### LINKED ISSUE #591: getUrl('conversion') and getPath('conversion') return a jpg when using keepOriginalImageFormat()
model
```php
    public function registerMediaConversions()
    {
            ->fit(Manipulations::FIT_CROP, 150, 150)
            ->keepOriginalImageFormat();
    }
``` 

When uploading a png conversions are made and result in  png's but the path and url methods are wrong for conversions

`$media->getUrl(); //  returns .../public/media/1/test.png` OK 
`$media->getUrl('thumb'); //  returns .../public/media/1/conversions/thumb.jpg` KO

I looked into it but can't seem to find where it goes wrong

AlexVanderbist:
Fixes #591

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to `laravel-medialibrary` will be documented in this file
 
+## 5.8.1 - 2017-03-30
+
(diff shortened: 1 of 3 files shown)


freekmurze:
👍

PR #597, opened 2017-03-31
Media is being returned in the wrong order

langeuh:
the function `mediaIsPreloaded` checks if the media is set, but at the same time retrieves it. This results in the parts of `loadMedia`not being used and the data never being sorted.

This is an attempt to correct this, but probably needs a better solution. 
I just want to show the problem. I detected the problem because the retrieved data did not follow the media's order.


freekmurze:
Could you open up another PR that contains a failing test that demonstrates the bug?

PR #598, opened 2017-03-31
HasMediaTrait->mediaIsPreloaded is incorrect

langeuh:
HasMediaTrait->mediaIsPreloaded  always returns true and fetches the media itself

see #597 
order seems to be ok though


freekmurze:
This is fixed in the latest version of the package.

PR #601, opened 2017-04-03
Fix bug where `mediaIsPreloaded` always returned true

AlexVanderbist:
Once the `media` relationship is lazy loaded it is cached on the model by default. This means that calling `loadMedia` a second time doesn't execute any DB queries (tested in `it_will_cache_loaded_media`).

Fixes the problem described in #598


freekmurze:
Great solution, thanks!

PR #607, opened 2017-04-08
Added ability to addMedia for non existing model

ModernArtisan:
Now you can attach medias for non existing model records.
Can provide some improvements, what do you think?

### DIFF
--- a/src/FileAdder/FileAdder.php
+++ b/src/FileAdder/FileAdder.php
@@ -247,10 +247,6 @@ class FileAdder
      */
     public function toMediaCollection(string $collectionName = 'default', string $diskName = '')
     {
-        if (! $this->subject->exists) {
-            throw ModelDoesNotExist::create($this->subject);
-        }
-
         if (! is_file($this->pathToFile)) {
             throw FileDoesNotExist::create($this->pathToFile);
         }
@@ -275,9 +271,7 @@ class FileAdder
 
         $media->fill($this->properties);
 
-        $this->subject->media()->save($media);
-
-        $this->filesystem->add($this->pathToFile, $media, $this->fileName);
+        $this->attachMedia($media);
 
         if (! $this->preserveOriginal) {
             unlink($this->pathToFile);
@@ -330,5 +324,31 @@ class FileAdder
     protected function sanitizeFileName(string $fileName) : string
     {
         return str_replace(['#', '/', '\\'], '-', $fileName);
+    }
+
+    /**
+     * Attach media to the model
+     *
+     * @param $media
+     */
+    protected function attachMedia($media)
+    {
+        if (!$this->subject->exists) {
+            $this->subject->unsavedMedias[] = $media;
+
+            $class = get_class($this->subject);
+
+            $class::created(function ($model) {
+                if (!empty($model->unsavedMedias)) {
+                    foreach ($model->unsavedMedias as $unsavedMedia) {
+                        $model->media()->save($unsavedMedia);
+                        $this->filesystem->add($this->pathToFile, $unsavedMedia, $this->fileName);
+                    }
+                }
+            });
+        } else {
+            $this->subject->media()->save($media);
+            $this->filesystem->add($this->pathToFile, $media, $this->fileName);
+        }
     }
 }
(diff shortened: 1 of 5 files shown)


src/FileAdder/FileAdder.php
```diff
@@ -331,4 +324,30 @@ protected function sanitizeFileName(string $fileName) : string
...
+            $class = get_class($this->subject);
+
+            $class::created(function ($model) {
+                if (! empty($model->unsavedMedias)) {
+                    foreach ($model->unsavedMedias as $unsavedMedia) {
+                        $model->media()->save($unsavedMedia);
+                        $this->filesystem->add($this->pathToFile, $unsavedMedia, $this->fileName);
+                    }
+                }
+            });
```
freekmurze: Add an early return here, so we can get rid of the `else` statement.

src/FileAdder/FileAdder.php
```diff
@@ -331,4 +324,30 @@ protected function sanitizeFileName(string $fileName) : string
     {
         return str_replace(['#', '/', '\\'], '-', $fileName);
     }
+
+    /**
+     * Attach media to the model.
+     *
+     * @param $media
+     */
+    protected function attachMedia($media)
```
freekmurze: Add a `Media` type hint here

src/FileAdder/FileAdder.php
```diff
@@ -331,4 +324,30 @@ protected function sanitizeFileName(string $fileName) : string
     {
         return str_replace(['#', '/', '\\'], '-', $fileName);
     }
+
+    /**
+     * Attach media to the model.
```
freekmurze: Let's lose the docblock: we can typehint the parameter and the function name is clear enough.

src/HasMedia/HasMediaTrait.php
```diff
@@ -23,6 +23,9 @@
     /** @var bool */
     protected $deletePreservingMedia = false;
 
+    /** @var array */
+    public $unsavedMedias = [];
```
freekmurze: rename this to `unsavedMediaItems`

src/FileAdder/FileAdder.php
```diff
@@ -331,4 +324,30 @@ protected function sanitizeFileName(string $fileName) : string
...
+     */
+    protected function attachMedia($media)
+    {
+        if (! $this->subject->exists) {
+            $this->subject->unsavedMedias[] = $media;
+
+            $class = get_class($this->subject);
+
+            $class::created(function ($model) {
+                if (! empty($model->unsavedMedias)) {
```
freekmurze: Refactor this to bail out with a `return` statement if `model->unsavedMedias` is empty. The the actual process loses an indentation level.

src/FileAdder/FileAdder.php
```diff
@@ -331,4 +324,30 @@ protected function sanitizeFileName(string $fileName) : string
...
+                    foreach ($model->unsavedMedias as $unsavedMedia) {
+                        $model->media()->save($unsavedMedia);
+                        $this->filesystem->add($this->pathToFile, $unsavedMedia, $this->fileName);
+                    }
+                }
+            });
+        } else {
+            $this->subject->media()->save($media);
```
freekmurze: These two lines appear twice in this function. Let's extract them to their own function. No sure on a good name here.

ModernArtisan:
@freekmurze Can you please clarify what exactly do you mean? 

> delete all stray media items that are not related to a model

Because if the save method was never called, media model record hasn't be stored into database.

freekmurze:
My train of thought was wrong. You're right, there's just nothing to clean. 👍

src/HasMedia/HasMediaTrait.php
```diff
@@ -411,7 +413,9 @@ protected function mediaIsPreloaded(): bool
      */
     public function loadMedia(string $collectionName)
     {
-        return $this->media
+        $collection = $this->exists ? $this->media : collect($this->unsavedMediaItems);
```
freekmurze: Could you add a testcase that proves this works when the model hasn't been saved yet?

ModernArtisan:
@freekmurze Testcase added, something else?

freekmurze:
Thank you very much for this! 👍

PR #612, opened 2017-04-16
Added progress bar for regenerate command

markvaneijk:
Your welcome!

https://twitter.com/freekmurze/status/853549889623646208

Happy Easter.

### DIFF
--- a/src/Commands/RegenerateCommand.php
+++ b/src/Commands/RegenerateCommand.php
@@ -63,7 +63,7 @@ class RegenerateCommand extends Command
         if (count($this->erroredMediaIds)) {
             $this->warn('The derived files of these media ids could not be regenerated: '.implode(',', $this->erroredMediaIds));
         }
-        
+
         $this->info('All done!');
     }
 
--- a/src/Commands/RegenerateCommand.php
+++ b/src/Commands/RegenerateCommand.php
@@ -63,11 +63,10 @@ class RegenerateCommand extends Command
         if (count($this->errorMessages)) {
             $this->warn('All done, but with some error messages:');
 
-            foreach($this->errorMessages as $mediaId => $message) {
+            foreach ($this->errorMessages as $mediaId => $message) {
                 $this->warn('Media id '.$mediaId.': "'.$message.'"');
             }
-        }
-        else {
+        } else {
             $this->info('All done!');
         }
     }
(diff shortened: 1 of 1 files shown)


src/Commands/RegenerateCommand.php
```diff
@@ -42,21 +42,27 @@ public function handle()
...
 
-        $this->getMediaToBeRegenerated()->each(function (Media $media) {
+        $mediaFiles = $this->getMediaToBeRegenerated();
+
+        $bar = $this->output->createProgressBar($mediaFiles->count());
+
+        $mediaFiles->each(function (Media $media) use ($bar) {
             try {
                 $this->fileManipulator->createDerivedFiles($media);
                 $this->info("Media {$media->id} regenerated");
```
Anahkiasen: This will need to be removed otherwise it'll mess up the output of the progress bar unless I'm mistaken (since you're technically outputting things in the middle of outputting the bar)

src/Commands/RegenerateCommand.php
```diff
@@ -42,21 +42,27 @@ public function handle()
...
+            $bar->advance();
         });
 
+        $bar->finish();
+
         if (count($this->erroredMediaIds)) {
             $this->warn('The derived files of these media ids could not be regenerated: '.implode(',', $this->erroredMediaIds));
         }
-
-        $this->info('All done!');
```
freekmurze: Let's leave this in

src/Commands/RegenerateCommand.php
```diff
@@ -42,21 +42,27 @@ public function handle()
             return;
         }
 
-        $this->getMediaToBeRegenerated()->each(function (Media $media) {
+        $mediaFiles = $this->getMediaToBeRegenerated();
+
+        $bar = $this->output->createProgressBar($mediaFiles->count());
```
freekmurze: Nitpick: rename `$bar` to `$progressBar`

freekmurze:
Thank you for this.

Could you take care of the nitpicks I added and the remark of @Anahkiasen ?

Then it'll be perfect.

markvaneijk:
Yes, I made the simple changes you suggested. But what about the errors messages? Do you want to collect them and display at the end? Or what should be the outcome?

freekmurze:
Yeah, collecting them and displaying them at the end seems a good way to go about it.

freekmurze:
Thank you very much!

PR #617, opened 2017-04-19
Bug fix if is $file not resource

reg2005:


freekmurze:
Could you give an example in which `$file` would be a `resource`?

freekmurze:
Closing due to inactivity.

PR #618, opened 2017-04-21
Default file system config key format changed to the snake_case

ModernArtisan:
```defaultFilesystem``` key changed to ```default_filesystem```, for uniformity all of config keys.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ return [
      * The filesystems on which to store added files and derived images by default. Choose
      * one or more of the filesystems you configured in app/config/filesystems.php
      */
-    'defaultFilesystem' => 'media',
+    'default_filesystem' => 'media',
 
     /*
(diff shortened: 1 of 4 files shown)


sebastiandedeyne:
This is necessary at some point, but it's a breaking change so I'm afraid we can't pull this in yet :/

ModernArtisan:
I have found only 2 calls of this value. Don't see cases which will break any functionality.

sebastiandedeyne:
It would break the package for users that have published the config file and set a different value for `defaultFilesystem`.

freekmurze:
It's breaking for people who have published their config file and upgrade their package. We'll fix this in the next major version.

ModernArtisan:
Yep, it's for next major release. With updated upgrade guide.

freekmurze:
This has been done in the `laravel55` branch.

PR #628, opened 2017-05-06
Change the default file system from media to local

eliurkis:
This is a fix for this error:
```php
DiskDoesNotExist in DiskDoesNotExist.php line 11:
There is no filesystem disk named `media`
```

This is because the default config come with `'defaultFilesystem' => 'media'` but there is no media Filesystem on Laravel, the default Filesystem in Laravel is `local`:
```php
'default' => env('FILESYSTEM_DRIVER', 'local'),
```

### DIFF
--- a/config/medialibrary.php
+++ b/config/medialibrary.php
@@ -6,7 +6,7 @@ return [
      * The filesystems on which to store added files and derived images by default. Choose
      * one or more of the filesystems you configured in app/config/filesystems.php
      */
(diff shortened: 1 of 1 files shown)


freekmurze:
Thank you for this, but I'm not going to merge it.

The installation page clearly states that a `media` disk should be created when installing the package.

https://docs.spatie.be/laravel-medialibrary/v5/installation-setup

PR #635, opened 2017-05-16
Fixing issue with multiple file uploads

mhamlet:
If we have a file input which allows to upload multiple files, the PHP throws the following exception:

```
FatalThrowableError in FileAdderFactory.php line 26:
Type error: Return value of Spatie\MediaLibrary\FileAdder\FileAdderFactory::createFromRequest() must be an instance of Spatie\MediaLibrary\FileAdder\FileAdder, array returned
```
That's because the ``static::createMultipleFromRequest($subject, [$key])`` in case of multiple files returns the following

```
Collection {#294 ▼
  #items: array:1 [▼
    0 => array:2 [▼
      0 => FileAdder {#303 ▶}
      1 => FileAdder {#306 ▶}
    ]
  ]
}
```

Instead of

```
Collection {#295 ▼
  #items: array:2 [▼
    0 => FileAdder {#303 ▶}
    1 => FileAdder {#306 ▶}
  ]
}
```
Collapsing collection will resolve this.

### DIFF
--- a/src/FileAdder/FileAdderFactory.php
+++ b/src/FileAdder/FileAdderFactory.php
@@ -42,7 +42,7 @@ class FileAdderFactory
             return array_map(function ($file) use ($subject) {
                 return static::create($subject, $file);
             }, $files);
-        });
+        })->collapse();
     }
 
     public static function createAllFromRequest(Model $subject): Collection


freekmurze:
Thank you for this, could you add a test to make sure this change works properly?

mhamlet:
Ok, will add tomorrow.

mhamlet:
You already have a test that covers my change - ``it_can_add_handle_file_keys_that_contain_an_array_to_the_medialibrary_from_the_current_request``. 

But it asserts that number of File Adders is 3.

```php
$this->assertCount(3, $fileAdders);
```

Can you confirm that if number of uploaded files is 4, the number of file adders must be 4 too?

mhamlet:
Also, we must use ``flatten``, not ``collapse``. I will change it in my next commit.

freekmurze:
Thanks, I'll fix this myself.

freekmurze:
Mmm, after pulling this in and changin `collapse` to `flatten` the tests fail.

Reverting your change entirely makes all test pass (locally).

Could you whip up a new PR?

PR #648, opened 2017-05-30
Added optional allowed mime types, fixes #645

### LINKED ISSUE #645: addMediaFromUrl mime check
Hi all,
while working with the library I thought about the following use case.

- an user is asked to insert an image or an url to an image;
- he insert an url to a file which has an extensions of an image (jpg, png, other) but is in fact a script or something malevolent;
- we use addMediaFromUrl to save the media;
- the user have succesfully injected a script on our storage disk and we don't even know.

Do you think this can be considered a security bug?
Is it possible to add a parameter to the addMediaFromUrl to specify which mime types are accepted or something similar?
With normal files is easy because Laravel 5 validation understand the mime type from the content (in this library it's taken from the extension, I think it should be changed) and can be filtered, but from a remote URL we cannot check the mime type before saving the media with addMediaFromUrl.

Thoughts?

bhulsman:
Added optional allowed mime types for addMediaFromUrl and addMediaFromBase64, to prevent adding other type of files than intended. Fixes #645.

### DIFF
--- a/src/HasMedia/HasMediaTrait.php
+++ b/src/HasMedia/HasMediaTrait.php
@@ -2,6 +2,7 @@
 
 namespace Spatie\MediaLibrary\HasMedia;
 
+use Illuminate\Http\File;
 use Spatie\MediaLibrary\Media;
 use Illuminate\Support\Collection;
 use Spatie\MediaLibrary\MediaRepository;
@@ -114,9 +115,7 @@ trait HasMediaTrait
         $tmpFile = tempnam(sys_get_temp_dir(), 'media-library');
         file_put_contents($tmpFile, $stream);
 
-        if (! empty($allowedMimeTypes) && Validator::make(['file' => $tmpFile], ['file' => 'mimetypes:'.implode(',', $allowedMimeTypes)])->fails()) {
-            throw MimeTypeNotAllowed::create(mime_content_type($tmpFile), $allowedMimeTypes);
-        }
+        $this->guardAgainstInvalidMimeType($tmpFile, $allowedMimeTypes);
 
         $filename = basename(parse_url($url, PHP_URL_PATH));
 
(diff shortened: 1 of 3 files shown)


freekmurze:
Hi, thank you for this. Instead of rolling our own check, I'd prefer that [Laravels own validation](https://laravel.com/docs/5.4/validation#mimetypes) would be used.

Could you rework this?

bhulsman:
Here you go! Not sure if this is exactly what you had in mind, but just let me know.

src/HasMedia/HasMediaTrait.php
```diff
@@ -111,6 +114,10 @@ public function addMediaFromUrl(string $url)
         $tmpFile = tempnam(sys_get_temp_dir(), 'media-library');
         file_put_contents($tmpFile, $stream);
 
+        if (! empty($allowedMimeTypes) && Validator::make(['file' => $tmpFile], ['file' => 'mimetypes:'.implode(',', $allowedMimeTypes)])->fails()) {
```
freekmurze: Could you put this logic in it's own method? Name it `guardAgainstInvalidMimeType`. Then you can reuse that in `addMediaFromBase64`.

Use early returns in that function instead of using `&&`. So in that function you could do:

```php
if (empty($allowedMimeTypes)) {
   return;
}

freekmurze:
Thanks. It's a step in the right direction, but there's still room for some improvement (see code comment).

bhulsman:
Created a `guardAgainstInvalidMimeType` method as you requested.  Besides that; the validator requires an instance of UploadedFile or File, fixed that too.

Sorry for the hassle!

freekmurze:
This look much better. Thank you. I'll pull it in, take care of some minor nitpicks and tag a new version soon.

PR #672, opened 2017-07-10
Added Media access to registerMediaConversions

DigicreateJens:
This update allows access to the Media Model in the `registerMediaConversions` method. This way the Media's `custom_properties` can be used when registering its conversions.


sebastiandedeyne:
I'd love to pull this, but unfortunately it's a breaking change which we won't be able to introduce until the next major version.

We'll revisit this when we work on v6.

freekmurze:
This has been added to our `laravel55` branch.

Issue #665, opened 2017-06-30
Artisan medialibrary:clean orphan file not working when using custom path


muswanto:
Hai, thank u for usefull package. :100: 
I've an issue when using custom path and custom id. Delete orphan file is not working. This's my configuration:

- Using guid as id for media database record
- Using custom path {year}{month}/{guid} (ex : 20170501/11e73a8029ef95bc976b3389a49dd6ec)

Everything is working well exept clean up command. I want to clean all orphan file but not working, i see in command code, it's not check if using custom path. Thank you very much for your help... :+1:

freekmurze:
Currently cleaning orphaned files only works if you're not using custom paths. 

If you need this to work for custom paths, I'd accept a fully tested, backwards compatible PR.

freekmurze:
Conversation contiunes in #677

PR #677, opened 2017-07-13
Fixed medialibrary:clean command for custom path generators

ModernArtisan:
Related to issue #665 
Custom paths can be with any nesting level. After deleting the directory, we should check for parent directories on empty recursively, and if it's empty - remove them too.
This is also justified for remove a single media model.
I'm tested it with several path generators. For example implementation of **partitioned id** structure:
```php
use Exception;
use Spatie\MediaLibrary\Media;
use Spatie\MediaLibrary\PathGenerator\PathGenerator;

class MediaPathGenerator implements PathGenerator
{
    /**
     * @param Media $media
     * @return string
     * @throws Exception
     */
    public function getPath(Media $media) : string
    {
        $id = $this->ensurePrintable($media->getKey());
        if (is_numeric($id)) {
            return implode('/', str_split(sprintf('%09d', $id), 3)) . '/';
        } elseif (is_string($id)) {
            return implode('/', array_slice(str_split($id, 3), 0, 3)) . '/';
        }

        throw new Exception('Unsupported type for generating correct file path storage');
    }

    /**
     * @param Media $media
     * @return string
     */
    public function getPathForConversions(Media $media) : string
    {
        return $this->getPath($media) . 'conversions/';
    }

    /**
     * Utility method to ensure the input data only contains
     * printable characters. This is especially important when
     * handling non-printable ID's such as binary UUID's.
     *
     * @param mixed $input
     *
     * @return mixed
     */
    protected function ensurePrintable($input)
    {
        if (! is_numeric($input) && ! ctype_print($input)) {
            // Hash the input data with SHA-256 to represent
            // as printable characters, with minimum chances
            // of the uniqueness being lost.
            return hash('sha256', $input);
        }

        return $input;
    }
}
```

### DIFF
--- a/src/Commands/CleanCommand.php
+++ b/src/Commands/CleanCommand.php
@@ -9,9 +9,10 @@ use Spatie\MediaLibrary\FileManipulator;
 use Spatie\MediaLibrary\MediaRepository;
 use Illuminate\Contracts\Filesystem\Factory;
 use Illuminate\Database\Eloquent\Collection;
+use Spatie\MediaLibrary\Filesystem\Filesystem;
 use Spatie\MediaLibrary\Exceptions\FileCannotBeAdded;
 use Spatie\MediaLibrary\Conversion\ConversionCollection;
-use Spatie\MediaLibrary\PathGenerator\BasePathGenerator;
+use Spatie\MediaLibrary\PathGenerator\PathGeneratorFactory;
 
 class CleanCommand extends Command
 {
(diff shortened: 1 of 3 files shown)


src/Commands/CleanCommand.php
```diff
@@ -126,20 +131,26 @@ protected function deleteOrphanedFiles()
...
-            })->each(function (string $directory) use ($diskName) {
-                if (! $this->isDryRun) {
-                    $this->fileSystem->disk($diskName)->deleteDirectory($directory);
-                }
-
-                $this->info("Orphaned media directory `{$directory}` ".($this->isDryRun ? 'found' : 'has been removed'));
-            });
+        $medias = $this->mediaRepository->all();
+
+        $mediaPaths = $medias->map(function ($media) {
```
freekmurze: Do we really need that `$medias` intermediate variable?

src/Commands/CleanCommand.php
```diff
@@ -126,20 +131,26 @@ protected function deleteOrphanedFiles()
...
+            return rtrim($this->pathGenerator->getPath($media), '/');
+        });
+
+        $mediaPaths = $mediaPaths->merge($medias->map(function ($media) {
```
freekmurze: Do we really need that `$mediaPaths` intermediate variable.

I've got nothing against a long collection chain.

src/Commands/CleanCommand.php
```diff
@@ -126,20 +131,26 @@ protected function deleteOrphanedFiles()
...
+            return rtrim($this->pathGenerator->getPathForConversions($media), '/');
+        }));
+
+        $diff = collect($this->fileSystem->disk($diskName)->allDirectories())->diff($mediaPaths);
```
freekmurze: Do we really need that `$diff` intermediate variable?

ModernArtisan:
Of course, we can remove this intermediate variables.
The main goal of my PR - it's to show, how we can resolve current issue.
What exactly do you think we need to test?
How we can avoid the adding a method to an interface in this case?

freekmurze:
> What exactly do you think we need to test?

We need an automated test that shows orphaned files are cleaned up correctly when using a custom path generator

> How we can avoid the adding a method to an interface in this case?

I think we shouldn't avoid it, it's a good solution. But breaking, ...

ModernArtisan:
Ok, I understood you, will start working on this as soon as I have enough time. Thanks!

freekmurze:
Going to close this for now.

PR #680, opened 2017-07-17
Re-applied fix for #326 

GaspariLab:
Bug #326 (which was fixed in commit 1145a204165af499db39b226c8c2bdcbae068d84) is open again, because the `rawurlencode` function was later removed in commit 48ed0b4.

It looks like it was removed by mistake. This causes a problem with filenames that contain spaces.

This pull request puts the `rawurlencode` function back.


freekmurze:
Your change causes the tests to fail. Could you take a look at that?

freekmurze:
Closing due to inactivity

Issue #698, opened 2017-08-03
ErrorException: fclose(): supplied resource is not a valid stream resource


joshforbes:
Hello, I am trying to use this package with Google Cloud storage using the flysystem adapter from superbalist: [https://github.com/Superbalist/laravel-google-cloud-storage](url).  

When I try to add media `$profile->addMedia('test.png')->toMediaCollection();` the media does get persisted but I am left with the following error:

```
1) Test\Account\Integration\CompanyProfileTest::testNew
ErrorException: fclose(): supplied resource is not a valid stream resource

/Users/josh/code/navjobs/vendor/sentry/sentry/lib/Raven/Breadcrumbs/ErrorHandler.php:36
/Users/josh/code/navjobs/vendor/spatie/laravel-medialibrary/src/Filesystem/DefaultFilesystem.php:61
/Users/josh/code/navjobs/vendor/spatie/laravel-medialibrary/src/Filesystem/DefaultFilesystem.php:30
/Users/josh/code/navjobs/vendor/spatie/laravel-medialibrary/src/FileAdder/FileAdder.php:357
/Users/josh/code/navjobs/vendor/spatie/laravel-medialibrary/src/FileAdder/FileAdder.php:345
/Users/josh/code/navjobs/vendor/spatie/laravel-medialibrary/src/FileAdder/FileAdder.php:275
/Users/josh/code/navjobs/tests/Account/Integration/CompanyProfileTest.php:20
```

The issue, as far as I can tell, is that at the point that the DefaultFilesystem calls fclose($file) the file is already closed.  If I dd($file) right before the fclose() I get `Closed resource @2024`. Is this unexpected behaviour that superbalist's google cloud adapter would have already closed the file?

joshforbes:
Looking into this a little more I don't think its anything the superbalist adapter is doing - they simply pass execution off to the official google cloud storage php library:

```
    /**
     * Uploads a file to the Google Cloud Storage service.
     *
     * @param string $path
     * @param string|resource $contents
     * @param Config $config
     *
     * @return array
     */
    protected function upload($path, $contents, Config $config)
    {
        $path = $this->applyPathPrefix($path);

        $options = $this->getOptionsFromConfig($config);
        $options['name'] = $path;

        $object = $this->bucket->upload($contents, $options);

        return $this->normaliseObject($object);
    }
```

[...]

PR #699, opened 2017-08-03
Making sure that the $file is a resource before calling fclose()

joshforbes:
Some flysystem adapters (google cloud storage) may have already closed the file at this point, making $file point to a ‘Closed resource’. If you try to call fclose() on a Closed resource then you will get an exception. This simply makes sure that $file is still a resource before attempting to close it.

This is related to issue #698.


freekmurze:
Thanks!

PR #733, opened 2017-08-29
Remove docblocks in sortable trait

a-komarev:
One more minor fix :smile:

### DIFF
--- a/src/SortableTrait.php
+++ b/src/SortableTrait.php
@@ -44,11 +44,6 @@ trait SortableTrait
         }
     }
 
-    /**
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks for this, but I'll just remove the entire docblock. Function signature is clear enough.

a-komarev:
Okay. I've removed it. And removed docblock from `determineOrderColumnName` because by your logic it's self described too. And resorted methods ordering (protected method `determineOrderColumnName` goes after all public ones).

freekmurze:
I've done this myself in the new major version. Thanks though!

PR #760, opened 2017-09-19
Add temporary URLs for media on S3

### LINKED ISSUE #758: Temporary URL
Using ML 6.0

I'm integrating DigitalOcean Spaces (which is S3 compatible) and am trying to figure out how to do a temporary URL.

My media on DO is set to private, but I would like to issue a temporary URL to authenticated users. How would I do this? I couldn't find anything when searching "temporary" in the docs.

Edit: It might be a nice feature to have a way to determine if a file is public or private when defining media conversions... :)

AlexVanderbist:
Fixes #758

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to `laravel-medialibrary` will be documented in this file
 
+## 6.1.0 - 2017-09-19
+
+- add `getTemporaryUrl` method for media stored on S3
+
 ## 6.0.0 - 2017-08-30
 
 - add compatiblity for Laravel 5.5
--- a/src/UrlGenerator/UrlGenerator.php
+++ b/src/UrlGenerator/UrlGenerator.php
@@ -2,6 +2,7 @@
 
 namespace Spatie\MediaLibrary\UrlGenerator;
 
+use DateTimeInterface;
 use Spatie\MediaLibrary\Media;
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to `laravel-medialibrary` will be documented in this file
 
+## 6.1.0 - 2017-09-19
+
+- add `getTemporaryUrl` method for media stored on S3
```
freekmurze: Bonus points for updating the changelog 👍

freekmurze:
Thanks!

src/UrlGenerator/UrlGenerator.php
```diff
@@ -16,6 +17,16 @@
...
 
     /**
+     * Get the temporary url for the profile of a media item.
+     *
+     * @param \DateTimeInterface $expiration
+     * @param array              $options
+     *
+     * @return string
+     */
+    public function getTemporaryUrl(DateTimeInterface $expiration, array $options = []): string;
```
freekmurze: @AlexVanderbist Did not spot this when reviewing, but adding a method to an interface is a breaking change (see #765).

Could you tag a new bugfix release where this method is removed from the interface. Also add a new issue with a label `revisit for next major version` with a note that we should re-add that method to the interface.

PR #763, opened 2017-09-20
Add ability to predefine order index on creating Media

vaniyokk:
Hi. Thank you for great package, that is very simple to use!

In my project I needed to import a whole lot of photos with existing ordering into the laravel-medialibrary.
Unfortunately I found that there is no existing simple method like usingName() or usingFilename() to set order, I need 2 more lines of code and one more database query, right after Media has been saved to set order_column value.

I've just made a small improvement over current implementation of ordering to solve this problem.
I add 2 methods, usingOrder() and setOrder(). I writed a test to proof that preordering is working fine.

### DIFF
--- a/src/FileAdder/FileAdder.php
+++ b/src/FileAdder/FileAdder.php
@@ -35,6 +35,9 @@ class FileAdder
     /** @var array */
     protected $customProperties = [];
 
+    /** @var int */
+    protected $orderColumnValue;
+
     /** @var string */
     protected $pathToFile;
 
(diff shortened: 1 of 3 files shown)


huglester:
Hello,
this is nice. 

Had similar issue few days ago.
I just imported all photos, and later altered DB using other script.
this will make life easier:)

freekmurze:
Thank you for this, but I'm probably not going to merge it.

I'm thinking that the majority of users doesn't need this.

Can't you just update the `order_column` value after the `Media` object has been created?

vaniyokk:
Maybe you are right at keeping things simple, but for me, default behavior is slightly annoying. It looks like all-around solution with "one more thing forgotten"


```

$media = $property->addMedia($this->tempFile)
            ->usingName($newName)
            ->usingFileName($newName . '.' . $originalFileExtension)
            ->withCustomProperties([
                'width'  => $width,
                'height' => $height,
            ])
            ->toMediaCollection('gallery');

// Oh, really?
$media->order_column = $image->index;
$media->save();

```
vs

```

$property->addMedia($this->tempFile)
            ->usingOrder($image->index)
            ->usingName($newName)
            ->usingFileName($newName . '.' . $originalFileExtension)
            ->withCustomProperties([
                'width'  => $width,
                'height' => $height,
            ])
            ->toMediaCollection('gallery');

```

Someone may need to reverse order of uploaded files or custom ordering etc.
But, it's totally up to you! Thank's for review. =)

freekmurze:
I'm going to pass on this, but thanks for your work on this.

If you need this add it to your own project.

PR #767, opened 2017-09-23
fix env variable usage in travis file

arubacao:
"--prefer-lowest" in travis is never actually used

Small mistake, but big implication for testing

Cheers,
Chris


freekmurze:
Thanks!

PR #775, opened 2017-10-02
Fixed URL-encoding for S3 files

GaspariLab:
S3 files return wrong URLs because they're not URL-encoded. For example:

> example.s3-eu-west-1.amazonaws.com/1234/This is an example.odt

instead of

> example.s3-eu-west-1.amazonaws.com/1234/This%20is%20an%20example.odt

This bug is similar to #326 but it only happens for S3 files.

I moved `rawUrlEncodeFilename()` into BaseUrlGenerator because that method is now used by both LocalUrlGenerator and S3UrlGenerator.


freekmurze:
Thanks!

PR #780, opened 2017-10-07
add compatibility with jenssegers/laravel-mongodb

vesper8:
This change may look trivial but without it it's not possible to retrieve media items assigned to a MongoDB model.

### DIFF
--- a/src/Media.php
+++ b/src/Media.php
@@ -15,6 +15,8 @@ class Media extends Model
     use SortableTrait;
 
     const TYPE_OTHER = 'other';
+    
+    protected $connection = 'mysql';
 
     protected $guarded = [];
 


arubacao:
You don't need a PR for this since you can do this in your own application. 
[https://github.com/spatie/laravel-medialibrary/blob/master/config/medialibrary.php#L26](https://github.com/spatie/laravel-medialibrary/blob/master/config/medialibrary.php#L26)
Create a own model class that extends the original one and edit as you wish. At least this works for me :)

freekmurze:
Thank you, but I'm not going to merge this in. Follow @arubacao's advice to solve the problem in your own app.

elfeffe:
@vesper8 you saved my day!

kamilkozak:
@vesper8 You saved my day too.

PR #792, opened 2017-10-16
lower and space

bmichotte:
This PR change the `sanitizeFileName`. No more `This%20Image.PNG` in our media folders :)

* it replaces spaces with `-`
* it lower case the filename

Test has been changed with this PR

### DIFF
--- a/src/FileAdder/FileAdder.php
+++ b/src/FileAdder/FileAdder.php
@@ -304,7 +304,7 @@ class FileAdder
      */
     protected function sanitizeFileName(string $fileName): string
     {
-        return str_replace(['#', '/', '\\'], '-', $fileName);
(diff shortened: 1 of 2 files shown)


freekmurze:
Although it isn't a breaking change per sé, it might be unwanted behaviour for some users. I'll add this in the next major version of the package (ETA: early-mid 2017)

bmichotte:
Ok, make sense.
Wwhat about adding a new method then ?

```php
$item->addMediaFromRequest('image')
         return strtolower(str_replace(['#', '/', '\\', ' '], '-', $fileName));
     })
     ->toMediaCollection();
```

freekmurze:
That seems pretty nice. Send a PR with tests.

freekmurze:
You'll be happy to know that these changes are now the `v7` branches, which will be released early next year.

PR #794, opened 2017-10-16
Set the fileName of the file using a callable

bmichotte:
Following #792

You can use a callable to sanitize the filename

```php
$item->addMediaFromRequest('image')
         return strtolower(str_replace(['#', '/', '\\', ' '], '-', $fileName));
     })
     ->toMediaCollection();
```

Test included


freekmurze:
Perfect, could you also send a PR to the documentation repo?

bmichotte:
Sure, I will as soon as I'll PR another feature here ;)

PR #798, opened 2017-10-16
Responsable

bmichotte:
This PR implements the `Illuminate\Contracts\Support\Responsable` interface in the `Media` model.

This allows us to do some stuff like this
```php
Route::get('/some/secret/path/{id}', 'SomeController@secret_download')
        ->middleware(['auth', 
            'role:super_secret_role',
            'permission:can download secret images'
         ]); 

class SomeController 
{
  public function secret_download($id)
  {
        return Media::findOrFail($id);
  }
}
```

Tests provided (image and text file)

### DIFF
--- a/src/Media.php
+++ b/src/Media.php
@@ -6,11 +6,12 @@ use DateTimeInterface;
 use Illuminate\Support\Collection;
 use Spatie\MediaLibrary\Helpers\File;
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Contracts\Support\Responsable;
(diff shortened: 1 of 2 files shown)


freekmurze:
Pretty cool stuff man!

Could you also send a PR to the docs repo?

bmichotte:
Sure, but I'm not sure on which page of the doc I should add it

freekmurze:
Create a new page "Responding with media" in the "Advanced usage" section.

freekmurze:
Thanks a lot for this!

Issue #805, opened 2017-10-18
problem when deleting model and using custom path that includes model properties


vesper8:
My custom generator path is made up of properties from the model

it seems that by the time the package tries to delete the attached media, the model itself has already been deleted so I get 

```
    public function getPath(Media $media): string
    {
        $submission = Submission::find($media->model_id);

        $fileFolder = md5($media->id);
        $path = "{$submission->game->slug}/{$submission->id}/{$fileFolder}/";
        return $path;
    }

```

I've confirmed this is the case by trying to output the value of $submission->id above and it is null when doing a forceDelete()

vesper8:
I got around it by caching the related model just before issuing the forceDelete(). With this, the files get deleted and the error is avoided

but this is definitely a bug

I also noticed that the top containing folder is left behind undeleted (the part identified by `{$submission->game->slug}` above)

freekmurze:
@brendt will take a look at this soon

brendt:
The PR above ( #808 ) address the first part of this issue: being able to get the parent model when deleting media.

Regarding the second part; the empty top level directory not being deleted: this happens because only the media files themselves are deleted, and not the whole directory. One could assume that if you're specifying a custom path for media to be saved in, you might want to use that same location to store other data in.

It might be possible to add a check after deleting all media files related to a model: "does the media directory contains any other files?". This check has to be recursive. If all underlying directories are empty, it's safe to completely remove the top level directory for that media file.

I'm not sure that this is responsibility the media library should be concerned with. It seems to me that a full cleanup of related files of a model should happen on the `deleting` or `deleted` event of the model and not from within the media library, used by that model.

@freekmurze what's your opinion?

vesper8:
it's safe to assume that when deleting a model you would want to delete any associated files. However I agree that this shouldn't be the responsibility of this package. You're right, and I'm going to add my own deletion for the top-level folder

freekmurze:
> It seems to me that a full cleanup of related files of a model should happen on the deleting or deleted event of the model and not from within the media library, used by that model.

We all agree here. Ok, we'll handle this further in #808

PR #808, opened 2017-10-19
Deleting media files with custom path generator

brendt:
This PR changes the observed event for deleting media files from `deleted` to `deleting`.

Removing media files before the parent model is removed, allows us to still use that model in a custom path generator, like so:

```php
public function getPath(Media $media): string
{
    $entry = TestModel::find($media->model_id);

    $fileFolder = md5($media->id);
    return "{$entry->id}/{$fileFolder}/";
}
```

This PR address the first part of this issue: #805

### DIFF
--- a/src/HasMedia/HasMediaTrait.php
+++ b/src/HasMedia/HasMediaTrait.php
@@ -32,7 +32,7 @@ trait HasMediaTrait
 
     public static function bootHasMediaTrait()
     {
-        static::deleted(function (HasMedia $entity) {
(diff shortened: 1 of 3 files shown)


brendt:
Another approach to solving this issue:

Caching the media paths when saving media in the database. And making it so that on delete, the path will not be re-generated, but rather be read from the DB.

The current approach seemed less intrusive; but the other solution could also be possible, if needed.

freekmurze:
Seems good, thanks!

PR #825, opened 2017-11-01
Added getFirstTemporyUrl

lalitvj:


freekmurze:
Thanks Lalit! Could you also PR the docs with an example of the new function?

lalitvj:
Cool, will update tonight.

PR #834, opened 2017-11-06
Update comments

juukie:
Just a few changes to make it a bit more readable and changed `app/config/filesystems.php` to `config/filesystems.php`.


freekmurze:
Thanks!

PR #838, opened 2017-11-08
Fix issue where soft deleted files are removed from the filesystem

Airidas10:
Hello,

I discovered what I think is a bug in the package. 

I was using soft deletes together with a custom model with the hope that soft deleted files will not be gone from my file system. However, whether I was using a soft delete or a hard delete, my files were disappearing. 

I then debugged it and realized that deleted function from the MediaObserver is triggered on both types of deletes and calls the removeFiles function, which removes them from my file system.

In this PR, I added the check to see whether the file is soft deleted or not and removeFiles function is now only called on hard deletes. This fix solved the issue in my project.
I have also made sure that all tests are still passing.

Will look forward for your feedback,

Airidas

### DIFF
--- a/src/MediaObserver.php
+++ b/src/MediaObserver.php
@@ -31,6 +31,19 @@ class MediaObserver
 
     public function deleted(Media $media)
     {
-        app(Filesystem::class)->removeFiles($media);
+        $softDeleted = false;
+        
+        if (in_array('Illuminate\\Database\\Eloquent\\SoftDeletes', class_uses($media))){
+            $softDeleted = $this->isSoftDeleted($media);
+        }
+        
+        if(!$softDeleted){
+            app(Filesystem::class)->removeFiles($media);
+        }
+    }
+
+    private function isSoftDeleted(Media $media)
+    {
+        return $media->isDirty($media->getDeletedAtColumn());
     }
 }
--- a/src/MediaObserver.php
+++ b/src/MediaObserver.php
@@ -3,6 +3,7 @@
 namespace Spatie\MediaLibrary;
 
 use Spatie\MediaLibrary\Filesystem\Filesystem;
+use Illuminate\Database\Eloquent\SoftDeletes;
 
 class MediaObserver
 {
@@ -33,16 +34,16 @@ class MediaObserver
     {
         $softDeleted = false;
         
-        if (in_array('Illuminate\\Database\\Eloquent\\SoftDeletes', class_uses($media))){
+        if (in_array(SoftDeletes::class, trait_uses_recursive($media))) {
             $softDeleted = $this->isSoftDeleted($media);
         }
         
-        if(!$softDeleted){
+        if(!$softDeleted) {
             app(Filesystem::class)->removeFiles($media);
         }
     }
 
-    private function isSoftDeleted(Media $media)
+    protected function isSoftDeleted(Media $media)
     {
         return $media->isDirty($media->getDeletedAtColumn());
     }


freekmurze:
Could you add tests to make sure that this works?

src/MediaObserver.php
```diff
@@ -31,6 +31,19 @@ public function updated(Media $media)
...
+        if (in_array('Illuminate\\Database\\Eloquent\\SoftDeletes', class_uses($media))){
+            $softDeleted = $this->isSoftDeleted($media);
+        }
+        
+        if(!$softDeleted){
+            app(Filesystem::class)->removeFiles($media);
+        }
+    }
+
+    private function isSoftDeleted(Media $media)
```
freekmurze: Use `protected`

src/MediaObserver.php
```diff
@@ -31,6 +31,19 @@ public function updated(Media $media)
 
     public function deleted(Media $media)
     {
-        app(Filesystem::class)->removeFiles($media);
+        $softDeleted = false;
+        
+        if (in_array('Illuminate\\Database\\Eloquent\\SoftDeletes', class_uses($media))){
```
freekmurze: Use `::class` syntax

Airidas10:
I pushed the changes you asked for :)

Regarding tests, I am not sure what tests need to be added, as soft deletes already had some tests in the DeleteTest suite (it_will_not_remove_the_files_when_delete_with_softdeletes_is_used and it_will_remove_the_files_when_forcedelete_with_softdeletes_is_used).
What's really confusing for me, is how these tests were passing, even when soft deletes were deleting files. I looked into those tests, but could not figure out why they weren't catching that. Is it because it happens in the Observer?

Any ideas? :)

Thanks!

Airidas10:
Any updates on this issue?

brendt:
First of all, the merge above was by mistake and is reverted, GitHub closed the PR though, but it can probably be re-opened. 

@Airidas10 To further address the issue, I can't seem to reproduce it.

Let's take a look at the original implementation of the MediaObserver:

```php
    public function deleted(Media $media)
    {
        app(Filesystem::class)->removeFiles($media);
    }
```

When a media file is remove, it files are also removed. This behaviour is correct.

The functionality of soft-deleting doesn't happen at this level; it happens when a model using a `Media` item is deleted from the database. When using the soft delete functionality, the `deleted` event isn't triggered at all.

That's what the `it_will_not_remove_the_files_when_delete_with_softdeletes_is_used` test tests for in `DeleteTest`, and seems to work fine.

If files are removed in your project, even when using the `SoftDeletes` trait; maybe something is happening on the project level which is triggering that `deleted` event. Maybe you could share some of your project specific code? The model class would be a good start.

Airidas10:
Thank you for the response! And sorry for not getting back to you quicker, finding myself rather busy these days :)

In my case MediaObserver's deleted was definitely being triggered - I had a Log statement inside. Hence this particular pull request :)  

When I encountered this issue in my project, I started a new Homestead project, just to test this functionality. I created a custom Model named `MediaModel.php`:

```
<?php

namespace App;

use Spatie\MediaLibrary\Media as BaseMedia;
use Spatie\MediaLibrary\HasMedia\HasMediaTrait;
use Spatie\MediaLibrary\HasMedia\Interfaces\HasMedia;
use Illuminate\Database\Eloquent\SoftDeletes;

class MediaModel extends BaseMedia implements HasMedia
{
    use HasMediaTrait;
    use SoftDeletes;

    protected $table = 'media';
    protected $dates = ['deleted_at'];
}
```

Then, in my FileController, after AJAX request comes in and after I do the validation, I do:

```
$file = MediaModel::find($request->file_id);
if($file){
    $file->delete();
    // Further AJAX code to return response omitted...
}
```

I expect the file to stay in my storage folder, but it's gone... The deleted_at timestamp is correctly inserted in the database, proving that it was indeed a soft delete. If I do forceDelete, the behavior is as expected (file is gone, the row in the DB is gone too).

The expected behavior with soft deletes is achieved, if I use the code proposed in this pull request.

This is a great package and I like using it for file management, so I'd like to have this issue fixed, whether it's the fault with the package or whether I am just using it wrong. Please let me know, if there's any other way I could help! :)

Airidas

Airidas10:
Any more info or should I resubmit this PR to be reopened? :)

brendt:
@Airidas10 I'm looking further into it today or tomorrow.

brendt:
@Airidas10 I've tested your case in a clean project, and it seems to work as expected, so at this point I'm unsure why your files are removed. Which version of the medialibrary are you using? Could you maybe start a new project and try it out like I did: create a model with soft delete enables, and an artisan command which saves a model, makes a loopuk and deletes it.

Airidas10:
@brendt Tried that, same results...

That's a step by step on how I tested it using medialibrary 6.6.3.:

I created a fresh Laravel Homestead project;

Pulled in the package, published a config and a migration, added $table->softDeletes() and migrated;

Made a MediaModel model. (code is Identical to the one I posted above);

Updated the config file to let it know that I am using a custom model;
 
Opened a User.php model and added

```
use Spatie\MediaLibrary\HasMedia\HasMediaTrait;
use Spatie\MediaLibrary\HasMedia\Interfaces\HasMedia; 
use App\MediaModel;

class User extends Authenticatable implements HasMedia
{
    use HasMediaTrait;

    [REST EMITTED]
```

I also manually created a user `(INSERT INTO USERS...`;

Then I made this artisan command and ran it: 

```
public function handle()
    {
        $user = User::first();
        $user->addMediaFromUrl('https://www.planwallpaper.com/static/images/desktop-year-of-the-tiger-images-wallpaper.jpg')->toMediaCollection('images');

        $file = MediaModel::first();
        $file->delete();
    }
```

Once again, if I commented out the delete part, I saw the file under /storage/app/public and I saw the database row in the media table.
If I did ->delete(), the Database row has deleted_at with a timestamp, suggesting it was soft deleted. The file was gone.
If I did ->forceDelete(), both the file and the DB entry were gone (as expected).

By the, way, I added a log statement in MediaObserver:

```
public function deleted(Media $media)
    {
        \Log::debug("Deleted triggered");
        app(Filesystem::class)->removeFiles($media);
    }
```

And Every time I was doing a delete (whether soft or hard), "Deleted triggered" was in my Log :)

brendt:
@Airidas10 Could you push your test project + a README on setting up to GitHub?

Airidas10:
@brendt There you go:

https://github.com/Airidas10/FileTest

:)

brendt:
I'll look at this in a few days!

brendt:
@Airidas10 The `SoftDeletes` trait should be added on the model using the `MediaModel`, not on the `MediaModel` itself. So in your case `SoftDeletes` should be added on `User`.

This makes sense because the `MediaModel` itself can be linked to different models using a polymorphic relation.

mwingren:
@brendt - wouldn't putting soft deletes on the linked model only protect the linked model record?

The problem @Airidas10 is seeing involves soft deleting the MediaModel object and having the associated file deleted from the file system.

Airidas10:
@brendt Exactly what @mwingren said! I looked again at the tests you have and seems to me like you delete the model (let's say a user) and then the files, associated to that user, stay.

What I want to do is not touch that user at all. I just want to be able to soft delete a file, and have it stay on the system, so I can hide it from some users, but display to others. 

Can't find a way to do that at the moment... 
If you need a test for that particular use case, I would be happy to make one :)

brendt:
@Airidas10 you're right, feel free to send a new PR with a testcase!

Airidas10:
@brendt I am trying to add some tests for the proposed solution. 

One of the tests I made looks like this:

```
/** @test */
    public function it_will_keep_soft_deleted_files_on_the_system()
    {
        $media = $this->testModelWithSoftDeletes->addMedia($this->getTestJpg())->toMediaCollection('images');

        $file = Media::first();

        $this->assertTrue(File::isDirectory($this->getMediaDirectory($file->id)));

        $file->delete();

        $this->assertTrue(File::isDirectory($this->getMediaDirectory($file->id)));
    }
```
But it fails, as it looks like the media table, used in those tests, does not have a 'deleted_at' attribute (I added it in my project). 

I also have another test:

```
public function it_will_remove_force_deleted_files_from_the_system()
    {
        $media = $this->testModelWithSoftDeletes->addMedia($this->getTestJpg())->toMediaCollection('images');

        $file = Media::first();

        $this->assertTrue(File::isDirectory($this->getMediaDirectory($file->id)));

        $file->forceDelete();

        $this->assertFalse(File::isDirectory($this->getMediaDirectory($file->id)));
    }
```

This one passes. 
So, I cannot figure out how to test this functionality with the current setup... Any ideas?

brendt:
We're going to move this discussion here: https://github.com/spatie/laravel-medialibrary/issues/876

PR #863, opened 2017-11-27
Add singleFile method to media collection class

introwit:
Edit: Done

### DIFF
--- a/src/FileAdder/FileAdder.php
+++ b/src/FileAdder/FileAdder.php
@@ -292,6 +292,16 @@ class FileAdder
 
         $this->attachMedia($media);
 
+        $mediaCollections = $this->subject->mediaCollections;
+
+        if (!empty($mediaCollections)) {
+            foreach ($mediaCollections as $mediaCollection) {
+                if ($mediaCollection->singleFile) {
+                    $this->subject->clearMediaCollectionExcept($mediaCollection->name, $this->subject->getMedia($mediaCollection->name)->last());
+                }
+            }
+        }
+
         return $media;
     }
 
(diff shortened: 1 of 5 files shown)


freekmurze:
Let me know when this is done.

src/FileAdder/FileAdder.php
```diff
@@ -290,7 +290,11 @@ public function toMediaCollection(string $collectionName = 'default', string $di
 
         $media->fill($this->properties);
 
-        $this->attachMedia($media);
+        $newMedia = $this->attachMedia($media);
+
+        if(isset($this->getMediaCollection($collectionName)->singleFile) && $this->getMediaCollection($collectionName)->singleFile) {
```
freekmurze: That `isset` check isn't necessary because the default is `false` right?

src/FileAdder/FileAdder.php
```diff
@@ -348,7 +352,7 @@ protected function attachMedia(Media $media)
             return;
         }
 
-        $this->processMediaItem($this->subject, $media, $this);
+        return $this->processMediaItem($this->subject, $media, $this);
```
freekmurze: The clearing of the media collection should go in process item. Otherwise the media collection could get cleared before a new item is added.

src/FileAdder/FileAdder.php
introwit: The other tests and tests for other collections break since there is no property called `singleFile` defined.

src/FileAdder/FileAdder.php
introwit: Didn't understand this, can you explain a bit more on this?

freekmurze:
Thanks!

PR #864, opened 2017-11-28
Allow clearing entire media collection except a single media instance

introwit:


src/HasMedia/HasMediaTrait.php
```diff
@@ -361,6 +361,10 @@ public function clearMediaCollection(string $collectionName = 'default')
      */
     public function clearMediaCollectionExcept(string $collectionName = 'default', $excludedMedia = [])
     {
+        if ($excludedMedia instanceof Media) {
+            return $this->clearMediaCollectionExceptMedia($collectionName, $excludedMedia);
```
freekmurze: Seems off, doesn't this cause an infinite loop? Instead of recalling the same funtion and returnning lets just pack that single item in a collection so the rest of the function can run.

introwit: That's a different function `clearMediaCollectionExcept ` & `clearMediaCollectionExceptMedia `

tests/HasMediaConversionsTrait/DeleteMediaTest.php
```diff
@@ -38,6 +38,17 @@ public function it_can_clear_a_collection()
         $this->assertCount(0, $this->testModelWithoutMediaConversions->getMedia('images'));
     }
 
+    /** @test */
+    public function it_can_clear_a_collection_excluding_a_single_media()
```
freekmurze: Here we've tested that the excluded media is not deleted, but this test should assert that the rest of the media is in fact being deleted. Just to be sure.

Issue #858, opened 2017-11-21
Feature wishlist for medialibrary v7


freekmurze:
We are currently buillding `v7` of the medialibrary. We'll work on this:

- add Vue components to help with uploading / administering media 
- add a controller to handle requests coming from those Vue components
- add support for responsive images / srcset (see #810)
- uploading stuff directly to S3
- add support for Media Collections, so you can specify how many items and what kind of items are allow in a collection
- general cleanup

Your ideas can help in us making medialibrary better. Feel free to add any feature requests to this issue.

killgt:
Chunked uploading for large assets can be a userful feature

plakhin:
Ability to separate originals and conversions like
./storage/media-library/{modelId}/{originalfilename}.jpg
./public/media/{modelId}/conversion1.jpg
./public/media/{modelId}/conversion2.jpg
Because usually it's better to backup originals only.
It is possible now, but a bit tricky and needs a lot of code.

ryankilf:
Being able to select a focal point for an image would be great for resizing. i.e. a user could get an image to resize around a point in the top left, instead of in the centre, if that was more appropriate.

Christophvh:
- A simplified way of generating a unique filename when adding a file.
- Storing conversions in db , still not sure why manipulations are stored in the db and conversions aren't

palypster:
Ability to create a `private` media collection. All media of this collections are served via Laravel and user can attach some permission logic (i.e. who has the permission to access this file etc.). This is useful when working with files containing personal data (i.e. PDF scan of signed contract full of personal data).

freekmurze:
@palypster you can already add extra info to a media object via custom properties. You can use those custom properties in your gate checks.

amirshawn:
This might be pushing it but optional image cropping in the vue component would be amazing.

ipalaus:
Afaik, there is no easy way to name a file on a disk atm. I think it would be useful for SEO to being able to name files... i.e.: `/media/1/users-freekmurze.jpg`, `/media/1/users-freekmurze_thumbnail.jpg`, `/media/1/users-freekmurze@x2.jpg`...

amirshawn:
@ipalaus I think the function is called usingFileName('name')

I haven't tested it but I know usingName() is used for changing the name.  I'm assuming usingFileName is for changing the file name.

javi-dev:
[...]

PR #874, opened 2017-12-14
[v7] Add the ability to define a CDN domain after x amounts of minutes

tjmartin69:
Original post here - https://github.com/spatie/laravel-medialibrary/issues/858#issuecomment-349457075

### DIFF
--- a/config/medialibrary.php
+++ b/config/medialibrary.php
@@ -48,7 +48,7 @@ return [
          */
         'domain' => 'https://xxxxxxx.s3.amazonaws.com',
 
-        /**
+        /*
          * The cdn domain that should be prepended when generating urls after the given amount of time has passed
          *
          * Sometimes a cdn can take a while to sync across all of the servers.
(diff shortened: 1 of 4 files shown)


freekmurze:
On second thought, couldn't this be achieved this using a custom `UrlGenerator`?

tjmartin69:
It can be, if you use multiple disks like we do you have to create a proxy url generator class that can switch between local and s3.

Maybe a better solution would be to allow adding a custom url generator per disk instead of a global one.

ielhag:
Hello @freekmurze, I saw that this suggestion is not likely the best and the same functionality proposed by @tjmartin69 could be implemented in a custom url Generator. 

As suggested, I created custom AzureUrlGenerator class to take care of the upload, It works for both upload and getUrl from the storage disk, but what I am trying to achieve is to switch between the 2 urls once for upload to azure storage disk and second for getting the url from the CDN but without timing it, since I will be getting the link ready once the upload is completed.

Any suggestions?

PR #879, opened 2017-12-19
Persist soft_deleted files on the system

Airidas10:
This PR contains the MediaObserver extension to support maintaining soft deleted files on the system and unit tests supporting it. In order for tests to pass, deleted_at column was added to media table.


freekmurze:
Thank you for this, but I'm not going to pull it in. Because we don't use soft deletes on our projects I don't want to maintain support for this. 

If you need soft deletes [use your own model](https://docs.spatie.be/laravel-medialibrary/v6/advanced-usage/using-your-own-model) and [override the default filesystem](https://docs.spatie.be/laravel-medialibrary/v6/advanced-usage/overriding-the-default-filesystem-behaviour).

PR #889, opened 2017-12-30
Fix wrong queue for responsive image generation (close #888)

### LINKED ISSUE #888: [v7] Wrong queue for responsive image generation
__**ISSUE UPDATED!**__

It seems that original responsive image generation is queued on the default Laravel queue, not the custom queue specified in medialibrary config. Looks like conversions are sent to the custom queue, but responsive images generation for the original image are sent to the default laravel queue.

Even if we've `withResponsiveImages` on both the original image, and on conversions, it send two queue messages to two different queues at the same time, one for the **correct** custom queue for conversions, and another for the original responsive image on the default laravel queue, strange!

Controller code:
```php
$model->addMediaFromRequest('file')
     ->withResponsiveImages()
     ->toMediaCollection();
```

Model code:
```php
    {
        $this->addMediaConversion('thumb')->withResponsiveImages()->greyscale();
    }
```

Omranic:
This PR should push original responsive image generation to the configured queue.
It should pass all current tests, and should fix #888

### DIFF
--- a/src/FileAdder/FileAdder.php
+++ b/src/FileAdder/FileAdder.php
@@ -397,7 +397,13 @@ class FileAdder
         }
 
         if ($this->generateResponsiveImages && (new ImageGenerator())->canConvert($media)) {
-            dispatch(new GenerateResponsiveImages($media));
+            $job = new GenerateResponsiveImages($media);
+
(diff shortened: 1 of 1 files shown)


freekmurze:
Thank you! 👍

PR #891, opened 2018-01-04
Meaningful default for s3 domain

IlCallo:
Laravel 5.5 suggest to use some keys for env variables needed to configure AWS services, it would be nice to follow their schema and put a meaningful default based on the AWS_BUCKET key.


freekmurze:
Makes sense, thanks!

PR #897, opened 2018-01-09
Pass tests in windows

unfalln:
While adding some functionality to this project, I found the tests were passing under WSL-bash, but not Powershell. I have updated areas of the code that caused issues when running in Windows so that the tests now pass under both WSL-bash and Powershell. My dev environment had no setup for S3, so this code has not been fully tested. Clearly it up to you whether you would want to consider this push :)


freekmurze:
Thank you for this but I'm not going to pull it in.We don't want to do the maintainance in making sure that the test keep working on Windows.

PR #898, opened 2018-01-10
Image generator params

unfalln:
As part of using this library in a project, I felt the need to generate thumbnails of each page in a PDF after they were uploaded. In order to implement this, I have added the ability to dynamically define ImageGenerator parameters that can be called while registering a conversion object. As I had already found that this functionality could be used for the 'extractVideoFrameAtSecond' I also refactored that bit.

### DIFF
--- a/src/Conversion/Conversion.php
+++ b/src/Conversion/Conversion.php
@@ -4,8 +4,8 @@ namespace Spatie\MediaLibrary\Conversion;
 
 use BadMethodCallException;
 use Spatie\Image\Manipulations;
-use Spatie\MediaLibrary\ImageGenerators\ImageGenerator;
 use Illuminate\Support\Collection;
+use Spatie\MediaLibrary\ImageGenerators\ImageGenerator;
 
 /** @mixin \Spatie\Image\Manipulations */
 class Conversion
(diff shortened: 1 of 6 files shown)


freekmurze:
Could you give some examples on how this could can be used?

unfalln:
A quick example of the code I'm currently using:
```php

    {
        $pdfConverter = new \Spatie\PdfToImage\Pdf($media->getPath());

        $pages = $pdfConverter->getNumberOfPages();

        for($page = 0; $page < $pages; $page++) {
            $pageName = $page + 1;
            $this->addMediaConversion("fieldDesigner_page{$pageName}")->page($pageName)->width(900);
        }

        $media->setCustomProperty("pages", $pages);
        $media->save();
    }
```

Note that this means I now have a thumbnail available for each page of the uploaded pdf that I can now iterate and display.

As previously stated, I also refactored the `extractVideoFrameAtSecond` parameter to demonstrate that these Generator-specific parameters can be added with no further intervention required in the Conversion object.

freekmurze:
Because imho not so much users require this functionality and (even though the code is good) the solution feels a bit heavy, I'm going to pass on this for now.

I might change my mind in the future.

oluwajubelo1:
@unfalln , this is what i've been looking for! Can you please create a gist on this. I'll be grateful. Thanks

PR #903, opened 2018-01-25
added option for table name

shiroamada:
addd option for changing table name instead of fix using media as the
modal name


freekmurze:
You can set another table name by using your own `Media` model and set the `$table` property on that model.

shiroamada:
Can show example step how to do it correct way? I am not familiar extend from other modal.
Thank you!

shiroamada:
OK, I found it in the documentation. And added pull request for class name typo error.
Thanks!

PR #906, opened 2018-01-30
update README.md for disks array added extra para

shiroamada:
update README.md for disks array added extra parameter.
```
'url' => env('APP_URL') . '/media',
'visibility' => 'public',
```


freekmurze:
Thanks!

PR #910, opened 2018-02-03
Change the order of deleting directories

Butochnikov:
First need to delete `conversions` directory otherwise the error on the FTP drive.


freekmurze:
Nice one, thanks!

PR #912, opened 2018-02-07
Adds support for passing a filter callable to `getFirstMediaUrl`

### LINKED ISSUE #911: Feature request: filter on getFirstMediaURL()
I am loving this package, thanks people from Spatie!

What would be the best/cleanest way to retrieve the URL of the featured image (image with custom property ['featured' => 1]?

I think it would be nice to have a filter for the getFirstMediaURL() function.

This is what I tried to do: `$post->getFirstMediaURL('images', ['featured' => 1])` with _featured_ a custom property. The error I get is:

`Argument 2 passed to App\School::getFirstMediaUrl() must be of the type string, array given`

I tried the same with `$post->getMedia('images', ['featured' => 1])->first()->getURL('thumb')`, but this does not work neither.

Thanks!

DivineOmega:
This should fix #911, and is my first contribution to this package. :)

Tested via the `it_can_get_the_url_to_first_media_in_a_collection_using_a_filter` method in `tests/HasMediaTrait/GetMediaTest.php`.

### DIFF
--- a/src/HasMedia/HasMediaTrait.php
+++ b/src/HasMedia/HasMediaTrait.php
@@ -217,7 +217,7 @@ trait HasMediaTrait
      *
      * @return Media|null
      */
-    public function getFirstMedia(string $collectionName = 'default', array $filters = [])
+    public function getFirstMedia(string $collectionName = 'default', $filters = [])
     {
         $media = $this->getMedia($collectionName, $filters);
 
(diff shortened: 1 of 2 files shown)


freekmurze:
Thanks for your work on this but in it's current state I'm not going to merge this in. I don't like the fact that the callable is now set as the third parameter.

The functionality should be implemented so that this is possible:

```php
$myCustomFirstMedia = $model->getFirstMedia(function(Media $media) {
     ... //  when returning true here, the given Media will be used and passed to `$myCustomFirstMedia`
});

The current behaviour should also work

$myCustomFirstMedia = $model->getFirstMedia('images');

So the first argument should be either a `callable` or a `string`.  `getFirstMediaUrl` should work the same.

Could you send a new PR for me to review? If you have any questions, do let me know!

PR #923, opened 2018-02-21
Prevent clearMediaCollectionExcept method from deleting all files.

dkristin:
If the method receives only Media object, the "where" condition below can not be met because of there is not a key in a collection, thus it finds nothing and returns empty collection (`count()` = 0). Therefore it deletes all files and does not respect excluded media. It is critical to wrap `$excludedMedia` into an array to create the key and prevent accidental deletion of all files.


freekmurze:
Could you also add a test for this?

freekmurze:
Closing this due to inactivity and merge conflicts.

PR #925, opened 2018-02-22
Optimize responsive images

brendt:
Responsive images weren't optimised yet, also they always used `gd` as driver.


freekmurze:
🎉  Thanks!

PR #938, opened 2018-03-02
Add disk root logic to S3UrlGenerator

peyobr:
Use optional disk root if existing in file system configuration


freekmurze:
Thanks!

peyobr:
Thank you!

PR #940, opened 2018-03-03
Add function withManipulations to FileAdder

JorisDR:
This feature allows to add manipulation parameters to Media in the fileAdder.

### DIFF
--- a/src/FileAdder/FileAdder.php
+++ b/src/FileAdder/FileAdder.php
@@ -35,6 +35,9 @@ class FileAdder
     /** @var array */
     protected $customProperties = [];
 
+    /** @var array */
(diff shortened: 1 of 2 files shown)


src/FileAdder/FileAdder.php
```diff
@@ -191,6 +194,20 @@ public function withCustomProperties(array $customProperties)
         return $this;
     }
 
+    /**
+     * Set the metadata.
```
freekmurze: Can you change this in `Set the manipulations.`?

JorisDR: done

freekmurze:
Thank you!

PR #943, opened 2018-03-03
Feature wildmark manipulations

JorisDR:
If you mark manipulations with ['all' => '...'] they run on all media conversions

### DIFF
--- a/src/Conversion/ConversionCollection.php
+++ b/src/Conversion/ConversionCollection.php
@@ -138,7 +138,7 @@ class ConversionCollection extends Collection
         if ($conversionName === '*') {
             $this->each->addAsFirstManipulations(clone $manipulations);
         } 
-    
+
         optional($this->first(function (Conversion $conversion) use ($conversionName) {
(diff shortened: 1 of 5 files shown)


src/Conversion/ConversionCollection.php
```diff
@@ -135,9 +135,13 @@ public function getQueuedConversions(string $collectionName = ''): self
      */
     protected function addManipulationToConversion(Manipulations $manipulations, string $conversionName)
     {
-        $this->first(function (Conversion $conversion) use ($conversionName) {
+        if ($conversionName === 'all') {
```
freekmurze: Compared to `all`, `*` has a lower change that it used as a collection name. 

Let's use `*` here.

JorisDR:
@freekmurze  
Up to you to decide: right now first the `*`, then the `thumb` is put in front. Meaning the specific `thumb` manipulation goes before the `*`... 
Feel free to change order

freekmurze:
Thanks!

PR #957, opened 2018-03-17
Add getLastMedia

jozeflambrecht:
I rather need the last media of a collection then the first media

I think I picked the wrong branch in my previous commit...


freekmurze:
Thanks for your contribution. Probably I won’t accept this functionality as you can easily use your own media model and put the function on there.

PR #963, opened 2018-03-20
publish static migration name

ahmed-aliraqi:
generate static migration file name to stop duplicate when run `php artisan vendor:publish`


freekmurze:
Going to keep things like they are for now.

PR #971, opened 2018-03-21
Custom headers for conversions

brendt:
This PR addresses #868 

The `DefaultFilesystem` is always re-created, and so looses the custom header information when saving a media item's conversions. This PR fixes that issue by persisting custom header configuration in the media item, and loading it again when saving a conversion.

### DIFF
--- a/src/FileAdder/FileAdder.php
+++ b/src/FileAdder/FileAdder.php
@@ -60,6 +60,9 @@ class FileAdder
     /** @var bool */
     protected $generateResponsiveImages = false;
 
+    /** @var array */
+    protected $customHeaders = [];
+
     /**
      * @param Filesystem $fileSystem
      */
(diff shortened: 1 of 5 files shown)


freekmurze:
> We should also do this for the responsive images that get uploaded to external filesystems.

This is already in order. Great PR, thanks!

omacesc:
@brendt thanks for the work, if you want to check something tell me.

brendt:
@omacesc 7.0.2 fixes the issue!

https://github.com/spatie/laravel-medialibrary/releases/tag/7.0.2

omacesc:
sorry for open this issue, but can you apply this fix to version 6.6? Thanks!

brendt:
It's a problem that needs solving anyway. PHP 7.0 is already in security-fixes only, and will be end of life before 2019: http://php.net/supported-versions.php

If PHP versions is the only reason holding you back, you and/or your company should take the time and upgrade to 7.1 or 7.2. The last one would be the best decision. It will take some time to do, but in the end it'll be beneficial.

Issue #967, opened 2018-03-20
Undefined index: base64svg


hatcham:
Version 7:

If you happen to try and load a page containing a responsive image that has not fully been processed in the queue, by that I mean, processing has begun but has not completed, the following exception is thrown:

Method Illuminate\View\View::__toString() must not throw an exception, caught ErrorException: Undefined index: base64svg (View: /home/forge/tfh-manuals.hatcham.com/vendor/spatie/laravel-medialibrary/resources/views/responsiveImageWithPlaceholder.blade.php)

So before the job has been processed has run I get the normal <img> tag with a single image. After the job has been processed I get the <img> tag with full srcset. This error occurs in between. 

It is a bit of an edge case but has popped up on more than one occasion today.

I'd love to send a PR but I don't know where to start!

Brilliant package by the way, really enjoying using it. Thank you.

brendt:
Hi @hatcham 

This was a more complex issue to write a good test for, so I went ahead and made a PR myself. If you're interested in how the issue is fixed, you can check out the PR: #973

We'll merge this as soon as the review is done.

brendt:
Fixed in 7.0.3.

https://github.com/spatie/laravel-medialibrary/releases/tag/7.0.3

PR #973, opened 2018-03-21
Null check for placeholder svg

brendt:
This PR addresses #967.

The problem lies in the placeholder not being rendered with an asynchronous queue when rendering the image in a blade file. I've added a test which mimics this scenario, without relying on the whole queue setup.

### DIFF
--- a/src/ResponsiveImages/RegisteredResponsiveImages.php
+++ b/src/ResponsiveImages/RegisteredResponsiveImages.php
@@ -50,16 +50,19 @@ class RegisteredResponsiveImages
             })
             ->implode(', ');
 
-        if (config('medialibrary.responsive_images.use_tiny_placeholders')) {
-            $filesSrcset .= ', '.$this->getPlaceholderSvg().' 32w';
+        $mayAddPlaceholderSvg = config('medialibrary.responsive_images.use_tiny_placeholders')
+            && $this->getPlaceholderSvg();
+
+        if ($mayAddPlaceholderSvg) {
+            $filesSrcset .= ', ' . $this->getPlaceholderSvg() . ' 32w';
         }
 
         return $filesSrcset;
     }
 
-    public function getPlaceholderSvg(): string
+    public function getPlaceholderSvg(): ?string
     {
-        return $this->media->responsive_images[$this->generatedFor]['base64svg'];
+        return $this->media->responsive_images[$this->generatedFor]['base64svg'] ?? null;
     }
 
     public function delete()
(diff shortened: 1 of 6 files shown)


src/ResponsiveImages/RegisteredResponsiveImages.php
```diff
@@ -50,16 +50,19 @@ public function getSrcset(): string
             })
             ->implode(', ');
 
-        if (config('medialibrary.responsive_images.use_tiny_placeholders')) {
+        $mayAddPlaceholderSvg = config('medialibrary.responsive_images.use_tiny_placeholders')
```
freekmurze: Rename this to `shouldAddPlaceholderSvg`. That's more in line with similar functions in other packages.

tests/Feature/ResponsiveImages/RegisteredResponsiveImagesTest.php
```diff
@@ -22,4 +23,27 @@ public function it_will_register_generated_responsive_images_in_the_db()
             'test___medialibrary_original_237_195.jpg',
         ], $media->responsive_images['medialibrary_original']['urls']);
     }
+
+    /** @test */
+    public function it_can_render_a_srcset_when_the_base64svg_is_not_rendered_yet()
```
freekmurze: Nice!

brendt:
@freekmurze Care to take a look at it? 

I figured it makes more sense to add an event after every responsive image is generated for a media item instead of an event for every variation, especially with cache clearing in mind.

tests/Feature/ResponsiveImages/ResponsiveImageGeneratorTest.php
```diff
@@ -29,4 +31,17 @@ public function its_conversions_can_have_responsive_images()
 
         $this->assertFileExists($this->getTempDirectory('media/1/responsive-images/test___thumb_50_41.jpg'));
     }
+
+    /** @test */
+    public function it_triggers_an_event_when_the_responsive_images_are_generated()
+    {
+        Event::fake([ResponsiveImagesGenerated::class]);
```
freekmurze: Would the test work with only `Event::fake()`?

brendt: I prefer to keep tests as specific as possible to avoid unforeseen future side-effects. But I've removed the `[]` because it was unnecessary.

hatcham:
Perfect! Thank you again.

PR #975, opened 2018-03-21
Fix responsive images rendering of conversions

simonellensohn:
At the moment the call to `hasResponsiveImages` in `img` only checks if there are any conversions on the current collection, ignoring the given conversion. In which case the rendered output of the conversion would not contain the `srcset` attribute.

This fix checks if there are any responsive images of the given conversion.

### DIFF
--- a/tests/Feature/Media/ToHtmlTest.php
+++ b/tests/Feature/Media/ToHtmlTest.php
@@ -94,6 +94,19 @@ class ToHtmlTest extends TestCase
     }
 
     /** @test */
+    public function it_can_render_itself_with_responsive_images_of_a_conversion_and_a_placeholder()
+    {
+        $media = $this->testModelWithResponsiveImages
+            ->addMedia($this->getTestJpg())
+            ->toMediaCollection();
+
+        $image = $media->refresh()->img('thumb');
+
+        $this->assertEquals(1, substr_count($image, '/media/2/responsive-images/'));
+        $this->assertTrue(str_contains($image, 'data:image/svg+xml;base64,'));
+    }
+
+    /** @test */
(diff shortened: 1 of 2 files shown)


tests/Feature/Media/ToHtmlTest.php
```diff
@@ -93,6 +93,19 @@ public function it_can_render_itself_with_responsive_images_and_a_placeholder()
...
+    /** @test */
+    public function it_can_render_itself_with_responsive_images_of_a_conversion_and_a_placeholder()
+    {
+        $media = $this->testModelWithResponsiveImages
+            ->addMedia($this->getTestJpg())
+            ->toMediaCollection();
+
+        $image = $media->refresh()->img('thumb');
+
+        $this->assertEquals(1, substr_count($image, '/media/2/responsive-images/'));
```
freekmurze: Can we use `assertContains` here?

simonellensohn:
I've replaced `assertEquals` with `assertContains`, feels a lot cleaner now.

tests/Feature/Media/ToHtmlTest.php
```diff
@@ -93,6 +93,19 @@ public function it_can_render_itself_with_responsive_images_and_a_placeholder()
...
+    public function it_can_render_itself_with_responsive_images_of_a_conversion_and_a_placeholder()
+    {
+        $media = $this->testModelWithResponsiveImages
+            ->addMedia($this->getTestJpg())
+            ->toMediaCollection();
+
+        $image = $media->refresh()->img('thumb');
+
+        $this->assertContains('/media/2/responsive-images/', $image);
+        $this->assertTrue(str_contains($image, 'data:image/svg+xml;base64,'));
```
freekmurze: Could you use `assertContains` here?

freekmurze:
Thanks!

PR #983, opened 2018-03-22
Improve check on filename in addMediaFromUrl method

TVke:


tests/Feature/FileAdder/IntegrationTest.php
```diff
@@ -296,10 +296,22 @@ public function it_can_add_a_remote_file_to_the_medialibrary()
...
 
+        $media = $this->testModel
+            ->addMediaFromUrl($url)
+            ->toMediaCollection();
+
+        $this->assertFileExists($this->getMediaDirectory("{$media->id}/test-image.jpeg"));
+    }
+
+    /** @test */
+    public function it_can_add_a_remote_file_with_no_name_or_directory_to_the_medialibrary()
```
brendt: This test now fails: https://travis-ci.org/spatie/laravel-medialibrary/jobs/356919799

src/HasMedia/HasMediaTrait.php
```diff
@@ -129,10 +129,15 @@ public function addMediaFromUrl(string $url, ...$allowedMimeTypes)
...
 
-            $filename = 'file.'.$mediaExt[1];
+        $mediaExt = explode('/', mime_content_type($tmpFile));
+
+        if ($filename === '') {
+            $filename = 'file';
+        }
+
+        if (! str_contains($filename, '.')) {
+            $filename = "{$filename}.{$mediaExt[1]}";
```
brendt: Are we sure `$mediaExt` will always be an array with at least two elements? Otherwise I'd add an `isset`-like check and a proper fallback.

TVke: Yes I just checked and there is always a slash in a mime type and it is always present. I can add the check but it is not necessary to do so.

tests/Feature/FileAdder/IntegrationTest.php
TVke: Yes the test fail because the PR at the docs site is not yet merged. After that it will pass.

src/HasMedia/HasMediaTrait.php
brendt: I can imagine there could be a corrupted image which mime type doesn't contain the slash? I would personally handle it like so.

```php
$extension = $mediaExt[1] ?? null;

if (!$extension) {
    // ...
}
```

src/HasMedia/HasMediaTrait.php
```diff
@@ -130,14 +130,20 @@ public function addMediaFromUrl(string $url, ...$allowedMimeTypes)
...
         if ($filename === '') {
             $filename = 'file';
         }
 
+        $mediaExt = explode('/', mime_content_type($tmpFile));
+
+        $extension = $mediaExt[1] ?? null;
+
+        if (! $extension) {
+            throw MimeTypeNotAllowed::create($filename, $allowedMimeTypes);
```
freekmurze: This doesn't seem right. Files without extensions are allowed.  The mimetype is already checked on line 129.

TVke: That is true, I'll remove the extra check

freekmurze:
Thanks!

PR #996, opened 2018-03-26
Add upgrade notes for renamed config keys

timothyasp:
Caught me by surprise during the upgrade from v6.* to v7.1.2

Keys changed in commit ba46d8008d26542c9a5ef0e39f779de801cd4f8f


freekmurze:
Thanks!

PR #1005, opened 2018-03-30
Fix remote exists check

### LINKED ISSUE #997: file_exists used for 'only if missing' option, fails for s3
We see the usage of the php native function `file_exists` here:
https://github.com/spatie/laravel-medialibrary/blob/master/src/FileManipulator.php#L79

However, this obviously fails for disks that are not local. This should be something like
`$this->filesystem->disk($media->disk)->exists($media->getPath($conversion->getName()))`
Maybe I'll tackle this soon together with my closed PR for the bug in the observer. (https://github.com/spatie/laravel-medialibrary/pull/965)

TVke:
Closes #997


freekmurze:
Thanks!

PR #1011, opened 2018-03-30
Sync both filename and disk

tomlankhorst:
Simplified version of my earlier PR. 
Medialibrary allows changing the `file_name` and moves filesystem files accordingly. 
This is not the case with `disk`. 
This PR refactors the renaming functionality to not only sync filename, but disk as well.

### DIFF
--- a/src/Filesystem/Filesystem.php
+++ b/src/Filesystem/Filesystem.php
@@ -120,49 +120,57 @@ class Filesystem
         $this->filesystem->disk($media->disk)->delete($path);
     }
 
-    public function syncFileNames(Media $media)
+    public function syncFiles(Media $media)
     {
-        $this->renameMediaFile($media);
+        $this->syncMediaFile($media);
 
-        $this->renameConversionFiles($media);
+        $this->syncConversionFiles($media);
     }
 
-    protected function renameMediaFile(Media $media)
+    protected function syncMediaFile(Media $media)
     {
         $newFileName = $media->file_name;
         $oldFileName = $media->getOriginal('file_name');
+
+        $newDisk = $media->disk;
+        $oldDisk = $media->getOriginal('disk');
 
         $mediaDirectory = $this->getMediaDirectory($media);
+        $oldMediaDirectory = $this->getMediaDirectory($media->setAttribute('disk', $oldDisk));
+        $media->setAttribute('disk', $newDisk);
 
-        $oldFile = $mediaDirectory.'/'.$oldFileName;
-        $newFile = $mediaDirectory.'/'.$newFileName;
+        $oldFile = $oldMediaDirectory.$oldFileName;
+        $newFile = $mediaDirectory.$newFileName;
 
-        $this->filesystem->disk($media->disk)->move($oldFile, $newFile);
+        $this->move($oldDisk, $oldFile, $newDisk, $newFile);
     }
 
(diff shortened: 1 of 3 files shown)


src/Filesystem/Filesystem.php
```diff
@@ -120,49 +120,57 @@ public function removeFile(Media $media, string $path)
...
+        $newDisk = $media->disk;
+        $oldDisk = $media->getOriginal('disk');
+
         $mediaDirectory = $this->getMediaDirectory($media);
+        $oldMediaDirectory = $this->getMediaDirectory($media->setAttribute('disk', $oldDisk));
+        $media->setAttribute('disk', $newDisk);
 
-        $oldFile = $mediaDirectory.'/'.$oldFileName;
-        $newFile = $mediaDirectory.'/'.$newFileName;
+        $oldFile = $oldMediaDirectory.$oldFileName;
```
tomlankhorst: There were '/' slashes inserted here, I removed this as they seemed to be inserted by `getMediaDirectory` already.

src/Filesystem/Filesystem.php
```diff
@@ -120,49 +120,57 @@ public function removeFile(Media $media, string $path)
         $this->filesystem->disk($media->disk)->delete($path);
     }
 
-    public function syncFileNames(Media $media)
+    public function syncFiles(Media $media)
```
freekmurze: Renaming this function is a breaking change. Please add `syncFileNames` back. It should call this new `syncFiles` function.

src/Filesystem/Filesystem.php
```diff
@@ -120,49 +120,57 @@ public function removeFile(Media $media, string $path)
...
     {
-        $this->renameMediaFile($media);
+        $this->syncMediaFile($media);
 
-        $this->renameConversionFiles($media);
+        $this->syncConversionFiles($media);
```
freekmurze: The responsive image files should be moved as well.

src/Filesystem/Filesystem.php
```diff
@@ -120,49 +120,57 @@ public function removeFile(Media $media, string $path)
...
     }
 
-    protected function renameMediaFile(Media $media)
+    protected function syncMediaFile(Media $media)
```
freekmurze: Wouldn't `renameMediaFiles` be a better name?

src/Filesystem/Filesystem.php
```diff
@@ -120,49 +120,57 @@ public function removeFile(Media $media, string $path)
...
         $oldFileName = $media->getOriginal('file_name');
 
+        $newDisk = $media->disk;
+        $oldDisk = $media->getOriginal('disk');
+
         $conversionDirectory = $this->getConversionDirectory($media);
+        $oldConversionDirectory = $this->getConversionDirectory($media->setAttribute('disk', $oldDisk));
+        $media->setAttribute('disk', $newDisk);
 
         $conversionCollection = ConversionCollection::createForMedia($media);
```
freekmurze: Wouldn't it be easier to just move the entire contents of the conversions directory (and the responsive images directory)?

bbashy:
Waiting on this (unless there's a good way to move from local to S3 without this?).

tomlankhorst:
Trying to find time to do this but I'm a little bit short. 
Will try later this week.

freekmurze:
Closing due to inactivity

PR #1022, opened 2018-04-05
Add multidisk zipfile stream

TVke:


freekmurze:
`maennchen/zipstream-php` on `0.5` still produces corrupt zip files when streaming a zip that uses files from multiple filesystems.

freekmurze:
Seems to be fixed in `0.5.2`

freekmurze:
I'll implement this myself.

PR #1027, opened 2018-04-09
 Customize the responsive images file type

anjemark:
Use the converted image extension to the responsive images

### DIFF
--- a/src/ResponsiveImages/ResponsiveImageGenerator.php
+++ b/src/ResponsiveImages/ResponsiveImageGenerator.php
@@ -77,7 +77,7 @@ class ResponsiveImageGenerator
         int $targetWidth,
         BaseTemporaryDirectory $temporaryDirectory
     ) {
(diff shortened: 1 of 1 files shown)


freekmurze:
Could you add tests to make sure this works correctly?

buttjer:
I tried this patch and it seem to convert the source image to a target format. So this conversion. works fine:

```php
$this->addMediaConversion('gallery')
    ->quality(1)
    ->background('blue')
    ->format(Manipulations::FORMAT_JPG)
    ->withResponsiveImages();
```

This can be very useful for uploading files as jpg and converting it to jpg with different background colors.

I'm not sure, but this should be the test?

```php
    /** @test */
    public function its_conversions_can_have_responsive_images_with_format_change()
    {
        $this->testModelWithResponsiveImages
                    ->addMedia($this->getTestPng())
                    ->format(Manipulations::FORMAT_JPG)
                    ->withResponsiveImages()
                    ->toMediaCollection();

        $this->assertFileExists($this->getTempDirectory('media/1/responsive-images/test___thumb_50_41.jpg'));
    }
```

Issue #1049, opened 2018-04-19
SoftDeletes trait on parent class


dwhite-nep:
https://github.com/spatie/laravel-medialibrary/blob/3683133d8c6a585225e7473d2a37c843de181e35/src/HasMedia/HasMediaTrait.php#L44

This soft deletes check does not currently work on classes that inherit the SoftDeletes trait.  I was able to get it to work in my code by moving the SoftDeletes trait to the child classes but it seems like it would work if `trait_uses_recursive` was changed to `class_uses_recursive` which would return all inherited traits.

freekmurze:
Thanks you for your research. Please PR this.

PR #1050, opened 2018-04-20
Checking traits of parent classes when checking for SoftDeletes trait

dwhite-nep:
`trait_uses_recursive` does not include any traits used by a parent class while `class_uses_recursive` does.  
This change will prevent media from being deleted if your class inherits from a class that uses the SoftDeletes trait.  Addresses Issue #1049.


freekmurze:
Thanks!

PR #1069, opened 2018-04-27
added path so it is useful as path has been changed

hardikdangar:
Path of hasmedia is changed as it is no longer using interface folder.


freekmurze:
Thanks!

PR #1074, opened 2018-05-01
Replace url encoded spaces in filenames when adding media from URL

chadjw:
URLs that have encoded spaces (ex. %20) cause problems when creating the filename. It will save it with the %20 in the filename and the file won't be found later during conversion.


chadjw:
I see that just about all of the tests for addMediaFromUrl use images within this location in your documents: 

https://docs.spatie.be/images/medialibrary/

Can you upload an image to this location that has a space in the image name (ex. https://docs.spatie.be/images/medialibrary/image name with space.jpg)?

freekmurze:
Here's an image with a space in its name: 
`http://spatie.github.io/laravel-medialibrary/tests/Support/testfiles/test with space.jpg`

freekmurze:
Thank you for your work on this!

PR #1079, opened 2018-05-02
Fixes #1076

### LINKED ISSUE #1076: Updated event handler always does createDerivedFiles
This is to keep track of an issue previously discussed in a closed PR (missing tests):
https://github.com/spatie/laravel-medialibrary/pull/965/files

The problem: 
The MediaObserver registers a callback for the `updated` event, which checks the `manipulations` property on the media object, and if it was changed it will make sure all conversions are there with the `createDerivedFiles` from `FileManipulator` , done here:
https://github.com/spatie/laravel-medialibrary/blob/7.3.0/src/MediaObserver.php#L29

The manipulations will always be an array (cast like that in the model I think from top of my head), so the comparison will ALWAYS fail as the second parameter to `json_decode` misses.

afraca:
See issue for details. The test setup was a bit tricky, because I couldn't really figure out how to check what the FileManipulator has done in the observer, so I just use the modification time.
Especially the test for not changed manipulations was tricky because by default we'll have an empty manipulations array, and then in the observer the `json_decode` will return an **array**, even though normally without the second parameter set to true it will return an object... Yeah, thanks php. (https://3v4l.org/2b5oD)

### DIFF
--- a/src/MediaObserver.php
+++ b/src/MediaObserver.php
@@ -26,7 +26,7 @@ class MediaObserver
             return;
         }
 
-        if ($media->manipulations !== json_decode($media->getOriginal('manipulations'))) {
+        if ($media->manipulations !== json_decode($media->getOriginal('manipulations'), true)) {
             app(FileManipulator::class)->createDerivedFiles($media);
         }
     }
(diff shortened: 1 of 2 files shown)


afraca:
What to do with the StyleCI failing?

afraca:
Cool. Keep in mind that `json_decode('[]')` will NOT return an object, but that was likely why it didn't really surfaced before as most people don't set the manipulations array themselves.

brendt:
Looks good! I'm going to merge the StyleCI and cleanup the tests after merging your PR, thanks for the help!

PR #1080, opened 2018-05-02
Reduce test failures on Windows by using DIRECTORY_SEPERATOR

### LINKED ISSUE #1077: Use DIRECTORY_SEPARATOR for windows compability (tests only?)
I was trying to make a PR for https://github.com/spatie/laravel-medialibrary/issues/1076 , but due to circumstances had to do so on Windows.. PhpUnit will crash currently because of this:

https://github.com/spatie/laravel-medialibrary/blob/7.3.0/src/UrlGenerator/LocalUrlGenerator.php#L55
The call to `getStoragePath` is problematic. Exactly how I can't remember but it will use a value from this function:
https://github.com/spatie/laravel-medialibrary/blob/7.3.0/tests/TestCase.php#L149
that uses hardcoded directory separators. Then the comparison in the LocalUrlGenerator will fail **on windows** (as one path has a backslash in it and one does not) and an exception will be thrown. (And it's in some __toString() so php is NOT happy with that)

I have worked around this (so it seems...) by just using `realpath` around the return value from https://github.com/spatie/laravel-medialibrary/blob/7.3.0/tests/TestCase.php#L149 , that might be helpful for someone looking into this.

edit: Maybe a bit confusing that I talk about two solutions
- use realpath
- use DirectorySeparator

I'm currently not at my Windows pc to test whether the DS solution works, but it should and I think it's the best of the two, though not sure why.

afraca:
Fixes #1077  .
I skipped fixing the test with the literal html strings in it, then you get really really ugly strings, but if you want it all I can add those as well.

Also, I though about doing 
`define('DS', DIRECTORY_SEPARATOR)`
Don't know how you feel about that, and somehow I got definition errors or something.

### DIFF
--- a/tests/Feature/Media/ToHtmlTest.php
+++ b/tests/Feature/Media/ToHtmlTest.php
@@ -89,7 +89,7 @@ class ToHtmlTest extends TestCase
 
         $image = $media->refresh()->img();
 
-        $this->assertEquals(3, substr_count($image, '/media/2/responsive-images/'));
+        $this->assertEquals(3, substr_count($image, DIRECTORY_SEPARATOR . 'media/2/responsive-images/'));
         $this->assertTrue(str_contains($image, 'data:image/svg+xml;base64,'));
     }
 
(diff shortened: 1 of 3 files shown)


afraca:
Also really important, I couldn't really figure out why it was often only necessary to replace **only the first directory separator**...

brendt:
Couldn't we use `realpath`? 

http://php.net/manual/en/function.realpath.php

brendt:
I think both: the tests should also work with `realpath`, instead of the manual use of `DIRECTORY_SEPARATOR` you've got in the PR. 

But changing the tests isn't enough, the `BasePathGenerator` should also use `realpath`

freekmurze:
I decided I'm going to leave things as they are for now.

PR #1081, opened 2018-05-02
Default Windows php has pdo_sqlite not enabled, error early

### LINKED ISSUE #1078: Missing php extension in require-dev for Windows
As noted in my other issue (https://github.com/spatie/laravel-medialibrary/issues/1077) I had to run the package tests on Windows...  By default the Windows distribution of php does not have the `php_pdo_sqlite` extension enabled, so I think "we" should include it in the composer.json. I'll open a PR soon.

afraca:
On Windows it's confusing to get error on database problem, so be more clear with this composer.json entry. All tests run with sqlite and an in-memory db.

Closes #1078

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,6 +35,7 @@
     "spatie/temporary-directory": "^1.1"
   },
   "require-dev": {
+    "ext-pdo_sqlite": "*",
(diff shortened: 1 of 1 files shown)


brendt:
Thanks!

Issue #1115, opened 2018-05-21
Regenerating images will keep adding same image urls to responsive images


DevDavido:
First I wanted to express a big **THANK YOU!** for this awesome package! It made dealing with media files and models much easier!

Back to the unexpected behaviour I just stumbled upon:
If I run `medialibrary:regenerate` with a conversion which has responsive images it will add those image urls to the database column `responsive_images` even it already has the exact same entries. Shouldn't be there a clear of the current image urls before adding the new ones?

Elhebert:
+1 The url should be cleared before the new ones are added.

robinvalk:
This can be closed as it's now implemented :+1:

DevDavido:
Thanks @Elhebert 👍

PR #1154, opened 2018-06-19
clean responsive_images files and urls on (re)generation

Elhebert:
Related to #1115.

When generating responsive images the new urls are appended to the `urls` array in db, but that array is never cleaned, which means if we regenerate the images another time (using the artisan command for example) this new urls will also be applied to the `urls` array instead of replacing the existing one.

I also saw that when generating responsive images it wasn't removing the old physical files. This could be potentially harmful.
If you make a modification in a conversion size and regenerate images on your production env, the old images would still be present, thus it could still use those images because it matches the `sizes` attribute better.

I'm not sure this is the best way to achieve this, so comments are welcome 👍


src/ResponsiveImages/ResponsiveImageGenerator.php
```diff
@@ -149,4 +154,25 @@ protected function guardAgainstInvalidTinyPlaceHolder(string $tinyPlaceholderPat
             throw InvalidTinyJpg::hasWrongMimeType($tinyPlaceholderPath);
         }
     }
+
+    private function cleanResponsiveImages(Media $media, string $conversionName = 'medialibrary_original') : Media
```
freekmurze: Make this `protected`

freekmurze:
Thanks!

PR #1174, opened 2018-07-11
Add ability to set relations to the saved media

dmitrybubyakin:
I have custom PathGenerator implementation which depends on some model attributes:

```php
class MyPathGenerator implements PathGenerator
{
    public function getPath(Media $media): string
    {
        return $this->getBasePath($media).'/';
    }

    public function getPathForConversions(Media $media): string
    {
        return $this->getBasePath($media).'/conversions/';
    }

    public function getPathForResponsiveImages(Media $media): string
    {
        return $this->getBasePath($media).'/responsive-images/';
    }

    {
        return $media->model->getMediaPath();
    }
}
```

When I add many media items, `$media->model->getMediaPath()` performs a query on the database 
 to load `model`. So we need to make N + 1 queries.

This PR adds ability to set relations to the saved media:

```php
foreach ($files as $file) {
    $model
        ->addMedia($file)
        ->withRelations(['model' => $model]) // solves the N + 1 problem
        ->toMediaCollection();
}
```

### DIFF
--- a/src/FileAdder/FileAdder.php
+++ b/src/FileAdder/FileAdder.php
@@ -63,6 +63,9 @@ class FileAdder
     /** @var array */
     protected $customHeaders = [];
 
+    /** @var array */
+    protected $relations = [];
+
     /**
      * @param Filesystem $fileSystem
      */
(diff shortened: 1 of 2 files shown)


freekmurze:
I don't understand why this is needed. Could you give a more comprehensive example?

dmitrybubyakin:
I've found a better way to solve this, without setting a relation. Thanks for you reply.

PR #1183, opened 2018-07-23
Fix skipped test messages for WebP

hohl:
I guess thats just a copy & paste issue, but the warning could become a bit miss-leading at some point and I just saw it by accident. So why not fix it...


fragkp:
Duplicate #1175 :)

hohl:
Oh, sorry. Missed that.

PR #1189, opened 2018-07-27
Added getFirstMediaFullUrl method to HasMediaTrait

hivokas:
I quite often need to get the **full** url of the image for first media, but currently I don't have such ability.
So I've added `getFirstMediaFullUrl` method to `HasMediaTrait`.


josegus:
why? this is a good pull request, i currently need an array of full urls. Would be great if we could get this array directly from the trait

PR #1195, opened 2018-07-30
Added support for absolute urls to use a responsive image sources

robinvalk:
This PR allows absolute urls to be generated as sources for responsive images. This can be helpful if the html will be used on a different domain than the laravel instance is running on.

### DIFF
--- a/config/medialibrary.php
+++ b/config/medialibrary.php
@@ -68,6 +68,12 @@ return [
          * the medialibrary will use a tiny blurred jpg image.
          */
         'tiny_placeholder_generator' => Spatie\MediaLibrary\ResponsiveImages\TinyPlaceholderGenerator\Blurred::class,
+
+        /*
+         * By enabling this setting, the sources generated for the responsive images
+         * will include the domain based on the app url.
+         */
+        'generate_absolute_urls' => false,
     ],
 
     /*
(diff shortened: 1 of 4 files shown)


freekmurze:
Wouldn't it be easier to just always generate absolute urls?

robinvalk:
As you can see I added the commit. Might not even need to use the PR :joy:

freekmurze:
Instead of concatenating I believe you can pass a string to the `url` function as an argument. Could you try to see if that works?

robinvalk:
You are right, I also updated the tests. I couldn't run these yesterday because I had some environment issues. Should all be succeeding now.

I'm using `http://localhost` as the hardcoded domain for now. I tried to keep it simple stupid so I'm keeping an eye on travis if that's using a different domain. If so we should look into a config option for the test.

robinvalk:
It seems that Travis is succeeding so the hardcoded domain should be good. I'm not too happy with the trailing slash that I had to add, but `url()` trims it off the end.

freekmurze:
I'm thinking the url will not be correct if the images are hosted on an external filesystem like S3.

robinvalk:
S3 has it's own URL generator, I think that would be a separate issue. It would be awesome if this fix can be merged separately. Then we can test the responsive images on staging for our project.

freekmurze:
Thanks!

Issue #1173, opened 2018-07-10
Custom image generator trying to convert file that isn't in supported extensions.


bcreeves:
I have a custom image generator with only `['ppt', 'pptx']` as supported extensions. When a .potx file is uploaded this generator tries to convert it. 

I checked the mime type of the file and it is "application/vnd.openxmlformats-officedocument.presentationml.presentation" which should be a pptx file and which I do have in the `supportedMimeTypes` collection.

My question is shouldn't the file have to pass both tests, mime-type and file extension requirements, before it's converted? Is there any way I can force it to pass both tests?

freekmurze:
There currently isn't a way to force this, I'd accept a fully documented and fully tested PR that adds this to the package.

PR #1201, opened 2018-08-07
Add support for forcing checks of both mimetype extension

Jasper9041:
This PR allows the BaseGenerator class to be configured so a check of both the mimetype and file extension are forced.

The default behaviour remains unchainged: a generator matches files with either a valid mimetype OR extension (or both).
Returning true in the method `shouldMatchBothExtensionsAndMimeTypes` only matches files with a valid mimetype AND extension.

This PR is in response to issue #1173

### DIFF
--- a/src/ImageGenerators/BaseGenerator.php
+++ b/src/ImageGenerators/BaseGenerator.php
@@ -13,15 +13,14 @@ abstract class BaseGenerator implements ImageGenerator
             return false;
         }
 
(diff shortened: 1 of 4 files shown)


freekmurze:
It's breaking in it's current status.

I'll consider this PR again when creating a new major version of the package.

freekmurze:
This change has been implanted on the `v8` branch 👍

Issue #1066, opened 2018-04-26
Old issue regression? keepOriginalImageFormat


virtruvio:
Encountering a bug when using registered keepOriginalImageFormat() method while converting .MP4 video.  Registered conversion (e.g. "thumbnail") is properly executed by conversions classes (resulting in a JPEG frame grab for a converted image), but, post-conversion, media->getUrl('thumbnail') returns original extension of .mp4.  The result is that the proper 'thumbnail' registered conversion is done and the file in place in the filesystem, but -getUrl() returns for instance.  ... thumbnail.... .mp4 instead of ... thumbnail... .jpg

freekmurze:
I'd accept a PR with tests that fixes this.

PR #1202, opened 2018-08-07
Only use keepOriginalImageFormat when the original file was an actual image

Jasper9041:
This pr fixes issue #1066 
It adds an extra check to make sure the keepOriginalImageFormat does not cause the extension of a non image file to be used.


freekmurze:
Thanks!

lostlian:
@Jasper9041 Hi Jasper. I trully have spent some time trying to know why the KeepOriginalFileExtension is not working for me.

```
       
protected $sizes = ['xs', 'sm', 'md', 'lg','xl','full','thumb'];

	
protected $size = [
		'xs' 	=> 80,
		'sm' 	=> 150,
		'thumb'	=> 300,
		'md'   	=> 600,
		'lg'    => 1200,
		'xl'    => 1900,
	];

    {
        foreach ($this->sizes as $size) {
            if ($size!='full') {
                $width = $this->size[$size];
                $height = $this->size[$size];

                $mediaConversion = $this->addMediaConversion($size);
                $mediaConversion = $mediaConversion->width($width);
                $mediaConversion = $mediaConversion->height($height);
                $mediaConversion = $mediaConversion->keepOriginalImageFormat();

                Log::info('Should keep Original format ' . $mediaConversion->shouldKeepOriginalImageFormat());
                Log::info('Result Extension ' . $mediaConversion->getResultExtension());
            }
        }
    }
```

**The uploading file is logo.png**

The Output is:

```
Should keep Original format 1
Result Extension jpg
```

I spent sometime in the conversion class and found something odd.

```
public function getResultExtension(string $originalFileExtension = ''): string
    {
        if ($this->shouldKeepOriginalImageFormat()) {
            if (in_array($originalFileExtension, ['jpg', 'pjpg', 'png', 'gif'])) {
                return $originalFileExtension;
            }
        }

        if ($manipulationArgument = $this->manipulations->getManipulationArgument('format')) {
            return $manipulationArgument;
        }

        return $originalFileExtension;
    }

    public function getConversionFile(string $file): string
    {
        $fileName = pathinfo($file, PATHINFO_FILENAME);

        $extension = $this->getResultExtension(); // The file extension is not send in the parameters
        //The extension will always return jpg as defined in the constructor

        if (! $extension) {
            $extension = pathinfo($file, PATHINFO_EXTENSION);
        }

        return "{$fileName}-{$this->getName()}.{$extension}";
    }
```
The method getResultExtension is always being called without parameters. So it will always return the format defined in the constructor. (in this case jpg)

am i doing something wrong?

PR #1209, opened 2018-08-10
Fix CleanCommand

nicolasbeauvais:
Threw an error when media did not have responsive images:

```
In CleanCommand.php line 139:
array_keys() expects parameter 1 to be array, null given  
```


freekmurze:
Thanks!

PR #1236, opened 2018-09-08
Simplify area calculation

danhunsaker:
`w * w * (h / w) == w * h`

The simplified calculation is a bit more obvious about this, and possibly also slightly faster.


freekmurze:
Thanks!

PR #1238, opened 2018-09-12
fix composer requires schema, fixes #1235

### LINKED ISSUE #1235: lumen can't install this package due to laravel/framework as package requirements
Began from laravel-medialibrary [7.4.1](https://github.com/spatie/laravel-medialibrary/blob/7.4.1/composer.json#L26), package requirements in composer.json are used `laravel/framework` to restrict laravel version. However, this will prevent lumen to install this package. 

Can you fallback the package requirements schema in composer.json to [7.4.0](https://github.com/spatie/laravel-medialibrary/blob/7.4.0/composer.json#L26-L30) version?

BePsvPT:


freekmurze:
Thanks!

PR #1278, opened 2018-10-22
Fix install step failing on some shells

nicolasbeauvais:
On zsh/prezto running the composer install command without quotes trigger an error:
`zsh: no matches found: spatie/laravel-medialibrary:^7.0.0`

Adding quotes solve the issue on all shells.


freekmurze:
Thanks!

Issue #1277, opened 2018-10-19
How to build my own Conversion?


pedrofurtado:
I want to build my own conversion, that converts ".wav" files in HLS. How are the steps to build and register my conversion that integrates with the package (the structure of files and the main classes involved)?

Anyway, thanks for the package! It is awesome! :+1:

nicolasbeauvais:
Hello @pedrofurtado,

the process to create your own ùedia generator is described in the documentation: https://docs.spatie.be/laravel-medialibrary/v7/converting-other-file-types/creating-a-custom-image-generator

You can take a look at the base Video generator that takes a video file in input to create a preview image in output using ffmpeg: https://github.com/spatie/laravel-medialibrary/blob/master/src/ImageGenerators/FileTypes/Video.php

pedrofurtado:
@nicolasbeauvais Thanks for feedback! I read the links you post, and it worked here. :+1:  :handshake:

pedrofurtado:
I've created the generator with code below:

```php
<?php

namespace App\Spatie\MediaLibrary\Generators;

use Illuminate\Support\Collection;
use Spatie\MediaLibrary\Conversion\Conversion;
use Spatie\MediaLibrary\ImageGenerators\BaseGenerator;

class Wav extends BaseGenerator
{
    public function convert(string $file, Conversion $conversion = null) : string
    {
        $pathToWavFile = pathinfo($file, PATHINFO_DIRNAME).'/'.pathinfo($file, PATHINFO_FILENAME).'.wav';
        $pathToM3u8File = pathinfo($file, PATHINFO_DIRNAME).'/'.pathinfo($file, PATHINFO_FILENAME).'.m3u8';
        $pathToTsFile = pathinfo($file, PATHINFO_DIRNAME).'/'.pathinfo($file, PATHINFO_FILENAME).'-%05d.ts';

        exec("ffmpeg -i {$pathToWavFile} -vn -ac 2 -acodec aac -f segment -segment_format mpegts -segment_time 100 -segment_list {$pathToM3u8File} {$pathToTsFile}");

        return $pathToM3u8File;
    }

    public function requirementsAreInstalled() : bool
    {
        return true;
    }

    public function supportedExtensions() : Collection
    {
        return collect(['wav', 'wave']);
    }

    public function supportedMimeTypes() : Collection
    {
        return collect([
            'audio/wav',
            'audio/wave',
            'audio/x-wav',
            'audio/x-pn-wav'
[...]

PR #1279, opened 2018-10-22
Fix for issue #1277

nicolasbeauvais:
This PR make two changes:

### Minor: 
The `performConversion` method only perform manipulations has the conversion is made in the `performConversions` method [on line 94](https://github.com/spatie/laravel-medialibrary/compare/master...nicolasbeauvais:master#diff-0d11c445b5c28ccade90cf390e511012L94). 
So it makes more sense to rename it `performManipulations`instead. It will also reduce confusion between `performConversions` and `performConversion`.

#### Problem:
Creating a custom generator that returns a file that is not an image (in that case a .m3u8 file) fail because manipulations are being automatically run on that file which is not compatible.

#### Proposed fix:
The end user needs to disable the default manipulations for the specific conversion like:
```php
$this->addMediaConversion('transcode')
            ->removeManipulation('format')
            ->removeManipulation('optimize')
            ->performOnCollections('file');

// We could also add a base method for this:

$this->addMediaConversion('transcode')
            ->removeAllManipulations()
            ->performOnCollections('file');
```

And in the core, we check if the running conversion has manipulation to be triggered, if not we skip this step and the converted file stay untouched.

I haven't dive in the core code for a long time so I might be missing something here, anyway do not hesitate to indicate any change that could be made to improve this fix.

### DIFF
--- a/src/FileManipulator.php
+++ b/src/FileManipulator.php
@@ -93,13 +93,17 @@ class FileManipulator
 
                 $copiedOriginalFile = $imageGenerator->convert($copiedOriginalFile, $conversion);
 
-                $conversionResult = $this->performConversion($media, $conversion, $copiedOriginalFile);
+                if (!$conversion->getManipulations()->isEmpty()) {
+                    $manipulationResult = $this->performManipulations($media, $conversion, $copiedOriginalFile);
+                } else {
+                    $manipulationResult = $copiedOriginalFile;
+                }
 
                 $newFileName = pathinfo($media->file_name, PATHINFO_FILENAME).
                     '-'.$conversion->getName().
                     '.'.$conversion->getResultExtension(pathinfo($copiedOriginalFile, PATHINFO_EXTENSION));
 
-                $renamedFile = MediaLibraryFileHelper::renameInDirectory($conversionResult, $newFileName);
+                $renamedFile = MediaLibraryFileHelper::renameInDirectory($manipulationResult, $newFileName);
 
                 if ($conversion->shouldGenerateResponsiveImages()) {
                     app(ResponsiveImageGenerator::class)->generateResponsiveImagesForConversion(
@@ -119,7 +123,7 @@ class FileManipulator
         $temporaryDirectory->delete();
     }
 
-    public function performConversion(Media $media, Conversion $conversion, string $imageFile): string
+    public function performManipulations(Media $media, Conversion $conversion, string $imageFile): string
     {
         $conversionTempFile = pathinfo($imageFile, PATHINFO_DIRNAME).'/'.str_random(16)
             .$conversion->getName()
--- a/src/FileManipulator.php
+++ b/src/FileManipulator.php
@@ -93,11 +93,7 @@ class FileManipulator
 
                 $copiedOriginalFile = $imageGenerator->convert($copiedOriginalFile, $conversion);
 
-                if (! $conversion->getManipulations()->isEmpty()) {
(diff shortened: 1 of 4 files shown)


pedrofurtado:
@nicolasbeauvais The CI failed because of some rules of coding standard 👍

pedrofurtado:
Ping for notice: @freekmurze

src/FileManipulator.php
```diff
@@ -93,13 +93,17 @@ public function performConversions(ConversionCollection $conversions, Media $med
 
                 $copiedOriginalFile = $imageGenerator->convert($copiedOriginalFile, $conversion);
 
-                $conversionResult = $this->performConversion($media, $conversion, $copiedOriginalFile);
+                if (! $conversion->getManipulations()->isEmpty()) {
+                    $manipulationResult = $this->performManipulations($media, $conversion, $copiedOriginalFile);
+                } else {
```
freekmurze: Could your refactor this so we don't need an `else` statement? A dedicated method with an early return seems the way to go.

freekmurze:
@nicolasbeauvais Thanks for this! 👍 Could you add a test to make sure that these changes work?

nicolasbeauvais:
WIP:
- [x] Refactored the fix using an early return in the `performManipulations` method.
- [x] Created a `withoutManipulations` helper function for conversions.
- [x] Added some missing return types in `Conversion.php`.
- [ ] Add tests for the fix.
- [ ] Add tests for the `withoutManipulations` method.
- [ ] Update the documentation to reflect this changes

nicolasbeauvais:
I should be done with this, of course, feel free to suggest any improvements.

I will update the documentation once this gets merged!

freekmurze:
Thank you so much for your work on this 👍

chelout:
@nicolasbeauvais @pedrofurtado could you please add an example who to work with this feature?

sp4r74cus:
hi @pedrofurtado, can you confirm your hls conversion works with this update? i tried to make a similar conversion (mp4 -> hls) using your code as a start ([#1277](https://github.com/spatie/laravel-medialibrary/issues/1277#issuecomment-431634011)) 

i'm facing 2 issues:

1. the conversion file is saved as video-{conversion-name}.mp4 and not .m3u8 (the file name i return)
2. all the .m3u8 & .ts generated in the tmp directory are not moved to the conversions folder 

Thanks

jayenne:
did the two issues above regarding conversio nfile name and moving t ocorrect dir get resolved?
- would you be able to share a howto for converting video/audio media to .m3u8 & .ts please?
- or point us to a guide/resource please?

joserick:
Hi all.

`the conversion file is saved as video-{conversion-name}.mp4 and not .m3u8 (the file name i return)`

I had the first problem @jayenne , I solved it in the following way:

```php
$this->addMediaCollection('audio')->singleFile()
    ->registerMediaConversions(function (Media $media) {
        $this->addMediaConversion('ogg')
            //->withoutManipulations() #Removed for redundancy in this case.
            ->setManipulations(function (Manipulations &$manipulations){
                $manipulations = new class extends Manipulations{
                    const FORMAT_OGG = 'ogg';

                    public function isEmpty(): bool
                    {
                        return true;
                    }
                };

                $manipulations->format('ogg');
            })->nonQueued();
    });
```
```php
MyModel::first()->addMedia('path/to/file.wav')->toMediaCollection('audio');
```
```php
MyModel::first()->getFirstMediaUrl('audio', 'ogg');
// http://localhost/storage/#/conversions/file-ogg.ogg <- Already change the extension
```

I hope it will help someone in the future :)

PR #1286, opened 2018-10-26
V6: Add support for Laravel 5.7

jkudish:
Hey there,

I'm still using v6 on a project and unfortunately the migration is not that easy and not a priority for this project; however we did upgrade to Laravel 5.7 so would be nice to have support for that!

Thanks!


freekmurze:
Thanks!

PR #1292, opened 2018-11-01
add missing condition - should sort when creating

Goldfinch:
Seems like **shouldSortWhenCreating** in **IsSorted** trait has been left at a loose end.


freekmurze:
Thanks!

PR #1322, opened 2018-11-29
Updates PHPDoc for media() method of HasMediaTrait

hivokas:
This method always returns `\Illuminate\Database\Eloquent\Relations\MorphMany`.
So why not to change the return type specified in PHPDoc in order to improve highlighting in IDE?


freekmurze:
Thanks!

PR #1342, opened 2018-12-20
Dynamic validation and legend strings generation.

Okipa:
@freekmurze, here is my implementation proposition (https://github.com/spatie/laravel-medialibrary/issues/858).  
This full tested pull request provides the following additional features :

#### Ability to add the mime types constraints for each collection.
In addition to the `acceptsFile` method, this implementation give the possibility to add the following constraint, which would be used for the following added features (see below), but also to trigger the `FileUnacceptableForCollection` exception if not respected : `acceptsMimeTypes(array $mimeTypes)` 
Example :  
```php
public function registerMediaCollections()
{
    $this->addMediaCollection('images')->acceptsFile(function (File $file) {
        return $file->size <= 30000;
    })->acceptsMimeTypes(['image/jpeg']);
}
```

#### Ability to dynamically generate a collection-related validation string.
Example :  
```php
// in your user storing form request
public function rules()
{
    return [
        'avatar' => app(User::class)->validationConstraints('avatar'),
        // example : "dimensions:min_width=60,min_height=20|mimetypes:image/jpeg,image/png"
        // other validation rules
    ];
}
```

This method call itself the `dimensionValidationConstraints($collectionName):string` and the `mimeTypesValidationConstraints($collectionName):string` methods, which can be called separately if needed.  

#### Ability to dynamically generate a collection-related input legend string.
Example :  
```html
// in your HTML form
<label for="avatar">Choose a profile picture :</label>
<input type=" id="avatar" name="avatar" value="{{ $avatarFileName }}">
<small>{{ app(User::class)->constraintsLegend('avatar') }}</small>
<!-- for example : "Min. width : 150 px / Min. height : 70 px. Accepted MIME Type(s) : image/jpeg, image/png." -->
``` 

This method call itself the `collectionDimensionsLegend($collectionName):string` and the `collectionMimeTypesLegend($collectionName):string` methods, which can be also be called separately.

#### Extra methods
Those features both use the `collectionMaxSizes(collectionName): array` method, which also can be used if needed (public method).

### DIFF
--- a/resources/lang/en/medialibrary.php
+++ b/resources/lang/en/medialibrary.php
@@ -1,0 +1,12 @@
+<?php
+
+return [
+    'constraint' => [
+        'dimensions' => [
+            'both'   => 'Min. width : :width px / Min. height : :height px.',
+            'width'  => 'Min. width : :width px.',
+            'height' => 'Min. height : :height px.',
+        ],
+        'mimeTypes'  => 'Accepted MIME Type(s) : :mimetypes.',
+    ],
+];
--- a/resources/lang/en/medialibrary.php
+++ b/resources/lang/en/medialibrary.php
@@ -1,6 +1,7 @@
 <?php
 
 return [
+
     'constraint' => [
         'dimensions' => [
             'both'   => 'Min. width : :width px / Min. height : :height px.',
@@ -9,4 +10,5 @@ return [
         ],
         'mimeTypes'  => 'Accepted MIME Type(s) : :mimetypes.',
     ],
+
 ];
--- a/resources/lang/fr/medialibrary.php
+++ b/resources/lang/fr/medialibrary.php
@@ -1,0 +1,12 @@
+<?php
+
+return [
+    'constraint' => [
+        'dimensions' => [
+            'both'   => 'Largeur minimale : :width pixels / Hauteur minimale : :height pixels.',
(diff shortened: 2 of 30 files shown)


freekmurze:
Thank you for your work on this, but i've decided not to merge this in for name. There's some more code changes involved than I anticipated. I like to keep the package light.

Okipa:
Hi @freekmurze, thanks for your review, too bad this was not accepted ;)  
I fully understand your point of view regarding the weight of this package (it must already take time to maintain it).  
Nerver the less, I would like to know if you would mind if I created a new package named `laravel-medialibrary-extension` for example, allowing people to add some extra features to this great package.  
In fact, I use the features I proposed every day at work and they are such time-saving that I would be ready to set this up in a separate package and to maintain it.

freekmurze:
Yeah sure, go ahead and create such a package. When you're done I'd accept a PR that adds a link to it in our readme. 👍

Okipa:
Hey @freekmurze, the https://github.com/Okipa/laravel-medialibrary-extension package is in development. Everything is functional, I still have a build issue to fix on Scrutinizer (package service provider not found on Scrutinizer when everything is going fine locally, strange).  
I'll let you know when the build will be in success and when the extension will be released.

PR #1350, opened 2019-01-02
Check if specified frame second exeeds video length

### LINKED ISSUE #1335: Video upload ends up maxing CPU and fills drive with temp files when thumbnail is past video length
I'm not sure why this is happening. I've uploaded a few small 1-3 MB files and now the CPU has spiked and `storage/medialibrary/temp` is filling with temporary files. This will continue until the disk is full. Any ideas what might be happening and how to resolve this? I am using a redis queue and the public disk.

larsjanssen6:
Fix #1335

### DIFF
--- a/src/ImageGenerators/FileTypes/Video.php
+++ b/src/ImageGenerators/FileTypes/Video.php
@@ -21,14 +21,10 @@ class Video extends BaseGenerator
         ]);
 
         $video = $ffmpeg->open($file);
-
-        $seconds = $conversion ? $conversion->getExtractVideoFrameAtSecond() : 0;
-
         $duration = $ffmpeg->getDuration();
-
-        if ($duration < $seconds) {
-            throw InvalidConversionParameter::invalidFrameSecond($seconds, $duration);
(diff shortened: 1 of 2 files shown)


freekmurze:
Could you update the required version of the package so the lower dependencies do not fail?

src/ImageGenerators/FileTypes/Video.php
```diff
@@ -23,6 +24,12 @@ public function convert(string $file, Conversion $conversion = null): string
 
         $seconds = $conversion ? $conversion->getExtractVideoFrameAtSecond() : 0;
 
+        $duration = $ffmpeg->getDuration();
+
+        if ($duration < $seconds) {
+            throw InvalidConversionParameter::invalidFrameSecond($seconds, $duration);
```
freekmurze: I'm not sure if we should throw an exception here. A more safe default is to get the first frame. If people want an exception instead that could use a custom ImageGenerator

freekmurze:
Thanks!

Issue #1336, opened 2018-12-17
Custom headers always added as a custom property


denitsa-cm:
`custom_headers` are always saved as `custom_properties`, even if not set. Is this for any particular reason? 

I always have `{"custom_headers": []}` present in the column.

freekmurze:
Mmmm, that shouldn't happen, I'd accept a fully tested PR that fixes this.

denitsa-cm:
@freekmurze I've made a pull request for this. First pull request to an open source repo - yay!

freekmurze:
@denitsa-cm Thank you and congrats on your code being merged and released.

PR #1351, opened 2019-01-04
Only set `custom_headers` property if explicitly set

denitsa-cm:
Fix for #1336. 

In the `toMediaCollection()` method, the call to `setCustomHeaders()` is now only made if the `customHeaders` array is filled.


freekmurze:
Thank you very much!

Issue #1101, opened 2018-05-12
return default file, if there is none


FaridAghili:
Hi,
Is it possible to return a default file (an image for example), if the media collection is empty?

I have a model with following config:
```php
public function registerMediaCollections()
{
    $this->addMediaCollection('image')
         ->singleFile()
         ->registerMediaConversions(function (Media $media) {
             $this->addMediaConversion('resized')
                  ->fit(Manipulations::FIT_FILL, 256, 256)
                  ->background('white')
                  ->nonQueued();
         });
}
```
and I access the image with following code:
```php
$model->getFirstMediaUrl('image', 'resized');
```
Works great.

I was thinking about a way to return a default image when there is no image in the collection.
It's quite useful, for example when you use a seeder and don't mind to seed the images, or when you have a model where the image is required for your html template but you don't want to force the user to upload its own.

I know I can manually check and use a default one with:
```php
$model->getFirstMediaUrl('image', 'resized') ?? asset('path/to/img');
```

But I was thinking about a better way, like defining a default file in `registerMediaCollections` method.

sebastiandedeyne:
Sounds like a solid request! Feel free to send a PR that implements this. 

The fallback should be some sort of `Media` null object, could be tricky to pull through but we'll see what it looks like when implemented.

FaridAghili:
I've got an idea about how this could be implemented:

We set the default file for each collection (or maybe for each conversion as well) like this:
```PHP
public function registerMediaCollections()
{
    $this->addMediaCollection('image')
        ->singleFile()
        ->defaultFile('path/to/file/which/will/be/returned/as/it/is')
        // This library won't touch the default file, it just get a file name
        // as string and return it exactly as it is
        // It doesn't care about file type, or if it exists or not
        // ->defaultFile('images/user/default.png')
        ->registerMediaConversions(function (Media $media) {
            $this->addMediaConversion('resized')
                ->fit(Manipulations::FIT_FILL, 256, 256)
                ->background('white')
                ->nonQueued();
        });
}
```
Then there will be a new set of methods to retrieve file:
`->getMediaOrDefault()`
`->getFirstOrDefaultMedia()`
`->getFirstOrDefaultMediaUrl()`

So, we can use this methods in our frontend code to get the media (or the default one if there is none), and we can continue to use current methods (`getMedia`, `getFirstMedia`, `getFirstMediaUrl`) in our admin panel because we don't need the default file in there.

[...]

PR #1382, opened 2019-02-06
Add fallback path and URL to media collections

jameslkingsley:
I needed this functionality but it doesn't look like #1101 made any progress, so here's a PR for the functionality outlined in that issue.

### DIFF
--- a/src/HasMedia/HasMediaTrait.php
+++ b/src/HasMedia/HasMediaTrait.php
@@ -239,7 +239,7 @@ trait HasMediaTrait
         $media = $this->getFirstMedia($collectionName);
 
         if (! $media) {
-            return '';
+            return $this->getDefaultMediaUrl($collectionName) ?: '';
         }
 
         return $media->getUrl($conversionName);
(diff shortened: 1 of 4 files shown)


ArmeniaH:
bump

clemjlt:
Hi ! What's the status on this one ? It would be pretty useful :)

freekmurze:
Thanks!

nelson6e65:
Hmm... Seems like this only works for `getFirstMediaUrl()` and `getFirstMediaPath()` methoda, but `getFirstMedia()` still returning null. :sweat_smile:

freekmurze:
I don't have time to look into this now, but if you PR a fix with tests, i'll likely merge it.

morawcik:
Any info about default image for getFirstMedia() and getMedia() ? Or any advice how to use default image with responsive images?

clemjlt:
`getFirstMediaUrl()` asks media library for a `Media` using `getMedia()` behind the scenes, and returns the fallback url if `getMedia()` returns `null`. 
Given this, it's impossible to get a default value when you use `getMedia()` because the check to return the default url is implemented after `getMedia()` is called (in `getFirstMediaUrl()`)

A solution to make this work could be to move the check in `getMedia()` instead of `getMediaUrl()`, and use a default media instead of a default Url. But I think it's not in the scope of this PR ...

PR #1386, opened 2019-02-12
Add getFirstMediaFullUrl

shaffe-fr:
Hi,
The getFirstMediaFullUrl is especially handy when the project is not at the document root of the server.
Thanks

### DIFF
--- a/src/HasMedia/HasMediaTrait.php
+++ b/src/HasMedia/HasMediaTrait.php
@@ -246,6 +246,22 @@ trait HasMediaTrait
     }
 
     /*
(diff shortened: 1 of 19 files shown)


shaffe-fr:
I updated the tests suite to comply with the $@¤$!@ change in PHPUnit 8 where the return type of the setUp method has been strongly typed.
https://phpunit.de/announcements/phpunit-8.html

shaffe-fr:
I fixed the test that I could but there are still errors in the vendor libraries.

PR #1372, opened 2019-01-29
fix call to undefined method FFMpeg::getDuration() #1371

acodeninja:
This pr fixes issue #1371

No tests needed as this fixes failing tests:

* `Spatie\MediaLibrary\Tests\Feature\Conversion\ConversionFileExtensionTest::it_always_defaults_to_jpg_when_the_original_file_is_not_an_image`
* `Spatie\MediaLibrary\Tests\Unit\ImageGenerators\VideoTest::it_can_convert_a_video`

Issue #1371, opened 2019-01-29
Call to undefined method FFMpeg\FFMpeg::getDuration()


acodeninja:
After setting up medialibrary for video thumbnailing I kept on getting the following exception in my conversion job.

```
Symfony\Component\Debug\Exception\FatalThrowableError: Call to undefined method FFMpeg\FFMpeg::getDuration() in /var/www/kiosk_manager/vendor/spatie/laravel-medialibrary/src/ImageGenerators/FileTypes/Video.php:23
Stack trace:
#0 /var/www/kiosk_manager/vendor/spatie/laravel-medialibrary/src/FileManipulator.php(94): Spatie\MediaLibrary\ImageGenerators\FileTypes\Video->convert('/var/www/kiosk_...', Object(Spatie\MediaLibrary\Conversion\Conversion))
#1 /var/www/kiosk_manager/vendor/laravel/framework/src/Illuminate/Support/Collection.php(418): Spatie\MediaLibrary\FileManipulator->Spatie\MediaLibrary\{closure}(Object(Spatie\MediaLibrary\Conversion\Conversion), 0)
#2 /var/www/kiosk_manager/vendor/spatie/laravel-medialibrary/src/FileManipulator.php(117): Illuminate\Support\Collection->each(Object(Closure))
#3 /var/www/kiosk_manager/vendor/spatie/laravel-medialibrary/src/Jobs/PerformConversions.php(32): Spatie\MediaLibrary\FileManipulator->performConversions(Object(Spatie\MediaLibrary\Conversion\ConversionCollection), Object(Spatie\MediaLibrary\Models\Media))
#4 [internal function]: Spatie\MediaLibrary\Jobs\PerformConversions->handle()
#5 /var/www/kiosk_manager/vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php(29): call_user_func_array(Array, Array)
#6 /var/www/kiosk_manager/vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php(87): Illuminate\Container\BoundMethod::Illuminate\Container\{closure}()
#7 /var/www/kiosk_manager/vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php(31): Illuminate\Container\BoundMethod::callBoundMethod(Object(Illuminate\Foundation\Application), Array, Object(Closure))
[...]

PR #1412, opened 2019-03-25
fix video duration

jclappiway:
fix `Call to undefined method FFMpeg\FFMpeg::getDuration()`


acodeninja:
I've submitted a slightly different code change to resolve this issue. PR #1372 and discussed on issue #1371 might save you from trying to fix those build errors.

jclappiway:
@acodeninja thanks, I wasn't looking to fix build errors since they're not connected to this PR.

riasvdv:
Added in #1372

Issue #1179, opened 2018-07-18
Removing '-' from conversion files


jwtea:
Hi, firstly thank you all for this library it is a great package and i've been enjoying using it.

I'm trying to find out how you can remove the appended '-' from the conversion files.

I initially tried overriding the FileManipulator and UrlGenerator\BaseUrlGenerator classes where i found the '-' being attached. However when doing this it causes the rename to not trigger the rename on converted files. 

Does anyone have any advice on how to accomplish this ?
Thanks again.

jwtea:
Ah i found another reference in the Conversion class getConversionFile function. 
Would there be any interest in making this configurable?

brendt:
Hi @jwtea 

Technically, it's already possible to change this behaviour on a model basis, it's just not documented. Conversions are created in the `HasMediaTrait`, in the `addMediaConversion` function (https://github.com/spatie/laravel-medialibrary/blob/master/src/HasMedia/HasMediaTrait.php#L412).

You could extend `\Spatie\MediaLibrary\Conversion\Conversion`, and override the `addMediaConversion` function in the model you're want this custom conversion to be used.

Note that this is no tested behaviour, but it should work from glancing at the code. I don't think we want to make conversions configurable, unless there's a good reason to do so. You say you don't want hyphens (-) in the file names, but the media library is designed in such a way that you should never worry about files themselves. So I'm interested to hear what's the reason behind your request.

I'm closing the issue for now, as we don't have plans to change this behaviour, but feel free to comment back here if you have any more thoughts or remarks.

jwtea:
Thanks @brendt for taking the time to reply to this. 
So the reasoning for doing this was for the way that images are loaded on IOS using the @2x and @3x notation for screen sizes so that images are displayed in the correct resolution.
While its fine to have the -@2x, it means that the file must be named file-.ext which just bugged me. This is the only reason i've had for this. 
I understand that this is not going to be a common use case for most people.

brendt:
Hi @jwtea 

Do you mean Safari on IOS? `srcset` is supported on all browsers except IE and Opera Mini: https://caniuse.com/#search=srcset

[...]

PR #1413, opened 2019-03-26
Moving Hard-Coded Hyphen to Configuration File

josemanuelsh:
Overriding the hyphen used when creating conversions can be a big problem.
This can be easily fixed by replacing the hard-coded hyphen used in three
functions with a configuration variable. This may also help for future development.
I ran into a similar situation as @jwtea in issue #1179


freekmurze:
To keep the code simple, I'm going to pass on this for now. 

When creating the next version of medialib, later this year, I'll make the naming the file more configurable.

PR #1418, opened 2019-04-06
Accept sync queue on config file

skalero01:
Right now i have as default queue value the database, i was trying to make the jobs from this package to execute inmediately with sync but it wasnt successfull.

So this pull request will help that anyone that puts `'queue_name' => 'sync',` on the config file will make the jobs from the package to be executed inmediately and not by a queue.

Hope you accept this pull request.
Regards


freekmurze:
If you use the `sync` driver, then  `dispatch` will execute the job immediately. I don't think the changes of this PR are necessary.

PR #1420, opened 2019-04-09
Optionally prepend original filename to conversions

pascalbaljet:
This PR makes the prepending of the original filename to conversions configurable. I was upgrading from v6 to v7 when I discovered all conversions were prepended with the original filename. The config setting is set to `true` by default so the behavior only changes whenever the setting is adjusted.


vesper8:
what is your workaround for this issue @pascalbaljet short of using a fork ? I just upgraded to v7 as well and found that all my urls are now wrong because getUrl() is appending the original filename even though the files themselves do not include this.. result: total breakage of my app

pascalbaljet:
@vesper8 yes still using the fork :)

vesper8:
I wrote a tiny script that renames all the conversions to the new format.. I was able to rename 75k conversions in under 30 seconds... must faster than regenerating which would have taken over an hour

```
                $media = Media::all();

                $conversions = [
                    'map',
                    'thumb',
                    'optimized',
                ];

                foreach ($media as $media) {
                    foreach ($conversions as $conversion) {
                        $oldPath = "public/media/{$media->id}/conversions/{$conversion}.jpg";
                        $newPath = "public/media/{$media->id}/conversions/{$media->name}-{$conversion}.jpg";

                        if (\Storage::exists($oldPath)) {
                            \Storage::move($oldPath, $newPath);
                        }
                    }
                }
```

freekmurze:
In the interest of keeping the codebase simple, I'm going to pass on this one.

Thanks!

PR #1421, opened 2019-04-12
Fix ->extractVideoFrameAtSecond(10)

huglester:
getDuration for videos should be taken via FFProbe, not FFMpeg directly.

As an example look here:
https://github.com/PHP-FFMpeg/PHP-FFMpeg/blob/master/src/FFMpeg/FFProbe.php#L185


acodeninja:
Actioned this already under #1372 👍

riasvdv:
Added in #1372

PR #1424, opened 2019-04-15
Add environment variable to change disk

Christophvh:
Ability to switch disk on different servers or environments.


freekmurze:
Thanks!

PR #1452, opened 2019-05-21
Zipstream Dependency - optionally allow later version of zipstream

rossriley:
This PR just allows the library to be more flexible with its zipstream dependency.

Current constraint is fixed to pre-release version which is now over a year out of date.

Since this library uses just the basic features of the ZipStream class, the usage is compatible with both 0.x versions and 1.x versions, adding either as a dependency allows users to specify either in their local `composer.json` file.

All tests pass with either version installed, currently tested 0.5.2 and 1.1 which are the latest versions on both constraints.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
     "illuminate/pipeline": "~5.5.0|~5.6.0|~5.7.0|~5.8.0",
     "illuminate/support": "~5.5.0|~5.6.0|~5.7.0|~5.8.0",
     "league/flysystem": "^1.0.8",
-    "maennchen/zipstream-php": "~0.4",
+    "maennchen/zipstream-php": "^0.4|^1.0",
     "spatie/image": "^1.4.0",
     "spatie/pdf-to-image": "^1.2",
     "spatie/temporary-directory": "^1.1"


freekmurze:
@rossriley Could you manually check if the produced zip is valid when it  comes from local systems and S3. Previous version were suposed to work but produced current ZIP file.

reppair:
any updates on this one?

freekmurze:
@reppair Could you manually check if the produced zip is valid when it comes from local systems and S3. Previous version were supposed to work but produced an invalid ZIP file.

reppair:
I’ll take a look, btw I tested today the StreamResponse::create() with the current implementation and it creates an .zip that works fine windows but when opened under Ubuntu and OS X it results in a .zip.cpgz then that one when extracted results in the same .zip and it loops like this forever. Is that what you mean by invalid ZIP?

freekmurze:
Yup, seems like we can’t upgrade to the latest version of Zipstream yet.

reppair:
@freekmurze I have cloned the `rossriley:feature/upgrade-zipstream` and added it as a local composer repository to a new laravel `5.8.28` project. I have made my User implement the `HasMedia` trait and added some local media to it. Then simply did:

```
Route::get('/', function () {
    return MediaStream::create('test.zip')->addMedia(User::first()->media);
});
```

And it returns a WORKING archive that can be opened without any problems. Tested the download from 3 local machines (windows, os x & Elementary OS - an Ubuntu basically) - all good.

Then I removed the  local media from the User and setup Amazon S3 filesystem and set the medialibrary to use `'disk_name' => 's3',`. Having the new setup and no media assigned to the User model I added new one and it took some time to go to aws S3. After this tested again from all 3 different machines (OSs) and it works as expected!

If this is all you need in order to merge this one I think you are good to go. Hope this is helpful.

reppair:
BTW if it is any help the broken archive issue disappears if `ob_clean()` is used before returning the `MediaStream` - I will be using this work around until this gets merged.

freekmurze:
Is using `ob_clean()` necessary when using `rossriley:feature/upgrade-zipstream`?

reppair:
No, that is with the current implementation, with `rossriley:feature/upgrade-zipstream` I did NOT use it.

freekmurze:
Thanks!

PR #1470, opened 2019-06-19
Filtering and collection name

matevisky:
These option for these helpers are super handy, we do not have to write custom collection filters...


freekmurze:
Going to leave it like it is for now. Thanks!

PR #1477, opened 2019-06-21
Fix absolute references to media.id

ebaliatico:
Numerous parts of the library break when the media model's primary key is not named id. I did not address the tests.


freekmurze:
Thanks!

PR #1479, opened 2019-06-24
Added dependency on the PHP fileinfo extension.

gerben-van-eck:
Added dependency on the PHP fileinfo extension. This is an extension which is enabled by default, but one of our customers had it disabled.
The dependency is used in (at least) src/Helpers/File.php


freekmurze:
Thanks!

PR #1489, opened 2019-07-02
Add HasAnyMedia functionality

ziming:
Sometimes we just need to check if a record has any media attached to it. Regardless of which collection

Feel free to flatten it if it gets merged


riasvdv:
You can get the media count of a model by calling a count on the relationship:

`$model->media()->count()`

It's a simple enough check that doesn't warrant a change to the interface or trait.

ziming:
Ah didn't thought of that. Okay then :)

PR #1490, opened 2019-07-03
Change increments to bigIncrements in migrations

ziming:
In Laravel 5.8, id are bigIncrements by default in migrations so might be good to follow the conventions.


riasvdv:
Thanks!

PR #1508, opened 2019-07-29
fix wrong documentation for new fallback methods

faustbrian:
The methods names were wrongly documented, noticed it when I tried to use them and received an exception that the methods were undefined.


riasvdv:
Thanks!

PR #1509, opened 2019-07-30
Add event on clearMediaCollectionExcept

lloricode:


freekmurze:
This PR is wrong, this will fire the event for the collection which has not been cleared. Also, this event should fire if there is excluded media.

freekmurze:
Thanks for bringing this to my attention. I've fixed this myself.

Issue #1506, opened 2019-07-27
Media specific manipulations get ignored when applying to multiple Libraries on same model


hoofdletterj:
Hi @freekmurze, first of all: awesome package, it's been a life saver!

One issue though: When I have a model with more than one media library, each one having the same (two) conversions, and I apply a media specific manipulation on all (first) images of a specific conversion, only the conversion of the first media library gets manipulated. 

**Example**
- I've added three libraries to a Dog: front, side and rear view. 
- On all libraries I have a "thumb" and "square" conversion.
- I've populated all three libraries with one image.
- I create a media specific manipulation (rotation) on the "square" conversion for the image in every library.

**Outcome**
- The manipulations for all the relevant items get correctly added to the database in the manipulations column.
- All three images (front/square, side/square and rear/square) get re-generated with the specified conversions.
- Only the first (front/square) is rotated.
- The other two images have a new timestamp but are unmodified relative to the initially generated image.

**The code I'm using**
(please note: this is a reproduced simplified version of a project ;) )
```
<?php

namespace App;

use Spatie\Image\Manipulations;
use Spatie\MediaLibrary\Models\Media;
use Illuminate\Database\Eloquent\Model;
use Spatie\MediaLibrary\HasMedia\HasMedia;
use Spatie\MediaLibrary\HasMedia\HasMediaTrait;

class Dog extends Model implements HasMedia
{
    use HasMediaTrait;

    public function registerMediaCollections()
    {
        $this
            ->addMediaCollection('front')
            ->registerMediaConversions(function (Media $media) {
                $this->addMediaConversion('square')
                    ->crop(Manipulations::CROP_CENTER, 300, 300);
                $this->addMediaConversion('thumb')
                    ->crop(Manipulations::CROP_CENTER, 600, 800);
            });
        $this
            ->addMediaCollection('side')
            ->registerMediaConversions(function (Media $media) {
                $this->addMediaConversion('square')
                    ->crop(Manipulations::CROP_CENTER, 300, 300);
                $this->addMediaConversion('thumb')
                    ->crop(Manipulations::CROP_CENTER, 600, 800);
            });
        $this
            ->addMediaCollection('rear')
            ->registerMediaConversions(function (Media $media) {
                $this->addMediaConversion('square')
                    ->crop(Manipulations::CROP_CENTER, 300, 300);
                $this->addMediaConversion('thumb')
                    ->crop(Manipulations::CROP_CENTER, 600, 800);
            });
    }

[...]

PR #1510, opened 2019-07-30
Custom manipulations get applied to all relevant conversions with the same name

hoofdletterj:
Fixed a bug where custom manipulations on a conversion get ignored after the first Media item, when manipulations are applied to multiple Media items with the same conversion name.

See issue #1506


freekmurze:
Thanks for this. I'll merge this in, polish it a bit and tag a new release.

PR #1512, opened 2019-07-30
Add Macroable trait to Media Collection

peterfox:
This would be be really useful non breaking change that for me would mean I can tweak some of the logic like the disk selection with a simple macro rather than extending multiple files or having a lot of duplicated code for interacting with the Media Collection's disk use.


freekmurze:
Thanks!

PR #1523, opened 2019-08-06
Make FileAdder macroable

nmfzone:
With this approach, we can add some additional functionality to the FileAdder on the fly.
For example to always set filename using Uuid4, etc.

I always set filename using uuid4, so I've something like this everywhere:
```
$user->addMedia($file)
    ->usingFileName(Uuid::uuid4()->toString() . '.' . $file->guessExtension())
    ->toMediaCollection('files');
```

If the FileAdder is macroable, I can make my code little cleaner.


freekmurze:
Thanks!

freekmurze:
Could you PR our docs with a little example on how to add an extra method to `FileAdder`? Thanks!

nmfzone:
@freekmurze Thanks! Sure if I could..

PR #1528, opened 2019-08-16
Update introduction.md

OmarMakled:
You should add `HasMediaTrait` trait into a modal


OmarMakled:
Already exists under prepare a model

PR #1532, opened 2019-08-18
fix #1258

### LINKED ISSUE #1258: $media->move() raises an error if the filename has no extension
If I try to move a media with a filename that doesn't contain an extension, I receive an error similar to the following:

`fopen(/myproject/storage/medialibrary/temp/SRzedmRMFTWXhiovt208SvevwZYiVacN/1538583003_2pMnshlcV2): failed to open stream: Is a directory in /myproject/vendor/spatie/laravel-medialibrary/src/Filesystem/Filesystem.php on line 89`

VincentKos:
Hello,

Basicly instead of letting `TemporaryDirectory::path()` create the temporary directory I manually call `TemporaryDirectory::create()` and use `TemporaryDirectory::path()` without argument to build `$temporaryFile`. 

I didn't write any test because I don't think it's needed.

### DIFF
--- a/src/Models/Media.php
+++ b/src/Models/Media.php
@@ -329,8 +329,9 @@ class Media extends Model implements Responsable, Htmlable
     public function copy(HasMedia $model, $collectionName = 'default'): self
     {
         $temporaryDirectory = TemporaryDirectory::create();
+        $temporaryDirectory->create();
 
-        $temporaryFile = $temporaryDirectory->path($this->file_name);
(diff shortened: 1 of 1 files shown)


VincentKos:
I spend the whole day trying to get imagick to work and I can't. It's installed and active but i still get the error ` - The requested PHP extension ext-imagick * is missing from your system. Install or enable PHP's imagick extension.` 

I can't do much more 😒

Issue #1524, opened 2019-08-08
Limit number of images in collection


jam1e:
Is there a simple way to limit the number of images a model can hold in a given collection?

For example, my Business model has one logo, which I have used `->single()` for, and up to 3 banner images. I would like to limit it so they cannot have more than 3, and any additional uploads would replace older ones.

freekmurze:
I'd accept a pull request that adds a `onlyKeepLatest($int)` function on `MediaCollection`. When adding a new item to the collection only the latest `$int` number of media models in it are kept, the rest is deleted. The PR should have updated test and updated docs. The `singleFile` method should, under the hood, call `onlyKeepLatests(1)`.

joshuadegier:
Working on it! Got the code ready, heading over to writing the tests in a minute.

freekmurze:
This has been added in the latest released. Thanks for the PR @joshuadegier

PR #1533, opened 2019-08-21
Added option to set a limit for MediaCollections

joshuadegier:
As requested here: https://github.com/spatie/laravel-medialibrary/issues/1524

This pull-request adds the `onlyKeepLatest(n)` method for MediaCollections, where the existing `singleFile()` method uses the new method under the hood. Whenever adding a new file to the collection exceeds the set limit, Medialibrary will delete the oldest file(s) and only keep the latest `n`.

The tests involved passed, I did however got 2 failures that involved Ghostscript and some of the tests skipped. As far as I can tell none of these tests involve the parts I've been working on.

And as request, the docs have been editted as well.

### DIFF
--- a/docs/working-with-media-collections/defining-media-collections.md
+++ b/docs/working-with-media-collections/defining-media-collections.md
@@ -95,9 +95,9 @@ You can still specify the disk name manually when adding media. In this example 
 $yourModel->addMedia($pathToFile)->toMediaCollection('big-files', 'alternative-disk');
 ```
 
-## Single file collections
+## Limited collections & single file collections
 
-If you want a collection to hold only one file you can use `singleFile` on the collection. A good use case for this would be an avatar collection on a `User` model. In most cases you'd want to have a user to only have one `avatar`.
+If you want a collection to hold only `n` file(s) you can use `onlyKeepLatests(n)` on the collection. A good use case for this would be an avatar collection on a `User` model. In most cases you'd want to have a user to only have one `avatar`. Whenever you add a file and the collection exceeds the limit, Medialibrary will delete the oldest files first and keep the latest file(s).
 
 ```php
 // in your model
@@ -106,9 +106,11 @@ public function registerMediaCollections()
 {
     $this
         ->addMediaCollection('avatar')
-        ->singleFile();
+        ->onlyKeepLatests(1);
 }
 ```
+
+You can also use the shorthand to `singleFile()` which is an alternative to `onlyKeepLatest(1)`.
 
 The first time you add a file to the collection it will be stored as usual.
 
--- a/docs/working-with-media-collections/defining-media-collections.md
+++ b/docs/working-with-media-collections/defining-media-collections.md
@@ -95,9 +95,9 @@ You can still specify the disk name manually when adding media. In this example 
 $yourModel->addMedia($pathToFile)->toMediaCollection('big-files', 'alternative-disk');
 ```
 
-## Limited collections & single file collections
+## Single file collections
 
-If you want a collection to hold only `n` file(s) you can use `onlyKeepLatests(n)` on the collection. A good use case for this would be an avatar collection on a `User` model. In most cases you'd want to have a user to only have one `avatar`. Whenever you add a file and the collection exceeds the limit, Medialibrary will delete the oldest files first and keep the latest file(s).
+If you want a collection to hold only one file you can use `singleFile` on the collection. A good use case for this would be an avatar collection on a `User` model. In most cases you'd want to have a user to only have one `avatar`.
 
 ```php
 // in your model
@@ -106,11 +106,9 @@ public function registerMediaCollections()
 {
     $this
         ->addMediaCollection('avatar')
-        ->onlyKeepLatests(1);
+        ->singleFile();
 }
 ```
-
-You can also use `singleFile()` which is an alternative to `onlyKeepLatest(1)`.
 
 The first time you add a file to the collection it will be stored as usual.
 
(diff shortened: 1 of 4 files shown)


docs/working-with-media-collections/defining-media-collections.md
```diff
@@ -106,10 +106,12 @@ public function registerMediaCollections()
 {
     $this
         ->addMediaCollection('avatar')
-        ->singleFile();
+        ->onlyKeepLatests(1);
```
Gummibeer: SHouldn't it stay `->singleFile()` and add an example with `->onlyKeepLatests(5)` to showcase the difference?

tests/Feature/FileAdder/MediaConversions/MediaCollectionTest.php
```diff
@@ -170,9 +170,32 @@ public function registerMediaCollections()
...
 
+        $model->addMedia($this->getTestJpg())->preservingOriginal()->toMediaCollection('images');
         $model->addMedia($this->getTestJpg())->preservingOriginal()->toMediaCollection('images');
         $model->addMedia($this->getTestJpg())->preservingOriginal()->toMediaCollection('images');
 
         $this->assertCount(1, $model->getMedia('images'));
     }
+
+    /** @test */
+    public function if_the_only_keeps_latest_method_is_specified_it_will_delete_all_other_media_and_will_only_keep_the_latest_n_ones()
```
Gummibeer: Could this test get extended by any custom property which clearly identifies the four images to check that the latest 3 are still there? 
The current test only checks the count - which was okay for the single one, because all got deleted except the added one. But now we want to really just delete the first one.

docs/working-with-media-collections/defining-media-collections.md
joshuadegier: Check. Added a seperate example in the docs as "Limited file collections" under "Single file collections".

Gummibeer: https://github.com/spatie/laravel-medialibrary/pull/1533/commits/026b43f4e479b0cd574633d081cbe7d6189b9cbc

tests/Feature/FileAdder/MediaConversions/MediaCollectionTest.php
joshuadegier: Added an assertion which checks wether the first file is no longer part of the returned collection of Media.

Gummibeer: https://github.com/spatie/laravel-medialibrary/pull/1533/commits/3fdf17ee03d635c396c93bc37067209f4cce2bfd

src/MediaCollection/MediaCollection.php
```diff
@@ -20,8 +20,8 @@ class MediaCollection
     /** @var callable */
     public $acceptsFile;
 
-    /** @var bool */
-    public $singleFile = false;
```
Rigby90: As this and the Media Collections are public, is this not a potential problem for bc?

joshuadegier: I don't think so. This variable was only set in `MediaCollection.php` (via the `singleFile()` method) and checked in `FileAdder.php`. If you would upgrade to this version, everything gets converted to these changes under the hood and don't need any additional changes in the public codebase.

Rigby90: Apologies, I should have provided an example.

In my models it is possible for me to do the below currently

`if($this->mediaCollections[0]->singleFile) {
            //
        }`

As `$mediaCollections` is public and so is `$singleFile`.

However renaming `$singleFile` to `$collectionSizeLimit` would result in an error in any code which did read the variable directly via it's public access.

joshuadegier: My bad, misunderstood what you meant. In that case, yes.. that could cause applications to crash. Not sure what is the right way to go, my feeling says:

1. Restore `$singleFile` as public variable to `MediaCollection.php`
2. Whenever `onlyKeepsLatest()` is called set the variable to true if `n=1`

This could later be refactored and removed in a new release. But I leave that up to @freekmurze and the people maintaining this package to decide.

Gummibeer: I would restore it and flag it as `@deprecated` to hint all IDEs. Another way would be to implement a `__get()` method which returns this value on a dynamic approach - because in theory you could also do `$mediaCollection-> collectionSizeLimit = 5;` because it's public. And having it overloaded would also prevent it from getting deprecated because the `singleFile()` method will also stay there!? So there should be a corresponding getter/property.

Rigby90: Was just seeing what it would take to implement something similar to the below on the setter -

```
if($int === 1) {
    $this->singleFile = true;
} else {
    $this->singleFile = false;
}

$this->collectionSizeLimit = $int;
```

But, after mulling it over. Perhaps as @Gummibeer mentioned a `__get()` method would be better; when trying to obtain the `$singleFile` value, `if($collectionSizeLimit === 1) { return true; }` etc.

Unsure as to whether the below would be required as well on the setter? Without looking into it I'm not sure how it would handle a 0 or negative value?

```
public function onlyKeepLatest(int $int): self
{
    if ($int <= 0) {
        // Should an error be thrown??? Should it default to a value???
    }
    ...
}
```

src/MediaCollection/MediaCollection.php
```diff
@@ -20,8 +20,8 @@ class MediaCollection
...
+    /** @var int */
```
Gummibeer:
```suggestion
    /** @var int|false */
```

src/MediaCollection/MediaCollection.php
Gummibeer: For the `$mediaCollection->singleFile` I would go with the `__get` - it's the cleanest solution and handles all circumstances.

The below `1` values are also an interesting special case - if using a default value we should switch to `false` (no limit). But throwing an `OutOfBoundsException` seems better to prevent doing some "magic" if it's just the result of a wrong written math logic. So the user should handle it.

But I would use `0` to remove the limit - so set it to `false`. Because otherwise we won't have a way to remove it again.

```php
if($int === 0) {
  $this->collectionSizeLimit = false;
}

if($int < 0) {
  throw new \OutOfBoundsException();
}

$this->collectionSizeLimit = $int;
```

joshuadegier: Should we also make the `$collectionSizeLimit` private on `MediaCollection.php`? I see no reason why it should be accessible or overwritable on runtime since that defeats the purpose it has in the first place.

Gummibeer: Because there are no getters for all the properties they are all `public`. Otherwise I can't use this value in my own logic extending the package.

But making them all private and adding corresponding getter methods would be a BC.

So far I know you also can't overload/read private properties via `__get`!? If this would be possible I would make them `private` and make them available via `__get`. If this sounds like a good idea but using the same name isn't possible we could get dirty and prefix the real properties with `_`.

joshuadegier: Point taken, I won't change that part. Working on the BC fix for `$singleFile` right now.

freekmurze:
Thanks, I'll merge this in, make sure the `singleFile` has the right property, mark it as deprecated an tag a new release.

joshuadegier:
Sorry, not 100% sure what I am expected to do now :-) I've restored the `$singleFile` property as such:

```php
    /** @deprecated @var bool */
    public $singleFile = false;
```

Anything else?

freekmurze:
I've already taken care of it! Thanks again for this PR.

Gummibeer:
@freekmurze this doesn't handle the integer edge cases and there's no way to remove a size constraint.

@Rigby90 has found some valuable cases which should get solved before releasing it.

freekmurze:
> no way to remove a size constraint.

Why would you want to remove a constraint after setting one?

Gummibeer:
Because they are an object which could be manipulated all the time.
For example the size is limited by some kind of subscription plan. It could be necessary to remove a limit added before.

freekmurze:
I think that after setting a limit, most people don’t want to remove it. I’ll reconsider my stance on this if issues of real life projects should start coming in.

Gummibeer:
But it's just an additional `if` to prevent invalid values at all with a little benefit.

PR #1534, opened 2019-08-21
Added option to delete media when parent model is SoftDeleted

### LINKED ISSUE #1474: clearMediaCollection() doesn't able to force delete if using softDeletes.
clearMediaCollection() doesn't able to force delete the database record and hence the files also not get removed from the storage. I'm using Laravel's softDeletes trait and while using the clearMediaCollection() method the record in the database gets soft deleted instead of forceDelete(). I think you should change the trait a little bit like what I do:

if (in_array(SoftDeletes::class, class_uses_recursive($entity))) {
    $this->getMedia($collectionName)
        ->each->forceDelete();
} else {
    $this->getMedia($collectionName)
        ->each->delete();
}

joshuadegier:
Fixes: https://github.com/spatie/laravel-medialibrary/issues/1474

I've added a new property to the `HasMediaTrait` named `$shouldDeleteMediaWhenSoftDeleting`, this property defaults to `false` in which case the Medialibrary works as it does right now.

As soon as `$shouldDeleteMediaWhenSoftDeleting` is set to true and the parent model is set to be softDeleteable, Medialibrary will delete all media belonging to the model anyway.

A use case for this scenario would be a system which holds user data (name, address, avatar, etc.) and you only want to keep the data but not the avatar and other collected media files. By setting `$shouldDeleteMediaWhenSoftDeleting` to true on the `User` model, the files would be deleted at soon as you softDelete the user.

### DIFF
--- a/src/HasMedia/HasMediaTrait.php
+++ b/src/HasMedia/HasMediaTrait.php
@@ -24,6 +24,9 @@ use Spatie\MediaLibrary\Exceptions\FileCannotBeAdded\MimeTypeNotAllowed;
 
 trait HasMediaTrait
 {
+    /** @var $forceDeleteMediaWhenModelIsSoftDeleted */
+    protected $forceDeleteMediaWhenModelIsSoftDeleted = false;
+
     /** @var Conversion[] */
     public $mediaConversions = [];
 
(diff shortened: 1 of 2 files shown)


freekmurze:
Going to pass on this, see #1474

PR #1539, opened 2019-08-25
fixed method name typo in documentation for onlyKeepLatest method

zaffron:
The original name of the method according to the source code of MediaCollection class is:
`onlyKeepLatest`
Method name used in the documentation: `onlyKeepLatest`


freekmurze:
Thanks!

PR #1544, opened 2019-08-27
Change to registerMediaConversions

EricBusch:
I think `registerModelConversions` should be `registerMediaConversions`. At least I couldn't find `registerModelConversions` anywhere.


freekmurze:
Good catch, thanks!

PR #1545, opened 2019-08-28
the properties default to null

Gummibeer:
and don't are set in the construct so they could stay `null`


freekmurze:
I want this to be slightly different, I'll quickly take care if it myself. Thanks!

PR #1569, opened 2019-09-24
Add url versioning based on media updated_at

riasvdv:
- This adds a config key `medialibrary.versioning`
- When set to `true`, adds the `updated_at` timestamp of the media model to the image url

### DIFF
--- a/config/medialibrary.php
+++ b/config/medialibrary.php
@@ -77,6 +77,12 @@ return [
     'url_generator' => null,
 
     /*
+     * Whether to activate versioning when urls to files get generated.
+     * When activated, this attaches a ?v=xx query string to the URL.
+     */
+    'versioning' => false,
+
+    /*
      * The class that contains the strategy for determining a media file's path.
      */
     'path_generator' => null,
--- a/config/medialibrary.php
+++ b/config/medialibrary.php
@@ -80,7 +80,7 @@ return [
      * Whether to activate versioning when urls to files get generated.
      * When activated, this attaches a ?v=xx query string to the URL.
      */
-    'versioning' => false,
+    'version_urls' => false,
 
     /*
      * The class that contains the strategy for determining a media file's path.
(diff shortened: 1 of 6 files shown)


config/medialibrary.php
```diff
@@ -76,6 +76,12 @@
      */
     'url_generator' => null,
 
+    /*
+     * Whether to activate versioning when urls to files get generated.
+     * When activated, this attaches a ?v=xx query string to the URL.
+     */
+    'versioning' => false,
```
freekmurze: Rename this to `version_urls`

docs/basic-usage/retrieving-media.md
```diff
@@ -20,6 +20,8 @@ $fullPathOnDisk = $mediaItems[0]->getPath();
 $temporaryS3Url = $mediaItems[0]->getTemporaryUrl(Carbon::now()->addMinutes(5));
 ```
 
+If you want to retrieve versioned media urls you can enable versioning by setting the `versioning` config to `true` in your `medialibrary.php` config file.
```
freekmurze: Maybe add a few sentences on why you'd want to use this features (cache busting).

src/UrlGenerator/BaseUrlGenerator.php
```diff
@@ -83,4 +83,13 @@ public function rawUrlEncodeFilename(string $path = ''): string
     {
         return pathinfo($path, PATHINFO_DIRNAME).'/'.rawurlencode(pathinfo($path, PATHINFO_BASENAME));
     }
+
+    public function versionUrl(string $path = ''): string
+    {
+        if (! $this->config->get('medialibrary.versioning', false)) {
```
freekmurze: I don't think the `false` default is needed, because the config of the project gets merged with the file in the package. I guess all the tests will still be green when you remove the default here.

freekmurze:
After you've taken care of the remarks, go ahead an tag a new version.

lloricode:
Hi!

in what scenario use this? im just curious

thank you

riasvdv:
For when you send cache headers for images, but want to refresh that cache as soon as an image changes

jorenvh:
It seems it does not support responsive images, is there a specific reason for this? If not I will create a PR to add it for responsive images.

riasvdv:
@jorenvh Feel free to create a PR 👍

PR #1570, opened 2019-09-24
Add a way to define accepted mime types

riasvdv:
- Adds a `acceptsMimeTypes()` function to the media collection
- Checks based on the files mimetype if it's allowed in the collection


freekmurze:
Perfect, thanks!

PR #1582, opened 2019-10-03
Add countMedia function in the hasMediaTrait.

bramr94:
I needed this functionality a lot lately so I thought I would submit a pull request for it.


freekmurze:
Thank you for this, not going to merge it because you can easilyt do this:

```
$this->getMedia($collectionName)->count();
```

PR #1585, opened 2019-10-07
Fixes #1258 when the file doesn't have extension

### LINKED ISSUE #1258: $media->move() raises an error if the filename has no extension

daniel-werner:
The temporary directory cannot identify if it is a file path, when the file doesn't have extension, and it creates a directory with the file's name.

It was tricky to reproduce the issue with tests: 
If the temporary directory path contained `../`, the bug didn't appear, because the isFilePath returned true even for files without extension. I've added the absolute path of the temporary directory to the config `medialibrary.temporary_directory_path`, this way I was able to reproduce the issue.

### DIFF
--- a/src/Models/Media.php
+++ b/src/Models/Media.php
@@ -330,7 +330,7 @@ class Media extends Model implements Responsable, Htmlable
     {
         $temporaryDirectory = TemporaryDirectory::create();
 
-        $temporaryFile = $temporaryDirectory->path($this->file_name);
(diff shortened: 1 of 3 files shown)


rubenvanassche:
Looks great! Thanks!

Issue #1586, opened 2019-10-07
The tests are failing


freekmurze:
This should be fixed.

freekmurze:
Test are still failing...

riasvdv:
Tests are no longer failing!

Issue #1581, opened 2019-10-02
Upload files from a non-local disk


liran-co:
It doesn't seem possible to use `addMedia` with files from non-local disks, such as s3. Is this intentional, or am I doing something wrong?

freekmurze:
Sure you can, take a look at the example here: https://docs.spatie.be/laravel-medialibrary/v7/advanced-usage/working-with-multiple-filesystems/

jameslkingsley:
Having the same issue - though the issue is not uploading **to** a separate disk, but uploading **from**. My use case right now is uploading an avatar via Laravel Vapor S3 stream (all handled on the client), and then on the server I want media-library to add the file that already exists in the S3 bucket in the `/tmp` directory. Something like `addMedia('tmp/something', 's3')` would be good.

liran-co:
@freekmurze, sorry if I wasn't clear. What @jameslkingsley described is what I'm looking for - uploading files **from** a non-local disk.

freekmurze:
Right... I think support for this should be added to the medialibrary. Currently I don't have time to look into it. If you need this quick, I'd accept a PR with tests and updated docs that adds support for this to the package.

jameslkingsley:
I'm currently working on getting a quick solution built into this package

martin-ro:
Any progress on that?

PR #1588, opened 2019-10-08
Add ability to upload files from a non-local disk

jameslkingsley:
**Work in Progress for #1581**
Pretty sure I'm doing some things wrong here, but at first try this solution worked perfectly.

This PR also fixes the Dotenv usage in the test suite (#1586).

**Usage:**
```
$request->user()
    ->addMediaFromDisk($request->key, 's3')
    ->preservingOriginal()
```

**Todo:**
- [x] Tests


src/FileAdder/FileAdder.php
```diff
@@ -206,8 +216,61 @@ public function toMediaCollectionOnCloudDisk(string $collectionName = 'default')
         return $this->toMediaCollection($collectionName, config('filesystems.cloud'));
     }
 
+    public function toMediaCollectionFromRemote(string $collectionName = 'default', string $diskName = ''): Media
+    {
+        $storage = Storage::disk($this->file->getDisk());
```
jameslkingsley: Is there a better way to instantiate `Storage` here?

src/Filesystem/Filesystem.php
```diff
@@ -33,6 +36,38 @@ public function add(string $file, Media $media, ?string $targetFileName = null)
...
+            $this->filesystem
+                ->disk($media->disk)
+                ->put($destination, $contents);
+
+            return;
+        }
+
+        $this->filesystem
+            ->disk($media->disk)
+            ->put($destination, $contents, $this->getRemoteHeadersForFile($file->getKey(), $media->getCustomHeaders(), $storage->mimeType($file->getKey())));
```
jameslkingsley: Is there a more optimised way of doing this - perhaps with a stream?

freekmurze:
Is this ready for review?

Could you rebase this with the current master, so all tests pass?

jameslkingsley:
Yea this is ready for review.

jameslkingsley:
One test is failing because it's using S3 instead of another disk - am fixing now.

jameslkingsley:
Is there anything else I need to do for this PR?

gdebrauwer:
Any idea when this feature might get merged/released? Really looking forward to using it 🤩

freekmurze:
Didn't have the time to review this one yet, I might be some time until I get around to this. Please have a little patience.

If you need this urgently, fork the package and temporarily use that fork.

src/Helpers/RemoteFile.php
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+     * Get the name (excluding extension).
+     *
+     * @return string
+     */
+    public function getName()
+    {
+        return pathinfo($this->getFilename())['filename'];
```
freekmurze: Use the second parameter of pathinfo to get the filename instead of retrieving a key from the result array.

src/Filesystem/Filesystem.php
```diff
@@ -33,6 +35,42 @@ public function add(string $file, Media $media, ?string $targetFileName = null)
...
+        $storage = Storage::disk($file->getDisk());
+
+        $destinationFileName = $targetFileName ?: $file->getFilename();
+
+        $destination = $this->getMediaDirectory($media, $type).$destinationFileName;
+
+        if ($file->getDisk() === $media->getDiskDriverName()) {
+            $storage->copy($file->getKey(), $destination);
+        } else {
+            $contents = $storage->get($file->getKey());
```
freekmurze: Wouldn't this be problematic when using very large files? Shouldn't we use streams here?

jameslkingsley: Updated

zetas:
This may not be appropriate here but I have to commend @jameslkingsley and @freekmurze. Adding a new feature in 2 weeks with this much work and communication is impressive. Good work guys! I can't wait for this to get merged.

src/Filesystem/Filesystem.php
```diff
@@ -33,6 +35,40 @@ public function add(string $file, Media $media, ?string $targetFileName = null)
...
+    public function copyToMediaLibraryFromRemote(RemoteFile $file, Media $media, ?string $type = null, ?string $targetFileName = null)
+    {
+        $storage = Storage::disk($file->getDisk());
+
+        $destinationFileName = $targetFileName ?: $file->getFilename();
+
+        $destination = $this->getMediaDirectory($media, $type).$destinationFileName;
+
+        if ($file->getDisk() === $media->getDiskDriverName()) {
+            $storage->copy($file->getKey(), $destination);
```
gdebrauwer: If the file on the disk has visibility set to private, then the copied file will also be private. Should the copied file not be public by default?

jameslkingsley: I'm not sure - probably a question for @freekmurze

jameslkingsley:
Noticing issues using this on a production Vapor app - I think when it's the same disk and it tries to copy, it's using the server as a middleman? Going to try and use stream for everything and see if that fixes it.

jameslkingsley:
Commit above has fixed one issue, but image conversions still fail to run on Vapor, even though I've set the media library temp directory to `/tmp` which according to AWS has 500 MB of space.

jameslkingsley:
Image conversion issue doesn't really exist in the scope of this PR -   however my solution is below for anyone who needs to have image conversions on serverless architecture.

This PR is ready for review again.

Serverless Image Conversions:

In order to handle image conversions on serverless architecture, I opted to move that processing to a separate [AWS Lambda function](https://aws.amazon.com/solutions/serverless-image-handler/) that will server on-demand conversions. Media library then has no registered media conversions, and instead the conversions are generated URLs from a simple function:

```php
public static function make(Media $media, string $size)
{
    $request = json_encode([
        'key' => $media->id . '/' . $media->file_name,
        'bucket' => config('filesystems.disks.s3.bucket'),
        'edits' => [
            'resize' => [
                'fit' => 'cover',
                'width' => config('services.image_resize.sizes.' . $size),
                'height' => config('services.image_resize.sizes.' . $size),
            ],
        ],
    ]);

    return config('services.image_resize.endpoint') . '/' . base64_encode($request);
}
```

zetas:
@jameslkingsley Is there any possibility of you creating a write-up or short readme on how to implement the AWS conversion pipeline with some more details? My experience with lambda is pretty much limited to vapor so it's a bit confusing still. 

It seems very much like it's relevant to this PR as this libraries usefulness is extremely restricted in a Serverless environment without conversions and optimizations.

CharlyZA:
I've dropped this into my dev environment and it's working perfectly. Thank you. It was very easy and seamless to change over to.  I don't know what I would've done without this PR.

georgeboot:
We have been using this branch for a day now and performed ~500 conversions.
On Vapor and with image conversions.

freekmurze:
Thank you @jameslkingsley for your work on this 👍

lloricode:
Looks good, :100:

ziming:
I had to downgrade to 7.14.2, as zip files generated by spatie media library couldn't be expanded in the latest release.

greatwitenorth:
@jameslkingsley amazing work. Do you haven any insight on why image conversions aren't working with Vapor? I'm only seeing it partially working and it seems very random (sometimes creates responsive images, sometimes not). I cannot see anything in the error logs (although I'm still a bit new to figuring out how all the logging works on aws).

jameslkingsley:
I think it has trouble because of the max 6 MB request payload - seems to be ok when processing a very small image. I recommend using something like [this](https://aws.amazon.com/solutions/serverless-image-handler) if all you need is simple resizing done - for anything more you might need to look into a more feature-complete service.

CharlyZA:
I was going to implement AWS's serverless image handler (The [one](https://aws.amazon.com/solutions/serverless-image-handler/) which @jameslkingsley mentioned) but I decided to give it a shot with letting Vapor and medialibrary handle it.

It's been working perfectly. I did have to increase the default cli-timeout as it would prematurely kill conversions for unusually large images.

Just a bit of context for my setup. My Laravel/Vapor setup is running as a purely JSON:API server with media sent as signed temporary URLs to S3. **Does anybody have any solutions they have experience with for converting document files (at least .docx and .xlsx) to .pdf on Vapor?** I'm just at a bit of a loss of how I should tackle this issue. I'm convinced that using an API, [Api2Pdf ](https://www.api2pdf.com/pricing), might be the best option. They're running on Vapor with minimal markups on computing time (3GB of memory) and bandwidth. 1000 documents for $2.

This isn't something which is really an issue as I don't see myself needing to implement this feature but I'm just curious as to why I'm getting the following error when trying to convert PDF files to images on Vapor. Is this just something which isn't possible because of the limitations of the Vapor environment?

```
ImagickException
1:17:21 AM
no decode delegate for this image format `/tmp/storage/medialibrary/temp/SoYEkByl7VlTNOlSNnmiqj459iwY1yXQ/lJtCk8OpG22UFLIV.pdf' @ error/constitute.c/ReadImage/544
/var/task/vendor/spatie/pdf-to-image/src/Pdf.php:44
```
Many thanks in advance.

PR #1594, opened 2019-10-14
created azure url generator class

myckhel:
created azure url generator class for azure storage disk users


freekmurze:
Thank you for this, but I'm not going to add this the package.

I'm not an Azure users myself, and I don't want to maintain code that I don't use myself.

PR #1604, opened 2019-10-23
MediaStream without rewind

Firtzberg:
Hi Spatie,
Thank you for your wonderful packages.

I did use ZipStream-PHP package recently and had difficulties with streaming data. Apparently the file contents get read twice because the crc is required upfront. This can be avoided by setting the zero header option in the archive. The ZIP format specification allows the crc to be added after the file contents, which is what happens.
I saw you use this package as well and wanted to give you a hint in order to make your package better.

This should not impact any behavior, so I'm not sure if I can write any test to demonstrate this working.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
     "illuminate/pipeline": "~5.5.0|~5.6.0|~5.7.0|~5.8.0|^6.0",
     "illuminate/support": "~5.5.0|~5.6.0|~5.7.0|~5.8.0|^6.0",
     "league/flysystem": "^1.0.8",
-    "maennchen/zipstream-php": "^0.4|^1.0",
+    "maennchen/zipstream-php": "^1.0",
     "spatie/image": "^1.4.0",
     "spatie/pdf-to-image": "^1.2",
     "spatie/temporary-directory": "^1.1",
--- a/src/MediaStream.php
+++ b/src/MediaStream.php
@@ -6,7 +6,6 @@ use ZipStream\ZipStream;
 use Illuminate\Support\Collection;
 use Spatie\MediaLibrary\Models\Media;
 use Illuminate\Contracts\Support\Responsable;
-use ZipStream\Option\Archive as ArchiveOptions;
 use Symfony\Component\HttpFoundation\StreamedResponse;
 
 class MediaStream implements Responsable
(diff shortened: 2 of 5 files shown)


Firtzberg:
It appears the archive options are not available until ZipStream-PHP version 1.0.0.
I am not sure if you're interested in upgrading the composer required from 0.4 to 1.0.

Firtzberg:
I might be wrong, but the Scrutinizer failed due to some storage issues.

rubenvanassche:
@freekmurze seems to work!

- cleaned up the code a bit
- removed zipstream 0.4 requirement

So what do you think, should we remove zipstream 0.4 in favor of 1.0?

freekmurze:
@rubenvanassche is this new version tested both locally and from S3? I know I did stick with 0.4 because streaming from S3 resulted in a corrupt zip.

Firtzberg:
@rubenvanassche Thank you for the changes. The code was in a similar state on https://github.com/spatie/laravel-medialibrary/commit/62011224ed91d86c68152b11f594f1ebddda8526
I am not sure if upgrading a dependency is a backward compatibility break. Your users could have the lower zipstream package in their own composer.json and use it themselves. They could also use a package depending on the lower zipstream package.
My last commit should be able to handle both. I don't know if we can configure composer to prefer 1.* and allow it to fall back to 0.* if the project or other packages require it.

freekmurze:
> I am not sure if upgrading a dependency is a backward compatibility break.
I'm thinking it is not. Composer will not pull in the new version of the medialibrary because the dependencies are not matched.

rubenvanassche:
> @rubenvanassche is this new version tested both locally and from S3? I know I did stick with 0.4 because streaming from S3 resulted in a corrupt zip.

Will try to check this today!

rubenvanassche:
> @rubenvanassche is this new version tested both locally and from S3? I know I did stick with 0.4 because streaming from S3 resulted in a corrupt zip.

I've added a test for streaming a zip from s3 files, this seems to work 🎉. The pull request looks fine to me so it can be merged! The only thing that still needs to happen is merging the fixes of StyleCI which for some reason I cannot do right know(emailed support).

@Firtzberg thanks for the PR!

rubenvanassche:
Hi @Firtzberg,

I've reverted this PR, because some zip extractors could not open the zip files with zero header (#1660). If you have a solution to the problem feel free to send in a new PR, thanks!

Firtzberg:
Thank you for the update.
This is probably an issue with the library.

PR #1615, opened 2019-11-01
version in read me. Update README.md

zohaibhassan156:
Need to update read me as its not installing on fresh copy of with that specified version


freekmurze:
Going to leave it as it is for now.

PR #1640, opened 2019-11-23
Type declaration removed

rafamont12:
Type declaration removed, because Media should be able to usage of other relationships, Polymorphic ManyToMany, for instance


freekmurze:
This isn't intended to be used with the morphs relationship.

Issue #1643, opened 2019-11-26
How to control maximum file size during runtime?


jameslkingsley:
I need to limit the upload file size based on the user's remaining storage space - right now I'm using this:

```php
class Medialibrary
{
    /**
     * Runs the given closure with the given size as the maximum file upload size.
     *
     * @return mixed
     */
    public static function maximumFileSize(int $bytes, callable $closure)
    {
        $default = config('medialibrary.max_file_size');

        config(['medialibrary.max_file_size' => $bytes]);

        $response = call_user_func($closure);

        config(['medialibrary.max_file_size' => $default]);

        return $response;
    }
}
```

and then I'm using it like this:

```php
Medialibrary::maximumFileSize($user->disk_usage->remaining, function () use ($request, $model) {
    return $model->addMediaFromDisk($request->key)
        ->toMediaCollection('attachments');
});
```

I'm wondering whether there is a better way to accomplish this, that is perhaps already built into the library? Maybe something like this:

```php
$model->addMediaFromDisk($request->key)
    ->maximumFileSize($user->disk_usage->remaining)
    ->toMediaCollection('attachments');
```

freekmurze:
You can use media collections to restrict file sizes.

https://docs.spatie.be/laravel-medialibrary/v7/working-with-media-collections/defining-media-collections/#only-allow-certain-files-in-a-collection

`Spatie\MediaLibrary\File` has a size property.

jameslkingsley:
This doesn't override the default `max_file_size` configuration option. I need the default max file size to remain at 10 MB, except on a particular model where it has an unlimited size (as uploading is being directly handled by S3).

jameslkingsley:
By using my original solution, will I run into issues with changing the config during runtime?

nmfzone:
+1 for controlling file size during runtime. Not really needed, but I think it's good to have it.

For now, we can set the max file size to big value, then let the Laravel Validation handle it per collection.

PR #1647, opened 2019-11-29
Add collection method maximumFileSize

jameslkingsley:
This PR adds the ability to specify a maximum file size per collection (#1643), which will override the default config value. The reason for this PR is that in my specific use-case I need a site-wide default max file size of 10 MB, but on one particular model I need the max to adjust to the user's allocated storage space.


freekmurze:
I'm going to pass on this for now as not many people need this functionality.

Implement this in your own project.

Issue #1576, opened 2019-09-25
Duplicate // in getResponsiveImagesDirectoryUrl()


ToxaGeek:
Duplicate // in https://github.com/spatie/laravel-medialibrary/blob/af0505b7e1f4c5c92b616aa30e5253f2c98fc647/src/UrlGenerator/LocalUrlGenerator.php#L99
`'/responsive-images/'` at `getPathForResponsiveImages()` and `.'/'` in the end of `getResponsiveImagesDirectoryUrl()`

PR #1653, opened 2019-12-03
Fix custom disk url giving invalid urls

arjanwestdorp:
When using a custom url for your disk the `UrlGenerator` will create a url with double `//`. This PR will prevent that from happening.

Previously, when the disk url was set to `https://localhost/media/` it would generate a url like `https://localhost/media//1/responsive-images//`

This solves #1576


freekmurze:
Thanks!

PR #1657, opened 2019-12-05
Document 'addMediaFromDisk' method

gdebrauwer:
The 'addMediaFromDisk' method (#1588) was not documented yet.


freekmurze:
Thanks!

PR #1666, opened 2019-12-11
Added diskName on copy and move methods in media model

Desperino:
Related to [this one](https://github.com/spatie/laravel-medialibrary/pull/1402#issuecomment-507608917).


freekmurze:
Thanks!

PR #1668, opened 2019-12-12
Correctly use the media item's disk when removing responsive images

jkudish:
Howdy @freekmurze and Spatie crew 👋 

I have a media collection setup that uses responsive images and also uses the non-default disk. I noticed that processing responsive images was always failing for this collection, and I tracked it down to the `removeResponsiveImages` method not specifying which disk to use. I noticed all the other methods call the filesystem's `disk` method, so I'm adding that here as well. It fixes the issue I encountered.

Thanks!


freekmurze:
Nice catch, thanks Joey!

PR #1709, opened 2020-01-25
Resolve duplication of migration file on republishing

### LINKED ISSUE #739: duplicated migration file
because of this line
https://github.com/spatie/laravel-medialibrary/blob/0491a18d95be3b8ef3e2e351cd14720d7484070a/src/MediaLibraryServiceProvider.php#L25

can we use only one data instead of re-generating new one on each publish ?

chrisbjr:
Resolves duplication of migration file when republishing the vendor "migrations" tag. Fixes #739 

Copied this method from laravel-permission:

https://github.com/spatie/laravel-permission/blob/master/src/PermissionServiceProvider.php#L23

### DIFF
--- a/src/MediaLibraryServiceProvider.php
+++ b/src/MediaLibraryServiceProvider.php
@@ -21,7 +21,7 @@ class MediaLibraryServiceProvider extends ServiceProvider
         ], 'config');
 
         $this->publishes([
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks for your work on this. I'm going to leave things as the are for now.

PR #1710, opened 2020-01-25
Add Laravel 7 support

driesvints:
I needed to work around a `json_decode` call in a model observer. Something seems to have changed here in the new major version but I'm not sure what's caused it 🤔


driesvints:
Not really sure why scrutinizer is failing. It's complaining about `composer.lock` not being present but that doesn't seems wanted.

driesvints:
Github actions also aren't running for some reason.

driesvints:
I think this PR is the one that affects this: https://github.com/laravel/framework/pull/30273

The current json call is actually a workaround imo so this seems to be fixed/improved for Laravel 7.

freekmurze:
Thanks!

freekmurze:
Refactored it a bit for readability.

PR #1730, opened 2020-02-11
Add support for loading attribute

freekmurze:


docs/installation-setup.md
```diff
@@ -73,7 +72,7 @@ return [
          * images. By default we optimize for filesize and create variations that each are 20%
          * smaller than the previous one. More info in the documentation.
          *
-         * https://docs.spatie.be/laravel-medialibrary/v8/advanced-usage/generating-responsive-images
+         * https://docs.spatie.be/laravel-medialibrary/v7/advanced-usage/generating-responsive-images
```
willemvb: Shouldn't this be v8?

freekmurze: Right!

config/medialibrary.php
```diff
@@ -63,6 +63,17 @@
...
+    /*
+     * When converting Media instances to response the medialibrary will add
+     * a `loading` attribute to the `img` tag. Here you can set the default
+     * value of that attribute.
+     *
+     * Possible values: 'auto', 'lazy' and 'eager,
+     *
+     * More info: https://css-tricks.com/native-lazy-loading/
+     */
+    "default_loading_attribute_value" => 'auto',
```
royduin: Single quotes maybe?

docs/installation-setup.md
```diff
@@ -90,6 +89,17 @@ return [
...
+    /*
+     * When converting Media instances to response the medialibrary will add
+     * a `loading` attribute to the `img` tag. Here you can set the default
+     * value of that attribute.
+     *
+     * Possible values: 'auto', 'lazy' and 'eager,
+     *
+     * More info: https://css-tricks.com/native-lazy-loading/
+     */
+    "default_loading_attribute_value" => 'auto',
```
royduin: Single quotes maybe?

config/medialibrary.php
freekmurze: Fixed on v8 👍

docs/installation-setup.md
freekmurze: Also fixed on v8

.editorconfig
```diff
@@ -9,7 +9,10 @@ indent_size = 4
 indent_style = space
 end_of_line = lf
 insert_final_newline = true
-trim_trailing_whitespace = true
+trim_trailing_whitespace = false
```
sebastiandedeyne: Why these changes?

sebastiandedeyne: Are the newlines at the end of the Blade file that bad? All Blade files have that anyway to keep Git happy.

freekmurze: Some of the blade files in this package render an `img` tag. I think it's best if they don't have that extra newline so the in the resulting html page there won't be extra empty lines added.

freekmurze:
This was at the request of @willemvb that we tied it to the conversion. Should anybody want a function to set it manually, I'd be open for a PR that adds that.

.editorconfig
sebastiandedeyne: In that case it would make more sense to add specific exceptions for the files where it matters in `.editorconfig` instead of having a global configuration that can introduce issues along the road.

freekmurze: I think it's fine for now. This behaviour is wanted for all currently present blade files, and I don't see us adding any new ones soon.

willemvb:
@freekmurze hmm must be some kind of misunderstanding, since my initial suggestion (in Slack) was something like `$yourModel->getFirstMedia()->loading('eager')`

PR #1732, opened 2020-02-12
Scope trim_trailing_whitespace setting change to blade files

sebastiandedeyne:
Is `trim_trailing_whitespace = false` even necessary for Blade files?


freekmurze:
Thanks!

PR #1736, opened 2020-02-18
V8 file value object

brendt:
@freekmurze this is a proof of concept of how I'd expand the current `File`'s functionality, and how to use it in `DefaultConversionFileNamer`. 

The use of `Media::getFile()` simplifies the code in `DefaultConversionFileNamer`.

I could do some more refactoring in other places, but wanted to hear your thoughts on this first. Maybe there's not enough to gain from it, what do you think?


freekmurze:
As discussed on Slack, let's leave it as it is for now.

PR #1738, opened 2020-02-19
Allow image_driver config to be set via .env

aryehraber:
This small change will allow the `image_driver` config option to be set via an `IMAGE_DRIVER` env variable. This is useful in cases where `gd` is available locally but `imagick` is used in production.

This change has no breaking-changes as the default is still `gd`.


freekmurze:
Thanks!

PR #1741, opened 2020-02-21
Include deletePreservingMedia() method

mattbartel:
The existing deletePreservingMedia() method on the HasMedia trait was missing from the docs.


freekmurze:
Thanks!

PR #1742, opened 2020-02-21
Add config to maintain files on model deletion

mattbartel:
I've added a config value `remove_files_on_delete` that, when set to false, will prevent the MediaObserver deleted observer from removing media files from the disk. (Default = true, which maintains current functionality)

My use case: 
I have built a content management system which uses the media library to upload image files associated with posts. Sometimes the image URLs are hard-coded into the content of the site, but admin users still wish to disassociate the media from the post to which they originally uploaded it. This setting allows that disassociation while maintaining the existing files.

### DIFF
--- a/config/medialibrary.php
+++ b/config/medialibrary.php
@@ -17,7 +17,7 @@ return [
     /*
      * If set to false, files will remain on the disk when Media model is deleted. 
      */
-    'remove_file_on_delete' => true,
+    'remove_files_on_delete' => true,
 
     /*
      * This queue will be used to generate derived and responsive images.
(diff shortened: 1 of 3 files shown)


freekmurze:
Some tests are failing for this one.

I'm not entirely sure if I want this In. Some of the clean up commands might have to be updated too. They now will delete any files that aren't coupled to a model. I guess those commands should be left alone when this new setting is `false`.

mattbartel:
Ah, thanks, I missed the cleanup functions. 

Short of decoupling the models from the files as I tried to do here, I'd really like to see optional soft deletes (which would maintain the file). Would you accept a pull request for that type of functionality?

freekmurze:
>  Would you accept a pull request for that type of functionality?

Maybe that could be added as `deletePreservingMedia()`. Closing this one for now.

mattbartel:
@freekmurze Thanks, I'll work on that and submit another PR.

PR #1750, opened 2020-02-28
Update using-custom-properties.md

designvoid:
Adding additional infor regarding filtering media collections by custom properties


freekmurze:
Thanks!

Issue #1151, opened 2018-06-16
The format() method has no effect on responsive images.


LichtAnd:
The format() method works fine for conversions but not for the generated responsive images.

On my model (App\User)I have the following methods:

```php
public function registerMediaCollections()
    {
        $this
            ->addMediaCollection('avatars')
            ->singleFile();
    }

public function registerMediaConversions(Media $media = null)
    {
        $this
            ->addMediaConversion('thumb')
            ->format(Manipulations::FORMAT_JPG)
            ->fit(Manipulations::FIT_CROP, 160, 160)
            ->withResponsiveImages();

public function addAvatar( $avatar)
    {
        return $this
            ->addMedia($avatar)
            ->toMediaCollection('avatars', 'media');
    }
```
When I upload an avatar.png I get:
.../public/media/1/conversions/avatar-thumb.jpg
.../public/media/1/responsive-images/avatar___thumb_160_160.png

Is this a bug or was it never intended to work like that?

freekmurze:
Could you submit a failing test for us to look at?

Issue #994, opened 2018-03-25
V7 How to customize the responsive images file type?


zhi8617:
Code like this ...  
```
 public function registerMediaConversions(Media $media = null)
    {
        $this->addMediaConversion('images')
            ->format(Manipulations::FORMAT_JPG)
            ->withResponsiveImages();   
    }
```
“Conversions” files Has been successfully converted ,but Responsive images did not. 
I look at documents and issues, I still no idea.

Thank all for making this.

brendt:
Hi @zhi8617 

Thanks for the question! At Spatie don't really have use cases for this. If you'd really want to save an image and its responsive variations in another format via the medialibrary, you could manually format the image before saving it as a media item.

If you're really interested in adding this feature, we might accept a PR if it could be done in a non-breaking way, so feel free to take the initiative.

At this point though, we're not going to invest time in this feature.

PR #1752, opened 2020-02-29
Allow to format change responsive images in conversions

buttjer:
#994 
#1027
#1151

This can be very useful for uploading files as png and converting it dynamically to jpg with different background colors for example.


freekmurze:
Thanks!

PR #1755, opened 2020-03-03
Update S3 url generator to use media disk

gdebrauwer:
The S3 Url generator currently assumes the disk that uses the "s3" driver also is called "s3". This is not always the case. In my case: I use a Digital Ocean Space, which also uses the "s3" driver, but the disk is not called "s3" of course. Currently it's not possible to generate a correct url using the standard s3 url generator.

### DIFF
--- a/src/UrlGenerator/S3UrlGenerator.php
+++ b/src/UrlGenerator/S3UrlGenerator.php
@@ -35,7 +35,7 @@ class S3UrlGenerator extends BaseUrlGenerator
 
         $url = $this->versionUrl($url);
 
-        return config('medialibrary.s3.domain').'/'.$url;
+        return config('medialibrary.'.$this->media->disk.'.domain').'/'.$url;
     }
 
     /**
@@ -76,6 +76,6 @@ class S3UrlGenerator extends BaseUrlGenerator
             $url = $root.'/'.$url;
         }
 
-        return config('medialibrary.s3.domain').'/'.$url;
+        return config('medialibrary.'.$this->media->disk.'.domain').'/'.$url;
     }
 }


freekmurze:
Thanks!

ragingdave:
Wait is this even a true issue? The code you edited has nothing to do (from what I can tell) with the config/filesystems and everything to do with the config/medialibrary.....so that would mean that each and every disk now would need a configuration in the medialibrary config? So that means that I now would have to add the like 10-15 disk configurations to config/medialibrary....this smells like a huge breaking change....unless I'm missing something

gdebrauwer:
You can’t really have multiple disks with the s3-driver in the old code, because you could only specify one base url that would be the same for all those s3-disks

ragingdave:
@gdebrauwer While it's true if all the disks are different, but if you have 1 actual disk, you can have multiple disks that each specify the root folder for that disk, then your update literally breaks the world for that config. Either way it seems like a half fix, and a breaking change rather than something that is just a minor version bump. As a setup that worked without issue before this change, is now utterly decimated and requires a whole new set of configs and testing which IMHO should be more than a minor fix.

ragingdave:
@freekmurze @gdebrauwer Beyond just from it being a breaking change, it would seem this isn't even a great solution as now each disk now requires its own key and array of config options in the so it will end up looking like:

```
'disk1' => [
    'domain' => DOMAIN1,
],
'disk2' => [
    'domain' => DOMAIN1,
],
```

Which is seemingly highly inefficient.

It would seem that perhaps just removing this config from the medialibrary config files in the first place would be better and use the already configured options in the disk instead or just having an optional config param added to the disks instead...

This just feels like avery short term fix that only addresses the issue partially and only solves one aspect of the issue while also breaking perfectly good and working apps all in one fell swoop.

freekmurze:
I agree that the PR could indeed be breaking and I've reverted it.

@gdebrauwer for your particular case, you could create your own url generator: https://docs.spatie.be/laravel-medialibrary/v7/advanced-usage/generating-custom-urls/

PR #1758, opened 2020-03-07
Use native file copy

mattdfloyd:
If a file and the media destination are on the same disk, use native copy operation instead of streaming.


freekmurze:
Thanks!

PR #1764, opened 2020-03-11
Set conversion disk when adding media from remote

georgeboot:
This seems to be a bug in v8.

When adding a file from s3, eg:
```php
$newImage = $event
    ->addMediaFromDisk('my/filename.jpg', 'uploads')
    ->toMediaCollection('main_image');
```

the `conversions_disk` is not being set, and thus the jobs that execute the conversions, fail.


freekmurze:
Thanks!

PR #1766, opened 2020-03-11
Allow "setName()" accept mixed values

orkhanahmadov:
In some situations, it is nice to have the "name" field accept something other than "string".
Example:
When used together with `spatie/laravel-translatable` we can set `name` DB field to JSON but, because `setName()` method does not accept array, it is not possible to set multiple translations when adding media like this:
``` php
$model->addMediaFromUrl('https://example.com/image.jpg')
        ->setName([
              'en' => 'Name in English',
              'de' => 'Name auf Deutsch'
        ])
        ->toMediaCollection('images');
```

### DIFF
--- a/src/MediaCollections/FileAdder.php
+++ b/src/MediaCollections/FileAdder.php
@@ -44,7 +44,8 @@ class FileAdder
 
     protected string $fileName = '';
 
-    protected string $mediaName = '';
+    /** @var string|mixed */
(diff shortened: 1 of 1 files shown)


freekmurze:
Could you add a test to make sure that passing an array works?

freekmurze:
Closing due to inactivity.

PR #1772, opened 2020-03-13
Fixed namespace in upgrade docs

tomcoonen:
There was a wrong namespace in the upgrade docs:
`Spatie\MediaLibrary\HasMedia\InteractsWithMedia` should be `Spatie\MediaLibrary\InteractsWithMedia` instead.


freekmurze:
Nice catch, thanks!

PR #1778, opened 2020-03-16
Backward compatibility for media created before conversions_disk was added

alibarthi:
Hi,

An exception is thrown for media (with conversions) created before the conversions_disk column was added to the media table. The conversions_disk is null for older media.

> Return value of Spatie\MediaLibrary\Support\UrlGenerator\BaseUrlGenerator::getDiskName() must be of the type string, null returned (View: /home/vagrant/my-project/resources/views/vendor/media-library/responsiveImageWithPlaceholder.blade.php) (View: /home/vagrant/my-project/resources/views/vendor/media-library/responsiveImageWithPlaceholder.blade.php) (View: /home/vagrant/my-project/resources/views/vendor/media-library/responsiveImageWithPlaceholder.blade.php)

Unless I'm missing something, the old media conversions disk is always the same as the media disk. This change solves the backward compatibility for me, I hope it helps.

Have a great day!

### DIFF
--- a/src/Support/UrlGenerator/BaseUrlGenerator.php
+++ b/src/Support/UrlGenerator/BaseUrlGenerator.php
@@ -57,7 +57,11 @@ abstract class BaseUrlGenerator implements UrlGenerator
 
     protected function getDiskName(): string
     {
-        return $this->conversion === null
+        if ($this->conversion === null) {
+            return $this->media->disk;
+        }
+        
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks for your work on this, but I'm not going to merge it in.

I'm thinking that we don't need this code if you would fill the `conversions_disk` name with the value of `disk` for each row, like suggested in the upgrade guide.

Issue #1779, opened 2020-03-16
InteractsWithMedia::getFirstMedia enforce $filters as array


vdauchy:
Hi,
I suppose this is a small bug as in InteractsWithMedia::class the function:

```
public function getFirstMedia(string $collectionName = 'default', array $filters = []): ?Media
    {
        $media = $this->getMedia($collectionName, $filters);

        return $media->first();
    }
```

enforce `$filters` type to `array` while `InteractsWithMedia::getMedia()` supports `$filters` as callable too.

Have a great day.

freekmurze:
I'd accept a PR with tests that fixes this.

freekmurze:
Fixed by #1781.

PR #1781, opened 2020-03-18
Add callable filter support to getFirstMedia()

50bhan:
This PR will add support of using callable filters to getFirstMedia() method. #1779


freekmurze:
Thanks!

PR #1794, opened 2020-03-24
Use float for extractVideoFrameAtSecond

francoism90:
I use this a lot to generate thumbnails of video's, however a rounded second isn't always the correct frame:
```
snapshot: 8
snapshot: 8.513062
```

### DIFF
--- a/src/Conversions/Conversion.php
+++ b/src/Conversions/Conversion.php
@@ -13,7 +13,7 @@ class Conversion
 
     protected ConversionFileNamer $conversionFileNamer;
 
-    protected int $extractVideoFrameAtSecond = 0;
+    protected float $extractVideoFrameAtSecond = 0;
 
     protected Manipulations $manipulations;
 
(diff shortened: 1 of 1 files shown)


francoism90:
Also note `TimeCode` of the ffmpeg package expects a float:
```
/**
     * Creates timecode from number of seconds.
     *
     * @param float $quantity
     *
     * @return TimeCode
     */
    public static function fromSeconds($quantity)
    {
        $minutes = $hours = $frames = 0;

        $frames = round(100 * ($quantity - floor($quantity)));
        $seconds = floor($quantity);

        if ($seconds > 59) {
            $minutes = floor($seconds / 60);
            $seconds = $seconds % 60;
        }
        if ($minutes > 59) {
            $hours = floor($minutes / 60);
            $minutes = $minutes % 60;
        }

        return new static($hours, $minutes, $seconds, $frames);
    }
```

freekmurze:
Thanks!

PR #1795, opened 2020-03-24
Update composer.json

francoism90:


freekmurze:
Thanks!

francoism90:
@freekmurze Thanks for the merge! Could you also please take a look at #1794?

PR #1798, opened 2020-03-25
Update readme to include actual config file

Christophvh:
Update readme to include the contents of the current config file.
It might be an option to add a link to the current config file?


freekmurze:
I've decided the move the config section from the readme.

PR #1803, opened 2020-03-28
Solve error when using flysystem-cached-adapter

### LINKED ISSUE #1800: [8.x] Media conversions breaks when using Cached Adapter
PHP version: 7.4.4
Laravel version: 7.3
Media Library version: 8.0.6

Following [Laravel docs](https://laravel.com/docs/7.x/filesystem#driver-prerequisites), user is advised to use `league/flysystem-cached-adapter` package when using remote storage (like AWS S3). The easiest way to use it is to modify AppServiceProvider boot() method and do something like this (memory cache):
```
use League\Flysystem\Filesystem;
use Illuminate\Support\Facades\Storage;
use League\Flysystem\Cached\CachedAdapter;
use League\Flysystem\Cached\Storage\Memory;

Storage::extend('s3', function ($app, $config) {
	return new Filesystem(new CachedAdapter($app['filesystem']->createS3Driver($config)->getDriver()->getAdapter(), new Memory));
});
```

Doing so breaks file upload with following error:
https://imgur.com/a/OGZZY9L

But the strange thing is that the file actually gets uploaded. The library throws that error when trying to run specified media conversions for that model. It's trying to get the path prefix from the cached adapter.

My media conversions are (just one):
```
{
		->width(828)
		->height(828)
		->keepOriginalImageFormat()
		->nonQueued();
}
```

EDIT: This was not the case with v7 of the library if that helps.

alibarthi:
Hi,

This change solves the error triggered on conversions when using an s3 driver disk:

`Call to undefined method League\Flysystem\Cached\CachedAdapter::getPathPrefix() in /home/vagrant/laravel-app/vendor/spatie/laravel-medialibrary/src/Support/UrlGenerator/DefaultUrlGenerator.php:31`

Fixes #1800

### DIFF
--- a/src/Support/UrlGenerator/DefaultUrlGenerator.php
+++ b/src/Support/UrlGenerator/DefaultUrlGenerator.php
@@ -28,7 +28,13 @@ class DefaultUrlGenerator extends BaseUrlGenerator
 
     public function getPath(): string
     {
-        $pathPrefix = $this->getDisk()->getAdapter()->getPathPrefix();
+        $pathPrefix = '';
+        
+        $adapter = $this->getDisk()->getAdapter();
+        
+        if (method_exists($adapter, 'getPathPrefix')) {
+            $pathPrefix = $adapter->getPathPrefix();
+        }
 
         return $pathPrefix.$this->getPathRelativeToRoot();
     }
--- a/src/Support/UrlGenerator/DefaultUrlGenerator.php
+++ b/src/Support/UrlGenerator/DefaultUrlGenerator.php
@@ -28,13 +28,13 @@ class DefaultUrlGenerator extends BaseUrlGenerator
 
     public function getPath(): string
     {
-        $pathPrefix = '';
-        
         $adapter = $this->getDisk()->getAdapter();
-        
-        if (method_exists($adapter, 'getPathPrefix')) {
-            $pathPrefix = $adapter->getPathPrefix();
+
+        if (! \method_exists($adapter, 'getPathPrefix')) {
+            $adapter = $adapter->getAdapter();
         }
+
+        $pathPrefix = $adapter->getPathPrefix();
 
         return $pathPrefix.$this->getPathRelativeToRoot();
     }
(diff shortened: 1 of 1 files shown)


freekmurze:
Won’t this return the wrong path when using the cached adapter?

alibarthi:
I think i`ve figured it out. Basically, the probelm is with both local and s3 disks. The explanation would be:

When the cache adapter is not installed, then 

`$this->getDisk()->getAdapter();`

returns an instance of League\Flysystem\Adapter\Local (or S3)

If the cached adapter is installed, then an instance of League\Flysystem\Cached\CachedAdapter is returned, which contains an instance of League\Flysystem\Adapter\Local. So we have to use:

`$this->getDisk()->getAdapter()->getAdapter();` 

(getAdapter twice, to get the adapter instance from the cached adapter instance).

Only after that we can use the getPathPrefix(), which is available in the Local (or S3) adapters but not available in the CachedAdapter instance. So the function will lokk like this:

`public function getPath(): string
    {
        $adapter = $this->getDisk()->getAdapter();

        if (! \method_exists($adapter, 'getPathPrefix')) {
            $adapter = $adapter->getAdapter();
        }

        $pathPrefix = $this->getDisk()->getAdapter()->getPathPrefix();

        return $pathPrefix.$this->getPathRelativeToRoot();
    }`

We can check if method_exists, to avoid using instanceof and make flysystem-cached-adapter a dependency.

I`ll can create a new pull request with these changes if you want.

src/Support/UrlGenerator/DefaultUrlGenerator.php
```diff
@@ -28,7 +28,13 @@ public function getBaseMediaDirectoryUrl()
 
     public function getPath(): string
     {
-        $pathPrefix = $this->getDisk()->getAdapter()->getPathPrefix();
+        $adapter = $this->getDisk()->getAdapter();
+
+        if (! \method_exists($adapter, 'getPathPrefix')) {
```
freekmurze: I think if we'd check on instanceof `CachedAdapter` would be more readable.

Als, drop the `\` before `method_exists`, it's a micro optimisation that we don't do anywhere in our code.

alibarthi:
You`re welcome and thank you for the great package :)

rcerljenko:
@alibarthi cheers!!  :)

freekmurze:
Thanks!

PR #1812, opened 2020-03-31
Allow ZipStream 2.0

mbabker:
The ZipStream library has a [2.0 release](https://github.com/maennchen/ZipStream-PHP/releases/tag/2.0.0) out which has a breaking change.  From my own use of the library in other applications, I haven't had noted any issues with the change, so I think this should be safe to apply here.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
         "illuminate/pipeline": "^6.18|^7.0",
         "illuminate/support": "^6.18|^7.0",
         "league/flysystem": "^1.0.64",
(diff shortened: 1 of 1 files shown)


freekmurze:
Before I merge this in, it should at least be tested if MediaStream still works. Could you test that out?

mbabker:
Gave it a quick spin with a model that had a collection with 3 ZIP files, worked like a charm.

```php
return MediaStream::create('packages.zip')->addMedia($model->getMedia('downloads'));
```

freekmurze:
Thanks!

PR #1818, opened 2020-04-06
Close #1712 - Save Media Files in ZIP Subfolders

### LINKED ISSUE #1712: Feature Request: Folders in Media Stream generated zip file
It will be great if for every media collection we can specify the folder name then all files in that media collection will be in that folder inside the generated zip file.

Would be really helpful when we use MediaStream to create a zip file consisting of files from multiple media collections attached to the same record.

jomo77:
Completion of the feature for custom property "zip_file_prefix" to store a media file in a specific ZIP Folder.

Just add a custom property named 'zip_filename_prefix' to a media model and the media will be saved to the specific Subfolder in the ZIP File.

### DIFF
--- a/src/Support/MediaStream.php
+++ b/src/Support/MediaStream.php
@@ -86,7 +86,7 @@ class MediaStream implements Responsable
     protected function getZipStreamContents(): Collection
     {
         return $this->mediaItems->map(fn (Media $media, $mediaItemIndex) => [
-            'fileNameInZip' => $this->getFileNameWithSuffix($this->mediaItems, $mediaItemIndex),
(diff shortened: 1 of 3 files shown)


freekmurze:
Nice, could you also send a PR to our docs with an example of this functionality. I'm thinking we could add a section "Special custom properties" on this page: https://docs.spatie.be/laravel-medialibrary/v8/advanced-usage/using-custom-properties/ 

(you can create a PR by clicking that "edit" button in the top right of the page)

Thanks!

Nyavowoyi:
In case anyone is still figuring out how to do this, I have answered a similar question in a Stackoverflow post.
Here is the link: https://stackoverflow.com/a/76824012/7880201

PR #1827, opened 2020-04-14
Update docs to reflect v8 removal of down()

jobyh:
Removal of the `down()` migration method in v8 breaks migration when using core command `php artisan migrate:refresh` or `php artisan migrate:rollback` then re-applying migrations (error as the table already exists).

I understand the removal is intentional as Spatie don't use these (see @freekmurze 's [comment](https://github.com/spatie/laravel-medialibrary/commit/1f4dabfb84a2f4cb795ed075e7a75a2daebfa549#r37984857) ) but I believe it would be useful to warn users in the documentation as this caught me out recently.


freekmurze:
Thanks!

PR #1829, opened 2020-04-14
Add page number support for Pdf image generator

albertarni:
We used this package for a project, where we needed to have to generate thumbnails also for the second page of a pdf. 
We saw that the video thumbnail generation is allowing us to set a specified timeframe for the thumbnail.
Based on this we extended the Pdf image generator to accept also the page number from where the thumbnail is generated.

The contribution was made with my colleague:
https://github.com/precubadrian


freekmurze:
Pretty neat! Thanks!

PR #1839, opened 2020-04-19
Allow model a chance to provide a media base path

inmanturbo:
Small change to allow developer to add mediaBasePath() method to a  [prepared model](https://docs.spatie.be/laravel-medialibrary/v8/basic-usage/preparing-your-model/) without having to implement the Spatie\MediaLibrary\Support\PathGenerator\PathGenerator  interface, or even publish the config file!

```php
    {
        // First we will check if our model  has its own
        // method for retrieving its media. Because
        // the developer's time is precious, and
        // this may save him from having to
        // implement the interface!
        if (method_exists($media->model, 'getMediaBasePath')) {
            return $media->model->getMediaBasePath($media);
        }

        return $media->getKey();
    }
```


freekmurze:
Thanks, but I'm going to pass on this for now.

PR #1842, opened 2020-04-22
Updated tags of the publshed assets for more package specific naming conventions

msonowal:
To have more granular control for end user to publish only migrations or config etc from this package.


freekmurze:
Could you also change this in the docs?

msonowal:
@freekmurze 
sure

PR #1853, opened 2020-04-28
Fixed duplicated path prefix for temporary urls

alibarthi:
Hi,

The path prefix (root) gets duplicated for temporary urls when disk root is not empty. This is not a problem with local adapter, or even s3 when the root is empty in the s3 disk config. 

First, when using $media->getTemporaryUrl(...), in `spatie/laravel-medialibrary/src/Support/UrlGenerator/DefaultUrlGenerator.php` we have the prefix added in the `getPath()` function:

```
    {
        ...
        return $pathPrefix.$this->getPathRelativeToRoot();
    }
```

Then `laravel/framework/src/Illuminate/Filesystem/FilesystemAdapter.php` also adds it in:

```
public function getAwsTemporaryUrl($adapter, $path, $expiration, $options)
    {
        ...
        $command = $client->getCommand('GetObject', array_merge([
            'Bucket' => $adapter->getBucket(),
            'Key' => $adapter->getPathPrefix().$path,
        ], $options));
        ...
    }
```

So the getTemporaryUrl() function should use ->getPathRelativeToRoot(), to make sure we get the path without the root prefix in disk config.

```
    {
        return $this->getDisk()->temporaryUrl($this->getPathRelativeToRoot(), $expiration, $options);
    }
```

### DIFF
--- a/src/Support/UrlGenerator/DefaultUrlGenerator.php
+++ b/src/Support/UrlGenerator/DefaultUrlGenerator.php
@@ -18,7 +18,7 @@ class DefaultUrlGenerator extends BaseUrlGenerator
 
     public function getTemporaryUrl(DateTimeInterface $expiration, array $options = []): string
     {
-        return $this->getDisk()->temporaryUrl($this->getPath(), $expiration, $options);
+        return $this->getDisk()->temporaryUrl($this->getPathRelativeToRoot(), $expiration, $options);
     }
 
     public function getBaseMediaDirectoryUrl()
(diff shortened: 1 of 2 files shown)


freekmurze:
Could you add a test for this, so we are sure we don't break this in the future?

alibarthi:
I hope it's ok, basically the test is just like `it_retrieves_a_temporary_media_url_from_s3()` but sets a root in s3 config and checks for a duplicated root name in url.

freekmurze:
Thanks!

alibarthi:
You're welcome

PR #1854, opened 2020-04-28
Fixed duplicated / in paths

alibarthi:
Hi,

The / gets duplicated in paths for both local and s3 adapters. If we `dd($this->getDisk()->url('/'));` we'll get something like `"/storage//"`.


freekmurze:
Thanks!

PR #1857, opened 2020-05-01
Allow the modal to be over ridden

lionslair:
This file assumes the file has not been over ridden / extended. This allows the package to work when the Media model is over ridden.


freekmurze:
Thanks!

lionslair:
> Thanks!

Did not expect that quick a response. thank you

PR #1858, opened 2020-05-02
Remove committed .php_cs.cache

driesvints:
Think this was accidentally committed?


freekmurze:
I was tired of the wave of PR that tried to add this, so I caved 😀

freekmurze:
Thanks!

PR #1860, opened 2020-05-03
Now using Laravel MediaLibrary's temporary directory during image con…

sander3:
…versions.

Hi there,

It seems like the underlying `GlideConversion` is relying on `sys_get_temp_dir()` for its temporary directory instead of `config('media-library.temporary_directory_path')`.

### DIFF
--- a/src/Support/ImageFactory.php
+++ b/src/Support/ImageFactory.php
@@ -8,6 +8,8 @@ class ImageFactory
 {
     public static function load(string $path): Image
     {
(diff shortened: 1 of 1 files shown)


freekmurze:
This PR breaks some of the tests: https://github.com/spatie/laravel-medialibrary/pull/1860/checks?check_run_id=640789105#step:9:24

Can you take a look at that?

sander3:
Seems to be something wrong with forwarding calls. I'll look into it.

Issue #1864, opened 2020-05-08
Media Eloquent model class name is pluralized to medias


synthor:
I recognized a bug triggered by a suddenly pluralized model name media to medias which causes the query builder to run into an exception.

`[2020-05-08 18:24:56] development.ERROR: SQLSTATE[42S02]: Base table or view not found: 1146 Table 'mee2.medias' doesn't exist`

We are using Laravel 5.5.49 with version 6.9.2 of the medialibrary.

It was working today but stopped through a breaking change somewhere. I'll look further for the cause now.

Adding

`protected $table = 'media';`

in the model class fixes the breaking database query of course.

matro-skin:
The same bug
But I have the same repo at 2 pc's. At one of them I have this error...

synthor:
Adding media as uncountable word in Laravel's Illuminate\Support\Pluralizer works aswell if I remove the $table override in the Media model class.

I don't understand yet where this breaking change came from.

matro-skin:
https://github.com/laravel/framework/issues/32723

olegmifle:
The same bug. 
For now u can fix it if use custom media model. In `medialibrary.php` change parameter `'media_model'` set your custom model.
For example
```
class CustomMedia extends \Spatie\MediaLibrary\Models\Media
{
    protected $table = 'media';
}
```

synthor:
That would be my workaround but we don't use a derivated media model yet.

mikemand:
I think this is related to the latest release of Inflector: https://github.com/doctrine/inflector/compare/1.3.1...1.4.0

It wasn't happening to me yesterday, but after updating packages today I have this error.

patrickbrouwers:
It's indeed related to Inflector, I've opened a PR to fix the breaking change: https://github.com/doctrine/inflector/pull/142

synthor:
Thanks for the PRs @patrickbrouwers

freekmurze:
Meanwhile I've also tagged `8.2.5` of the medialibrary which explicitly set the table name to be used to `media`.

synthor:
Can you "backport" this fix into 6.9 aswell please @freekmurze

freekmurze:
I'll do that after the weekend, if the PRs to inflictor are not merged by then.

synthor:
Alright. I have a workaround by using a derived model class over the weekend and for the future.

But other people on 6.9 will run into this without an own model class.

Have a nice weekend!

[...]

PR #1865, opened 2020-05-08
V7: set the media table for now due to medias bug

bomshteyn:
set the media table name explicitly to `media` to avoid failure caused by doctrine/inflictor now using `medias` as the plural

addresses: #1864


freekmurze:
Thanks! Meanwhile, I already took care of this.

freekmurze:
This will be fixed in doctrine/inflictor.

Issue #1868, opened 2020-05-09
Laravel Vapor errors since version 8.2.2 (I can create PR to resolve it)


stefro:
Hi,

I've updated the package and it immediately crashed (500) our app on Vapor. On Vapor, it crashes because of an AWS validator (see below in json error log). I've been reverting the package numbers in our app and noticed it happens from version 8.2.2.

This is the error log:
```json
"class": "ErrorException",
    "message": "Found 1 error while validating the input provided for the GetObject operation:\n[Key] expected string length to be >= 1, but found string length of 0 (View: /var/task/resources/views/livewire/main/car-tile.blade.php)",
    "code": 0,
    "file": "/var/task/vendor/aws/aws-sdk-php/src/Api/Validator.php:65",
    "previous": {
        "class": "InvalidArgumentException",
        "message": "Found 1 error while validating the input provided for the GetObject operation:\n[Key] expected string length to be >= 1, but found string length of 0",
        "code": 0,
        "file": "/var/task/vendor/aws/aws-sdk-php/src/Api/Validator.php:65"
    }
```

The view file `car-tile.blade.php` contains this:
```php
 {{ $car->getFirstMedia('car')->img('car') }}
```
When I remove this line from the code (so medialibrary is not used) the app works fine!

## The solution
Version 8.2.2 is only a small change in file `src/Support/UrlGenerator/DefaultUrlGenerator.php`:
```php

    // OLD
    public function getBaseMediaDirectoryUrl()
    {
        return $this->getDisk()->url('/');
    }

    // NEW
    public function getBaseMediaDirectoryUrl()
    {
        return $this->getDisk()->url('');
    }
```

I've manually changed the `/` back just to try it out on vapor, and now the app doesn't crash anymore. **So this fixes it!**

Would you accept a PR for reverting to `return $this->getDisk()->url('/');` since this resolved all the errors?

freekmurze:
Yup, send a PR for this! 👍

PR #1869, opened 2020-05-09
Fix: Revert slash because of Vapor errors (Issue #1868)

stefro:
Revert the slash in `return $this->getDisk()->url('/');` that was removed in version 8.2.2.

His is because of errors in Vapor as described in issue [#1868](https://github.com/spatie/laravel-medialibrary/issues/1868).


stefro:
I'm not sure why the tests in github actions on `prefer-lowes` fail. The errors don't seem to be related to the changed code.

freekmurze:
Thanks!

stefro:
Thanks for accepting the PR @freekmurze!

PR #1873, opened 2020-05-11
fixing table name issue

sgc-fireball:
medias vs media in media model


freekmurze:
This will be fixed in doctrine/inflictor soon.

PR #1876, opened 2020-05-12
Add default media class

tontonsb:
This is a solution attempt on problem discussed in #1875 

I think that this is not a breaking change as it is just a default and the same as the default was before. It's just that this is available even if the config is cached and doesn't get the new value.


freekmurze:
I'm going to leave it like this, as the problem only occurs when caching config locally, which you shouldn't do.

tontonsb:
Caching locally as opposed to what? Caching globally?

It seems that people are running into this quite a lot. https://github.com/spatie/laravel-medialibrary/issues?q=Class+name+must+be+a+valid+object+or+a+string

freekmurze:
https://laravel.com/docs/master/configuration#configuration-caching

> The command should not be run during local development as configuration options will frequently need to be changed during the course of your application's development.

Issue #1862, opened 2020-05-07
Generated responsive images filesize larger than original


ericp-mrel:
If I do something like this where I manually define the quality level of the image.

```php
// in your model
public function registerMediaConversions(Media $media = null): void
{
    $this->addMediaConversion('hero')
            ->optimize()
            ->width(3840)
            ->height(2160)
            ->quality(85)
            ->withResponsiveImages();
}
```

The normal conversion images have a reasonable filesize, but the responsive images will usually have a larger filesize since they are being generated at quality level 90 no matter what quality level is defined in the conversion.

It seems, that's mostly to do with this block of code where it doesn't take into account the possibly user defined quality level from the conversion. https://github.com/spatie/laravel-medialibrary/blob/459c3e6ab93e8b65f735b417882ba81fdba22865/src/ResponsiveImages/ResponsiveImageGenerator.php#L90-L93

PR #1881, opened 2020-05-14
Assure Responsive Image Generation Respects Conversion Quality Setting

MattStrauss:
I saw the help wanted tag on [this issue](https://github.com/spatie/laravel-medialibrary/issues/1862), and I wanted to contribute as all of the awesome Spatie packages are such a boon to the Laravel community. 

## Original Issue
The [original issue](https://github.com/spatie/laravel-medialibrary/issues/1862) was that the the responsive image generation process did not respect the quality setting of a conversion. 

## Proposed Solution
This pull request adds a private method to determine if the conversion has a quality setting, and if so, it applies that quality setting to the generated responsive images. 

## Tests
I also created a test for the change to assure it actually worked. I wanted to be more precise, i.e. show that at 60% quality the image would be 60% smaller, but while the numbers are close, they're not exact. So I simply asserted that the lower quality responsive image should have a smaller file size than the standard responsive image. 

## Concerns 
I added two additional conversions in `TestModelWithResponsiveImages`, and I'm hoping this doesn't slow the test suite down too much. I looked for alternative ways to test the changes but all of the current conversions within this class didn't quite unambiguously achieve what the test needed to prove, IMO. 

If any changes are needed or this just isn't the solution you are looking for, don't hesitate to let me know, I will take no offense. And thanks again for all of your amazing packages!

### DIFF
--- a/src/ResponsiveImages/ResponsiveImageGenerator.php
+++ b/src/ResponsiveImages/ResponsiveImageGenerator.php
@@ -47,7 +47,7 @@ class ResponsiveImageGenerator
         $media = $this->cleanResponsiveImages($media);
 
         foreach ($this->widthCalculator->calculateWidthsFromFile($baseImage) as $width) {
-            $this->generateResponsiveImage($media, $baseImage, 'media_library_original', $width, $temporaryDirectory);
+            $this->generateResponsiveImage($media, $baseImage, 'media_library_original', $this->getConversionQuality(), $width, $temporaryDirectory);
         }
 
         event(new ResponsiveImagesGenerated($media));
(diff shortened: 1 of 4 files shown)


src/ResponsiveImages/ResponsiveImageGenerator.php
```diff
@@ -64,18 +64,30 @@ public function generateResponsiveImagesForConversion(Media $media, Conversion $
...
+        }
+
+        return $conversion->getManipulations()->getManipulationArgument('quality') ?: $defaultConversionQuality;
+    }
+
     public function generateResponsiveImage(
         Media $media,
         string $baseImage,
         string $conversionName,
+        int $conversionQuality,
```
freekmurze, line 90: Addin a parameter here is a breaking change.

MattStrauss: Sure thing, I'll fix that shortly, likely just make it an optional parameter.

MattStrauss: See the [newest commit](https://github.com/spatie/laravel-medialibrary/pull/1881/commits/bd3b36a27380628d38ed21baa4e1f37ddb41ba35) that (hopefully) achieved the changes you requested.

freekmurze:
This looks better 👍 

Could you also update the docs with an example of the new functionality?

MattStrauss:
I updated the docs per @freekmurze request. Let me know if that makes sense, or if you prefer that I add a small **standalone** section in the docs for this. I just tacked on some details to the most pertinent existing section as this seems like a smallish change/feature.

freekmurze:
Thanks!

PR #1892, opened 2020-05-22
Remove docs from export

mpociot:
The docs currently add a whopping 2.8 megabytes to the vendor folder, which is not required :)


freekmurze:
Thanks, agent!

Issue #1896, opened 2020-05-24
Queued conversion delay on fallback URL 


samirmh-dev:
I want create PR for some small feature improvement on `InteractsWithMedia` trait.

On both `getFirstMediaUrl` and `getFirstMediaPath` method, trait checks,

if there is some record on DB return it's URL
if there is not any record found on DB return defined fallback URL.

So there is one point, if img conversion is queued and queue is not completed work yet. 
At this time `getFirstMediaUrl` will return correct IMG url, but there is not such img exists yet. Because queue is not saved it yet. 

Event queue is working normally there is **1sec delay** happenes when queue saves is as file. So user seeing just missing IMG with correct URL:

![image](https://user-images.githubusercontent.com/54883542/82751480-96f6d400-9dc8-11ea-8f2f-302cd4fa9572.png)

After 1second just refresing page and img will be there. Because queue saved it now.

![image](https://user-images.githubusercontent.com/54883542/82751506-be4da100-9dc8-11ea-87ca-e655544b588d.png)

We need to add additional `file_exists()` check for `getFirstMediaPath()` and `file_get_contents()` check for `getFirstMediaUrl()` . 

If DB record exists but that IMG URL is not exists yet, return fallback IMG URL.

Also this changes applied to related methods where uses `getFallbackMediaUrl` and `getFallbackMediaPath`.

**NOTE:**

I added `file_get_contents` with `@` to don't throw 404 error. It will just return false if note exists.

freekmurze:
We'll continue this in #1897

PR #1897, opened 2020-05-24
Patch for queue delay #1896

samirmh-dev:
> I want create PR for some small feature improvement on `InteractsWithMedia` trait.
> 
> On both `getFirstMediaUrl` and `getFirstMediaPath` method, trait checks,
> 
> if there is some record on DB return it's URL
> if there is not any record found on DB return defined fallback URL.
> 
> So there is one point, if img conversion is queued and queue is not completed work yet.
> At this time `getFirstMediaUrl` will return correct IMG url, but there is not such img exists yet. Because queue is not saved it yet.
> 
> Event queue is working normally there is **1sec delay** happenes when queue saves is as file. So user seeing just missing IMG with correct URL:
> 
> image
> 
> After 1second just refresing page and img will be there. Because queue saved it now.
> 
> image
> 
> We need to add additional `file_exists()` check for `getFirstMediaPath()` and `file_get_contents()` check for `getFirstMediaUrl()` .
> 
> If DB record exists but that IMG URL is not exists yet, return fallback IMG URL.
> 
> Also this changes applied to related methods where uses `getFallbackMediaUrl` and `getFallbackMediaPath`.
> 
> **NOTE:**
> 
> I added `file_get_contents` with `@` to don't throw 404 error. It will just return false if note exists.

Main issue is #1896

### DIFF
--- a/src/InteractsWithMedia.php
+++ b/src/InteractsWithMedia.php
@@ -245,11 +245,11 @@ trait InteractsWithMedia
     {
         $media = $this->getFirstMedia($collectionName);
 
-        if (! $media) {
+        if (! $media || ! file_get_contents($url = $media->getUrl($conversionName)) ) {
             return $this->getFallbackMediaUrl($collectionName) ?: '';
         }
 
-        return $media->getUrl($conversionName);
+        return $url;
     }
 
     /*
@@ -265,11 +265,11 @@ trait InteractsWithMedia
     ): string {
         $media = $this->getFirstMedia($collectionName);
 
-        if (! $media) {
+        if (! $media || ! file_get_contents($url = $media->getTemporaryUrl($expiration, $conversionName))) {
             return $this->getFallbackMediaUrl($collectionName) ?: '';
         }
 
-        return $media->getTemporaryUrl($expiration, $conversionName);
+        return $url;
     }
 
     public function getMediaCollection(string $collectionName = 'default'): ?MediaCollection
@@ -299,11 +299,11 @@ trait InteractsWithMedia
     {
         $media = $this->getFirstMedia($collectionName);
 
-        if (! $media) {
+        if (! $media || ! file_exists($url = $media->getPath($conversionName))) {
             return $this->getFallbackMediaPath($collectionName) ?: '';
         }
 
-        return $media->getPath($conversionName);
+        return $url;
     }
 
     /*
--- a/src/InteractsWithMedia.php
+++ b/src/InteractsWithMedia.php
@@ -245,7 +245,7 @@ trait InteractsWithMedia
     {
         $media = $this->getFirstMedia($collectionName);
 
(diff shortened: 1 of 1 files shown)


samirmh-dev:
> Could you remove `@` and add some tests?
> 
> Thanks!

sure! I will work on it asap

samirmh-dev:
@freekmurze 

I researched little bit about `@` for handling URL check and currently we have 4 choice:

### 1) keeping `@`.
 
If we use `file_get_contents()` without `@` it will throw Throwable if URL isn't exists. When using `@file_get_contents()` it just returns `FALSE` if any exception throws:

````php
if(!$media || !@file_get_contents($url = $media->getUrl($conversionName))) {
     return $this->getFallbackMediaUrl($collectionName) ? : '';
}
````

When using without `@`:

image

<hr>

### 2) using `try(){...}catch(){}`

[Referance from PHP.net](https://www.php.net/manual/ru/function.file-get-contents.php#120366):

image

<hr>

### 3) removing `@` and using `curl` check. 
We can add private method for checking.

````php
if(!$media || !$this->url_exists($url = $media->getUrl($conversionName))) {
     return $this->getFallbackMediaUrl($collectionName) ? : '';
}
````

````php
/**
 * Check given URL exists and returns trusted status code
 * 
 * @param string $url
 *
 * @return bool
 */
 public function url_exists(string $url) : bool {
    $curl = curl_init($url);
    curl_setopt($curl, CURLOPT_NOBODY, true);
    $result = curl_exec($curl);

    //if request did not fail
    if (!$result) return false;
    
    $statusCode = curl_getinfo($curl, CURLINFO_HTTP_CODE);

    if ($statusCode != 200) return false;
  
    curl_close($curl);
 
    return true;
 }
````

Personally I don't preferer `curl` method to `file_get_contents`

<hr>

### 4) using `get_headers()` method

As I mentioned at `curl` method above, we need to add private method for `get_headers` method too. 

````php
if(!$media || !$this->url_exists($url = $media->getUrl($conversionName))) {
     return $this->getFallbackMediaUrl($collectionName) ? : '';
}
````

````php
/**
 * Check given URL exists and returns trusted status code
 * 
 * @param string $url
 *
 * @return bool
 */
 public function url_exists(string $url) : bool {
     $headers = get_headers($url);
     return substr($headers[0], 9, 3) === 200;
 }
````

<hr>

@freekmurze Which method do you prefer? Personally I suggest to use just `@` without any additional method.

samirmh-dev:
actually I tried get_headers method for a few requests and it fails sometimes with `HTTP Request Failed!` error. Currently best option is using `@file_get_contests()`

samirmh-dev:
@freekmurze currently all tests passed. You can merge this when u avaiable

freekmurze:
Hi, I'm going to hold off on this for now as these methods only work for local filesystems and not remote ones.

PR #1903, opened 2020-05-27
SVG to PNG Conversion Loses Transparency

richardjkeys:
I found that PNG conversions were being created with white backgrounds.

Seems you must set the background color before reading the file, otherwise your transparent background is not respected.

I've simply changed the lines around in the Image Generator, and set it to use the result extension as the Imagick file type.


freekmurze:
Closing due to inactivity.

Issue #812, opened 2017-10-24
Unable to decode image from file error, introduced since 6.2.1


oad-dwarf:
This change:

https://github.com/spatie/laravel-medialibrary/compare/6.2.0...6.2.1

.. broke media conversions for me as I updated spatie/laravel-medialibrary from a version before 6.2.1. 

I get this error:

Intervention\Image\Exception\NotReadableException
Unable to decode image from file (/tmp/GlideBKMlDI).

When I look at the temp file, I see that it is consistently shorter than the original file.

I was able to fix it by changing be fread() to fgets(). Why it makes a difference I have no clue. It doesn't seem right to use fgets() for binary data :-)

The environment is alpine linux 3.6 and php 7.1 with nginx and fpm.

AlexVanderbist:
That's really strange... I tried to reproduce this issue but couldn't seem to make anything break. 

Does this happen with all images? Can you try this with a regular text file?

oad-dwarf:
It happened with a 616696 bytes png file.

freekmurze:
Is this still a problem in the latest version of the package?

MartinJac:
@freekmurze unsure about @oad-dwarf, but I've been able to replicate this when using s3.

Changing from fread() to fgets() resolved the issue (Windows).

freekmurze:
Closing because we can't reproduce it.

tobz-nz:
@oad-dwarf Did you figure this one out? I'm getting the same issue. Also Alpine-Linux. (in a K8 setup)

simioluwatomi:
I have been getting this error on all PNG images when uploading to s3. Very weird. Uploads are being performed with `addMediaFromBase64`

tobz-nz:
I think its an Alpine issue. We swapped to an Ubuntu image(no other changes) and all working perfectly now.

oad-dwarf:
@tobz-nz No I still have my patch in place. Haven't tested since then if the problem is still there. Perhaps I'll check it again when we switch to Alpine 3.8 and PHP 7.2

Aryan1:
Having same issue, the original image is uploaded successfully but conversions simply fails for all png files when using s3 driver, same error as mentioned above.

on Windows + PHP 7.2.8

ShinHyungJune:
Having same issue Too T.T
when using s3 driver, same error as mentioned above.

on Windows + PHP 7.1.2

skalero01:
[...]

PR #1909, opened 2020-06-05
Changed freads to fgets as per issue #812

mikemike:
This fixes the bug on Windows that prevents png resizing when using s3 driver


freekmurze:
Thanks!

Issue #1737, opened 2020-02-18
If using remote filesystem (FTP), removeAllFiles() throws exception


jruszo:
Hello!

I wrote a simple URL and Path generator, to store media files on FTP backend.
If I store the files like that:
```
images/1/62 - media directory
images/1/62/conversions - conversions directory
images/1/62/responsive-images - responsive images
```

Deleting an image throws exception, as Filesystem.php uses the following order:
https://github.com/spatie/laravel-medialibrary/blob/master/src/Filesystem/Filesystem.php#L144
```       
collect([$mediaDirectory, $conversionsDirectory, $responsiveImagesDirectory])
````
Due to this ordering, first it will delete the mediaDirectory, which recursively removes conversions and responsive-images as well, which results in exception: "local.ERROR: ftp_rmdir(): 4/4/62/conversions: No such file or directory"

I simply put $mediaDirectory  to the end of the list in collect() and it resolved my issue

It seems issues #1625 is caused by this similarly

leoflapper:
@jruszo having the same issue. The only possible solution to fix this issue for now is to edit the file directly?

freekmurze:
I accept a PR that improves this behaviour.

DenMette:
@freekmurze, see my pull request for this issue

freekmurze:
Fixed in latest version.

Issue #1625, opened 2019-11-10
moving files from one model to another in ftp filesystem raise error: ftp_rmdir(): Can't remove directory: No such file or directory


fsaravi:
in ftp filesystem when I want to move file from one model to another model this error raise:
ftp_rmdir(): Can't remove directory: No such file or directory

when I trace code found that when function removeAllFiles in class Spatie\MediaLibrary\Filesystem\Filesystem calls (line 104), it tries to remove root files folder and in subsequent it tries to remove **conversions** and **responsiveImages** folders:
```
public function removeAllFiles(Media $media)
    {
        $mediaDirectory = $this->getMediaDirectory($media);

        $conversionsDirectory = $this->getMediaDirectory($media, 'conversions');

        $responsiveImagesDirectory = $this->getMediaDirectory($media, 'responsiveImages');

        collect([$mediaDirectory, $conversionsDirectory, $responsiveImagesDirectory])

            ->each(function ($directory) use ($media) {
                $this->filesystem->disk($media->disk)->deleteDirectory($directory);
            });
    }
```
but deleteDirectory method will removes directory and all of its files and subdirectories. 
when $mediaDirectory removes and if **conversions** and **responsiveImages** are subdirectories of $mediaDirectory then these folders will remove automatically.

so any try for removing  $conversionsDirectory or $responsiveImagesDirectory in ftp filesystem will raise this error.

so if you force **conversions** and **responsiveImages** becomes subfolder of mediaDirectory then you could replace this method with this one:

```
public function removeAllFiles(Media $media)
    {
        $mediaDirectory = $this->getMediaDirectory($media);

        $this->filesystem->disk($media->disk)->deleteDirectory($mediaDirectory);
    }
```

hope it helps and fix this issue

PR #1912, opened 2020-06-06
fixing  ftp_rmdir(): Can't remove directory error in #1737 and   #1625

ibrahem-kamal:
fixing the error " ftp_rmdir(): Can't remove directory " when using ftp disk . i simply applied #1737 solution and added a try & catch for the deleting function that logs the error instead of throwing it Ï


freekmurze:
The tests are failing for this one, could you take a look at that?

freekmurze:
Fixed in latest version.

PR #1927, opened 2020-06-17
Add @ to ignore tempnam() notice

simoebenhida:
In PHP 7.1 `tempnam()` function started to throw a notice if falling back to temp dir!

https://bugs.php.net/bug.php?id=69489

And that prevents the flow from working properly!

### DIFF
--- a/src/InteractsWithMedia.php
+++ b/src/InteractsWithMedia.php
@@ -130,7 +130,7 @@ trait InteractsWithMedia
             throw UnreachableUrl::create($url);
         }
 
-        $temporaryFile = tempnam(sys_get_temp_dir(), 'media-library');
+        $temporaryFile = @tempnam(sys_get_temp_dir(), 'media-library');
         file_put_contents($temporaryFile, $stream);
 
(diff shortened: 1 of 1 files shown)


freekmurze:
This package in not meant to be used in PHP 7.1

https://github.com/spatie/laravel-medialibrary/blob/a3634f6645e07f4982ff64bd75e90c9e8adafba0/composer.json#L25

simoebenhida:
@freekmurze It is happening from PHP 7.1+! So it will happen also on PHP 7.4

freekmurze:
I don't see this notice when running the tests. Can this notice be recreated in a testing environment?

simoebenhida:
The notice is firing when the `tempnam` creates the folder with the process!
Hmm we can do ```ini_set('upload_tmp_dir', '/not-exists/path')``` to force tempname to create a folder but not sure if that will work or `ini_set` is a good approach inside a test

freekmurze:
Closing this for now, as we don't have an issue with the current behaviour and no other users are reporting problems around this.

PR #1928, opened 2020-06-18
Update the migration stub to include dropIfExists clause

nexxai:
There was no dropIfExists clause in the stub so I've added one here.

### DIFF
--- a/database/migrations/create_media_table.php.stub
+++ b/database/migrations/create_media_table.php.stub
@@ -28,4 +28,9 @@ class CreateMediaTable extends Migration
             $table->nullableTimestamps();
         });
     }
+    
+    public function down()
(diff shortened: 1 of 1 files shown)


freekmurze:
We don’t migrite down in any of our package or projects.

nexxai:
I’m totally ok with that, I’m just wondering if you can explain why (maybe I shouldn’t be migrating down either?)

freekmurze:
I think 99.9% of the time you are not gonna use a down migration, you’re probably never test them, so why even bother writing them. 

Just write an up migration if you want to drop a column again. 👍

nexxai:
Interesting. Thanks for that perspective, man!

PR #1935, opened 2020-06-19
Add ext-exif to prevent composer from stuck.

lomotech:
Composer sometimes will stuck without any useful message about the inactive extension when install using composer install or composer update. Only can be trace when using composer require method.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,6 +23,7 @@
     ],
     "require": {
         "php": "^7.4",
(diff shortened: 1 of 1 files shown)


freekmurze:
To be on the safe side, I'm not going to merge it in now, but I'll revisit this issue when creating a new major version.

freekmurze:
I've added this on the `v9` branch

PR #1937, opened 2020-06-22
Change to url decode

### LINKED ISSUE #1929: Responsive images when adding media from a url
Hi,

I'm importing an image from a url but I have an issue in that the image's file name has an accented character in [AntarèsThumb.jpg](https://orbit.brightbox.com/v1/acc-jqzwj/Marquis-Leisure/reviews/images/000/000/898/original/Antar%C3%A8sThumb.jpg)

Because of the accent fopen fails, I can use urlencode on the filename which will then import correctly but media library looks for the accented file name AntarèsThumb.jpg but because of the url encoding (I assume) it was actually saved as Antar%C3%A8sThumb.jpg which then means that the responsive images generation can't find the image because on downloading media library correctly identifies the filename from the image it downloaded.

I've resolved this by url encoding the filename for the url then explicitly setting usingFileName
```php
$filename = 'AntarèsThumb.jpg';
$url = $baseUrl . urlencode($fileName);
$model->addMediaFromUrl($url)->usingFileName($fileName)->toMediaCollection('image');
```

DenMette:
Resolves #1929

- I added the file "antarèsthumb.jpg" in "TestSupport\testfiles" to make sure I didn't break the copy.
- It would be nice to copy the file to the spatie docs, so we can change the url. I used the one provided in the issue.
- Added 2 tests.


freekmurze:
Thanks!

PR #1938, opened 2020-06-22
Report an error when it can't delete a directory

DenMette:
Resolves issue: #1737, #1625 and #1912

Duplicate of: https://github.com/spatie/laravel-medialibrary/pull/1912, because he didn't fix the issue.


freekmurze:
Thanks!

PR #1965, opened 2020-07-22
Revert use statement with correct path to MediaStream Class

disappeared1:
I am an idiot, really sorry about this but a while back I sent a pull request to update this. This needs to be reverted back as it was originally correct. I didn't realize I was on v7 and v8 got a huge change in namespaces. 

Pull request below that needs to be reverted. 
https://github.com/spatie/laravel-medialibrary/pull/1958/commits/bea8f9c3e6a9c7e356e9a3532d9f721dc0a595a1


freekmurze:
Thanks!

Issue #1159, opened 2018-06-20
List media collections


ConstantMath:
Hi, thanks for this great package.
I was wondering if there is a way to list the Media Collections that are defined inside the model with the  `registerMediaCollections()`function.

Something like `Article->getAllRegisteredMediaCollections()` 
Tx

ConstantMath:
Thanks, yes I got the idea but my problem is that I need this information even if there is no medias in the DB.

freekmurze:
Right now there is no easy way, but I'd accept a fully documented and tested method that adds this to the package.

Issue #803, opened 2017-10-18
load ALL media & collections


angryalf:
Hi!

1) is any way to load all media (insteed $model->load('media'), like getMedia() but from all collections ) ?

2) is any way to  get all collections (with or without media) ?

paulprok:
$model->media
$model->media->keyBy('collection_name')

maybe

freekmurze:
This indeed seems like the way to go:

```
$model->media->keyBy('collection_name')
```

PR #1976, opened 2020-08-05
Get registered media collections

squatto:
I have the need to dynamically retrieve the registered media collections on different models in my project and didn't see any existing functionality to do so. I found issue #1159 where it was requested and decided to implement it. (edit: also issues #312 and #803)

It's a simple change, but it removes the need to manually call `$model->registerMediaCollections()` and then access `$model->mediaCollections` directly (with the added bonus of it returning a `Collection` instead of an `array`).

The supporting test checks that the `avatar` media collection that is registered in the test-wide `TestModel` instance is returned, is the right class, and is the right name.

Feedback/suggestions/ideas are absolutely welcomed. Thanks!

### DIFF
--- a/docs/working-with-media-collections/defining-media-collections.md
+++ b/docs/working-with-media-collections/defining-media-collections.md
@@ -23,6 +23,16 @@ public function registerMediaCollections(): void
 }
 ```
 
+## Getting registered media collections
+
+To retrieve all registered media collections on your model you can use the `getRegisteredMediaCollections` method.  
+
+```php
+$mediaCollections = $yourModel->getRegisteredMediaCollections();
+```
+
+This returns a collection of `MediaCollection` objects.
+
 ## Defining a fallback URL or path
 
 If your media collection does not contain any items, calling `getFirstMediaUrl` or `getFirstMediaPath` will return `null`. You can change this by setting a fallback url and/or path using `useFallbackUrl` and `useFallbackPath`.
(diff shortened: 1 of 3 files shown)


freekmurze:
Thanks!

squatto:
My pleasure!

jenniferpoo:
hi freekmurze

i'm having hard time trying to figure out HOW i could save files into a dynamic subfolder according to record_type created. 

I have implemented customPathGenerator from PathGenerator -> however the path is still "fixed" .

Is there a way i could pass a dynamic subfolder name and store the path in media table?

squatto:
> hi freekmurze
> 
> i'm having hard time trying to figure out HOW i could save files into a dynamic subfolder according to record_type created.
> 
> I have implemented customPathGenerator from PathGenerator -> however the path is still "fixed" .
> 
> Is there a way i could pass a dynamic subfolder name and store the path in media table?

@jenniferpoo you will have better luck submitting this question somewhere besides here - I would recommend [creating an issue with your question](https://github.com/spatie/laravel-medialibrary/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc) or you could try [making a post on r/laravel on Reddit](https://www.reddit.com/r/laravel/).

jenniferpoo:
Sorry for posted in wrong place.   
  
Resubmit to the link you provided.   Thank U   
  

  

  
  

  
  
>

PR #1977, opened 2020-08-07
Add hyphenated separator to default sanitizer

squatto:
I've been adding a lot of media files today and came across a minor annoyance. If you have a filename that has a hyphenated separator " - " (space hyphen space) you end up with an odd filename due to the way the default sanitizer works. " - " becomes "---", when really it should become a single hyphen "-". For example:

Filename: `A9 - Bird (Quail).pdf`
Sanitized: `A9---Bird-(Quail).pdf`
Should Be: `A9-Bird-(Quail).pdf`

Filename: `Templates - Vehicles - Truck.pdf`
Sanitized: `Templates---Vehicles---Truck.pdf`
Should Be: `Templates-Vehicles-Truck.pdf`

This PR simply adds " - " to the array of strings that are replaced by a single hyphen in the default sanitizer. I also updated the sanitizer example in the docs and added a supporting test.

Thanks!

### DIFF
--- a/docs/basic-usage/retrieving-media.md
+++ b/docs/basic-usage/retrieving-media.md
@@ -69,7 +69,7 @@ You can sanitize the filename using a callable:
 $yourModel
    ->addMedia($pathToFile)
    ->sanitizingFileName(function($fileName) {
-      return strtolower(str_replace(['#', '/', '\\', ' '], '-', $fileName));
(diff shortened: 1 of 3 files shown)


freekmurze:
Although I think not many users will be effected, this is a breaking change. I'll revisit this when creating the next major version.

squatto:
That's understandable. For now I will use a custom sanitizer and I'll keep an eye on the next major release. Thank you!

PR #1991, opened 2020-08-13
Update MediaStream.php

MoogyG:


divdax:
In my case we have mp3 files uploaded from users with ugly filenames. We read the id3 tag from the mp3 to generate a consistent filename like "Artist - Title (Version).mp3", so it would be a good feature to use a dynamic filename.

@freekmurze any chance this feature get merged?

darkmodeee:
can we implement this feature?

PR #1993, opened 2020-08-14
[fix]  keep sizes 1px if width is 0px

ahmedsayedabdelsalam:
when the parent element of the image tag is `display:none` 
this expression `this.getBoundingClientRect().width` return 0
and **sizes** attribute will equal **'0vw'** 
that makes the image not displayed with the aspect ratio 
I think we have to use the default '1px' if the width is zero.

Screen Shot 2020-08-14 at 5 15 54 AM

laravel-media-library


sebastiandedeyne (approved):

sebastiandedeyne:
Looks good to me! Could you resolve the merge conflict? Will merge after :)

ahmedsayedabdelsalam:
fixed

freekmurze:
Thanks!

PR #1995, opened 2020-08-14
Allow for attributes when copying a media item

Chris1904:
The issue I am trying to solve right now is that when you create a media item `withAttributes()`:

```php
auth()->user()
    ->addMediaFromDisk($image, $this->diskName())
    ->preservingOriginal()
    ->withAttributes([
        'team_id' => $teamId,
    ])
    ->toMediaCollection();
```

and down the road you want to "duplicate"/copy this item, I could usually use the following API:

```php
$media->copy(auth()->user());
```

The problem that arises is that the media item was created *with* custom attributes, which will throw an error when `copy`-ing the item because those attributes are disregarded, similar to this error:

`SQLSTATE[HY000]: General error: 1366 Incorrect integer value: '...' for column 'team_id' at row 1`

This PR adds the ability to add custom attributes when copying media.

### DIFF
--- a/src/MediaCollections/Models/Media.php
+++ b/src/MediaCollections/Models/Media.php
@@ -248,7 +248,7 @@ class Media extends Model implements Responsable, Htmlable
         return $newMedia;
     }
 
-    public function copy(HasMedia $model, $collectionName = 'default', string $diskName = ''): self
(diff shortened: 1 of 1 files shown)


freekmurze:

freekmurze (changes requested):
Please add a test.

freekmurze:
Closing due to inactivity.

PR #1996, opened 2020-08-14
Add ability to change queue connection in config

Luksys3:
Added ability to provide queue connection in config, sometimes you need to use different connections.


freekmurze:
Closing due to inactivity.

PR #2011, opened 2020-08-21
Add the ability to pass options to ZipStream.

mercuryseries:
Example of use:

```php
<?php

namespace App\Http\Controllers\Products;

use App\Models\Product;
use Spatie\MediaLibrary\Support\MediaStream;
use App\Http\Controllers\Controller;
use ZipStream\Option\Archive as ArchiveOptions;

class DownloadProductAttachmentsController extends Controller
{
    public function __invoke(Product $product)
    {
        $downloads = $product->getMedia('attachments');

        $zipOptions = new ArchiveOptions;

        $zipOptions->setZeroHeader(true);

        return MediaStream::create(str_slug($product->name).'.zip', $zipOptions)->addMedia($downloads);
   }
}
```

### DIFF
--- a/src/Support/MediaStream.php
+++ b/src/Support/MediaStream.php
@@ -13,7 +13,7 @@ class MediaStream implements Responsable
 {
     protected string $zipName;
     
-    protected ?ArchiveOptions $zipOptions = null;
+    protected ArchiveOptions $zipOptions;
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks!

PR #2016, opened 2020-08-24
Avoid useless renaming in zip

MoogyG:
Using "zip_filename_prefix" custom property, a zip file can contain files with same name but not in the same directory, so rename it is useless.


freekmurze:
Thanks!

PR #2024, opened 2020-08-26
Added getMedia('*') to retrieving all media with any collection_name

version31:
how can I delete model's media with any collection type.

getMedia() returns all media with default value. but I need all media with any collection type

I offer you to add this feature: getMedia("*")


freekmurze:
Thanks for your work on this, but you can get access to all media by using the relation: `$yourModel->media`.

To delete everything you can do `$yourModel->media->each->delete()`

PR #2031, opened 2020-09-03
Some extra clarifications

ndeblauw:
Added some clarifications/extra information on points where I got stuck with


freekmurze:
Thanks!

PR #2032, opened 2020-09-03
responsive image inline script

willemvb:
Since the `load` event in Safari and Chrome can be triggered before the actual image is rendered or CSS is parsed, the image size calculated in the script is derived from the `width` attribute and can exceed 100vw. 
This renders the responsive effort useless. 

This fix uses `requestAnimationFrame` to return the actual width in the layout.


freekmurze:
@sebastiandedeyne what do you think about this change?

sebastiandedeyne:
Looks good to me!

freekmurze:
Thanks!

PR #2036, opened 2020-09-05
add support to include `ResponsiveImages` based on condition

ctf0:
allow `withResponsiveImages` to be added based on condition, this is useful if you have multiple media collections with the same characteristics but you need the responsive images for only one ex.

```php
{
    $singles = ['cover', 'scan'];

    foreach ($singles as $item) {
        $this->addMediaCollection($item)
            ->useFallbackUrl(asset('images/ph.png'))
            ->useFallbackPath(public_path('images/ph.png'))
            ->withResponsiveImages($item != 'scan')
            ->singleFile();
    }
}
```

### DIFF
--- a/src/MediaCollections/MediaCollection.php
+++ b/src/MediaCollections/MediaCollection.php
@@ -123,11 +123,7 @@ class MediaCollection
 
     public function withResponsiveImagesIf($condition): self
     {
-        if (is_callable($condition)) {
-            $this->generateResponsiveImages = $condition();
-        } else {
-            $this->generateResponsiveImages = $condition;
-        }
+        $this->generateResponsiveImages = (bool) (is_callable($condition) ? $condition() : $condition);
 
         return $this;
     }
--- a/src/MediaCollections/MediaCollection.php
+++ b/src/MediaCollections/MediaCollection.php
@@ -2,8 +2,8 @@
 
 namespace Spatie\MediaLibrary\MediaCollections;
 
-use Illuminate\Support\Traits\Macroable;
 use InvalidArgumentException;
+use Illuminate\Support\Traits\Macroable;
 
 class MediaCollection
 {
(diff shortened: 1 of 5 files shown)


freekmurze:
I like the requested functionality, but not the implementation. Instead of changing the existing function, let's add `withResponsiveImagesIf` that accepts a boolean.

Also add a test to make sure that it is working.

freekmurze:
Feel free to work further on this PR.

ctf0:
i thought you dont want to bloat the lib with extra methods, specially that the new change is backward compatible with the current version usage.

i was also thinking of turning the variable into a closure to give more flexibility, what u think ?

freekmurze:
Let's support both a bool and a callable

ctf0:
np

ctf0:
as for tests, am not sure where to add the test for this one, i found `withResponsiveImages` in 8 test file.

src/MediaCollections/MediaCollection.php
```diff
@@ -120,4 +120,15 @@ public function withResponsiveImages(): self
 
         return $this;
     }
+
+    public function withResponsiveImagesIf($condition): self
+    {
+        if (is_callable($condition)) {
```
freekmurze, line 126: Use a ternary here, so we can lose the `else` statement.

freekmurze (changes requested):

src/MediaCollections/MediaCollection.php
ctf0: will do, was also thinking of adding an exception to make sure we only get this 2 types.

freekmurze:
> as for tests, am not sure where to add the test for this one, i found `withResponsiveImages` in 8 test file.

Put them where you think is a logical place

ctf0:
any updates ?

freekmurze:
Thanks!

rcerljenko:
Can we also have this?
https://github.com/spatie/laravel-medialibrary/discussions/2083

PR #2054, opened 2020-09-20
Allow to change the way the images are being downloaded

chrysanthos:
We have a case were the images cannot be downloaded using `fopen` but instead a client should be used. With this changes we will be able to change how the image is downloaded and instead use either curl (with additional headers) or something else.

Let me know if you want me to add an interface there or even another downloader (curl) as an option.

This is backwards compatible.

### DIFF
--- a/config/media-library.php
+++ b/config/media-library.php
@@ -92,13 +92,6 @@ return [
     'url_generator' => Spatie\MediaLibrary\Support\UrlGenerator\DefaultUrlGenerator::class,
 
     /*
-     * When using the addMediaFromUrl method you may want to replace the default downloader.
-     * This is particularly useful when the url of the image is behind a firewall and
-     * need to add additional flags, possibly using curl.
-     */
-    'media_downloader' => Spatie\MediaLibrary\Downloaders\DefaultDownloader::class,
-
-    /*
      * Whether to activate versioning when urls to files get generated.
      * When activated, this attaches a ?v=xx query string to the URL.
      */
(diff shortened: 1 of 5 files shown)


src/Downloaders/DefaultDownloader.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Spatie\MediaLibrary\Downloaders;
+
+use Spatie\MediaLibrary\MediaCollections\Exceptions\UnreachableUrl;
+
+class DefaultDownloader
+{
+    public function getTempFile($url)
```
freekmurze, line 9: Add a typehint and return type hint here

config/media-library.php
```diff
@@ -91,6 +91,13 @@
      */
     'url_generator' => Spatie\MediaLibrary\Support\UrlGenerator\DefaultUrlGenerator::class,
 
+    /*
+     * When using the addMediaFromUrl method you may want to replace the default downloader.
+     * This is particularly useful when the url of the image is behind a firewall and
+     * need to add additional flags, possibly using curl.
+     */
+    'media_downloader' => Spatie\MediaLibrary\Downloaders\DefaultDownloader::class,
```
freekmurze, line 99: I don't think this functionality warrants at place this high in the config file. Let's move this to the bottom.

freekmurze (changes requested):
Could you also add a tests and update the docs with an example of the new functionality.

chrysanthos:
Thanks for the review @freekmurze. I made the fixes and added the docs. 

It seems like there are a lot of tests that already cover the whole method [here](https://github.com/spatie/laravel-medialibrary/blob/master/tests/Feature/FileAdder/IntegrationTest.php#L298-L319) and [here](https://github.com/spatie/laravel-medialibrary/blob/master/tests/Feature/FileAdder/IntegrationTest.php#L346-L392). Let me know if there is special case which is not handled and you want tested.

src/Downloaders/DefaultDownloader.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace Spatie\MediaLibrary\Downloaders;
+
+use Spatie\MediaLibrary\MediaCollections\Exceptions\UnreachableUrl;
+
+class DefaultDownloader
```
freekmurze, line 7: Let's make this class implement an `Downloader` interface with that `getTempFile` method in it. In the readme and config file it should be mentioned. That a valid downloader is any class that implements that new interface.

freekmurze (changes requested):

chrysanthos:
Hey @freekmurze, does this seem ready to be merged or do you have any more changes you'd like to see?

freekmurze:
I need to find some time to review this.

freekmurze:
Thanks!

PR #2064, opened 2020-09-24
Allow ZipStream 2.0 (backport from v8)

JaZo:
I'm running into dependency constraints as other packages require ZipStream 2.0. Unfortunately we can't use PHP 7.4 yet (latest minus one policy...), so we can't install v8. I was hoping we can backport #1812 to v7 so we're able to install those other dependencies as well.


JaZo:
Thanks!

PR #2073, opened 2020-09-28
Fix file deletion problem

erik-epineer:
When you delete a media item from a media collection the original file and its directory won't be deleted if you use a different disk for storing the conversions, only the files and directories used on the conversion disk will be deleted.

I noticed this when working with a single file collection where I use a private directory for the original uploaded file and a public directory for the conversions. 

My solution is to add a dedicated **deleteDirectory** call for the original file path if a different disk is being used for the conversions.  I've added a simple assertion test to make sure the original file is being deleted when this is the case.


freekmurze:
Nice, thanks!

PR #2076, opened 2020-09-29

ahmed-abdelnasser:
when try migrate:refresh its get an error bec media table not droped


freekmurze:
We don't use down. Add this to your own project.

PR #2080, opened 2020-09-30
Fixed conversions when disk != conversions_disk

rcerljenko:
I didn't write tests because I don't know how to set up test case for that specific case (it involves two different disks).


freekmurze:
Thanks!

Issue #2079, opened 2020-09-30
W3C break : loading="auto" is not an HTML valid value


Okipa:
According to these resources, it seems that the the `auto` value for the `loading` HTML attribute does not exist:
* https://html.spec.whatwg.org/multipage/urls-and-fetching.html#lazy-loading-attribute
* https://html.spec.whatwg.org/multipage/embedded-content.html#attr-img-loading

It is actually breaking W3C XHTML validator.

In the package documentation, it is told that it is a possible value and it seems that it is quite often used: https://spatie.be/docs/laravel-medialibrary/v8/api/defining-conversions#useloadingattributevalue.

Would you accept a PR that would remove the use of the `auto` value, in favor of the `eager` and `lazy` ones ?

sebastiandedeyne:
<img width="936" alt="image" src="https://user-images.githubusercontent.com/1561079/94659339-6984de80-0304-11eb-9ae5-e53dd579067d.png">

I suppose we missed the warning :)

I'd accept a PR that changes the default configuration `default_loading_attribute_value` to `eager`.

Okipa:
Thanks for your quick answer !

Just another question: why not to trigger `lazy` by default as it is one of the real gain offered by responsive images ? (It would still be possible to opt for `eager` loading if the standard behaviour is wished).

sebastiandedeyne:
Because `eager` is still the default browser behaviour, and I don't believe we should be hijacking that.

Maybe the best solution would be to not add any `loading` attribute unless explicitly specified.

Okipa:
I think you are right: the `auto` attribute is supposed to apply the default browser lazy loading behavior, which is the same as not declaring anything.

I'll make my PR forward this direction.

freekmurze:
We'll continue this in #2081

PR #2081, opened 2020-09-30
Fix for default lazy="auto" value

Okipa:
As discussed here https://github.com/spatie/laravel-medialibrary/issues/2079, here is my proposal to replace the default `default_loading_attribute_value` by `null` instead of `auto`, as this last value is not mentioned in W3C specifications and is breaking the W3C Markup Validation Service.

Setting `null` allow to set no `loading` HTML attribute by default (same as the expected behavior for `auto`).


freekmurze:
Thank you!

PR #2086, opened 2020-10-02
added withResponsiveImagesIf() on conversions

rcerljenko:


freekmurze:
Could you add tests that prove that this works?

rcerljenko:
Ok I'll try (I don't have that much experience in writing tests)  :)

sebastiandedeyne:
Why not change the signature of `withResponsiveImages()` to `withResponsiveImages(bool $withResponsiveImages = true)` instead? That way we don't have to add a new method to the API.

rcerljenko:
@sebastiandedeyne because that was already discussed in the simmilar PR here https://github.com/spatie/laravel-medialibrary/pull/2036

sebastiandedeyne:
I wasn't aware of that, carry on 😅

freekmurze:
Tests seem to fail, could you take a look at that?

rcerljenko:
@freekmurze hi i'm looking it but i think it's not related to my part of the code
maybe i'm wrong but it fails in a completly weird place

freekmurze:
Could you rebase this to the master branch? With any luck we don't get any failing tests after that.

PR #2094, opened 2020-10-05
Global queue conversions management

Okipa:
Following our discussion [here](https://github.com/spatie/laravel-medialibrary/discussions/2088), here is my PR to add an easy way to globally enable/disable queue generation in config.

I've also detected one or two small errors that I have fixed (typos and wrong config call in `tests/Feature/Media/DeleteTest.php`).

Let me know if it's OK for you or if there's something to change.

### DIFF
--- a/config/media-library.php
+++ b/config/media-library.php
@@ -21,6 +21,12 @@ return [
     'queue_name' => '',
 
     /*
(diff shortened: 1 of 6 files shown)


freekmurze:
Thanks!

PR #2098, opened 2020-10-07
[WIP] v9

freekmurze:
- adds support for media library pro
- names of the generated conversions will now be save in a dedicated `generated_conversions` in the `media` table


mikemand:
Is it safe to clear the `generated_conversions` key on the `custom_properties` column when upgrading from v8? Or should I leave that key alone? The upgrade guide says to "copy" the data but doesn't say if I have to leave it in the original column or not.

PR #2104, opened 2020-10-09
Add support for getLastMedia() convenience-method

marijoo:
With this pull request a new convenience-method ~`getLatestMedia()`~ `getLastMedia()` is introduced which returns the latest added media of a collection. This can be useful if a collection is used to store multiple versions of a file and the newest should be returned, for example.

The associated functions `getLastMediaUrl()`, `getLastMediaPath()` and `getLastTemporaryUrl()` were also implemented – the test cases and documentation were adapted accordingly.

**Edit:** Since collections can be re-ordered manually, I renamed `getLatestMedia()` to `getLastMedia()` because this is less opinionated.

### DIFF
--- a/docs/basic-usage/retrieving-media.md
+++ b/docs/basic-usage/retrieving-media.md
@@ -29,6 +29,13 @@ $media = $yourModel->getFirstMedia();
 $url = $yourModel->getFirstMediaUrl();
 ```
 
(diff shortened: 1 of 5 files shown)


freekmurze:
Thank you for your work on this. Because there are no other users asking for this and I like to keep the package as light as can be, I'm going to pass on these methods for now. Add them to your own project.

PR #2105, opened 2020-10-11
Fixed width calculator

rcerljenko:
Introducing the new WidthCalculator => `FixedWidthCalculator`. Sometimes we just need responsive images for fixed breakpoints (eg. 200, 600, 1200 px). It adds one extra config option to set the breakpoints. In addition, if you choose to use this width calculator and the breakpoints array is empty, it fallbacks to good old `FileSizeOptimizedWidthCalculator`. Also, if you set the breakpoint too high for some smaller images, it just won't use that breakpoint (it filters the breakpoints array where the actual picture width is smaller).

PS. If you're interested in this kind of thing I'm sure I can add tests and docs updates...

### DIFF
--- a/config/media-library.php
+++ b/config/media-library.php
@@ -2,31 +2,31 @@
 
 return [
 
-	/*
+    /*
      * The disk on which to store added files and derived images by default. Choose
      * one or more of the disks you've configured in config/filesystems.php.
      */
-	'disk_name' => env('MEDIA_DISK', 'public'),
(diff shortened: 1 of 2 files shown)


freekmurze:
Thank you for our work on this, but we're going to stick to only having and maintaining the current WidthCalculator.

boryn:
Hi @freekmurze! And what about using different (custom) width calculators per collectionName? Now we can use only one from the config for all the collections.

PR #2110, opened 2020-10-12
Add down() method in migration file

theinhlamaw:
This pull request added a down() method in database migration file. Without down() method, the following error occured whenever we refresh the migrations by using the command `php artisan migrate:refresh`.

Error
```
Illuminate\Database\QueryException

SQLSTATE[42S01]: Base table or view already exists: 1050 Table 'media' already exists
```


freekmurze:
We don't use `down` methods for our projects. Add it to your project.

PR #2113, opened 2020-10-15
[Fix] add support for dropping media table

hinimajesty:
On php artisan rollback the media table would be dropped.


freekmurze:
We don't use down in our projects

hinimajesty:
> We don't use down in our projects

Ohk noted. I added it because locally whenever i try to migrate:refresh the media table not deleting breaks the command. Is there something i am doing wrong ? Or perhaps i should force the refresh.

PR #2114, opened 2020-10-16
File namer#1784

Nielsvanpach:
It's already possible to use a custom file namer on conversions. 
This PR allows responsive images to use a mutual custom file namer.


docs/advanced-usage/naming-generated-files.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Naming responsive image files
+
+By default, all conversion files will be named in this format:
+
+```
+{original-file-name-without-extension}___{name-of-the-conversion}_{width}_{height}.{extension}
+```
+
+Just like the conversion file names, you can use another format for naming your files.
```
freekmurze, line 52: This is unclear. So `getConversionFileName` is used as the basis for the responsive image file names?

tests/TestSupport/TestFileNamer.php
```diff
@@ -0,0 +1,21 @@
...
+use Spatie\MediaLibrary\Conversions\Conversion;
+use Spatie\MediaLibrary\Support\FileNamer\FileNamer;
+
+class TestFileNamer extends FileNamer
+{
+    public function getFileName(string $fileName): string
+    {
+        $fileName = pathinfo($fileName, PATHINFO_FILENAME);
+
+        return 'prefix_' . $fileName . '_suffix';
```
freekmurze, line 14: Use string interpolation here

tests/ResponsiveImages/ResponsiveImageTest.php
```diff
@@ -6,6 +6,8 @@
 
 class ResponsiveImageTest extends TestCase
 {
+    public string $file_name = 'test';
```
freekmurze, line 9: For instance variables always use camelCase names.

tests/ResponsiveImages/ResponsiveImageGeneratorFileNamerTest.php
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Spatie\MediaLibrary\Tests\ResponsiveImages;
+
+use Spatie\MediaLibrary\Tests\TestSupport\TestFileNamer;
+
+class ResponsiveImageGeneratorFileNamerTest extends ResponsiveImageGeneratorTest
+{
+    public function setUp(): void
+    {
+        parent::setUp();
```
freekmurze, line 11: Make this breathe by adding newlines between these three lines

tests/ResponsiveImages/ResponsiveImageFileNamerTest.php
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Spatie\MediaLibrary\Tests\ResponsiveImages;
+
+use Spatie\MediaLibrary\Tests\TestSupport\TestFileNamer;
+
+class ResponsiveImageFileNamerTest extends ResponsiveImageTest
+{
+    public function setUp(): void
+    {
+        parent::setUp();
```
freekmurze, line 11: Make this breathe by adding newlines between these three lines

freekmurze (changes requested):
We're almost there!

docs/advanced-usage/naming-generated-files.md
Nielsvanpach: `DefaultFileNamer` needs 2 methods:
- `getFileName` used by conversions and responsive images. For example, a timestamp prefix can be added here. This will affect both the conversion file names and the responsive image file names.
- `getConversionFileName` is only used for conversions. For example `file-name---thumb` can be renamed to `file-name_thumb`.

freekmurze: I still don't understand this fully.  To me it seems strange that `getFileName` is also used for conversion files, as `getConversionFileName` already handles that. It might be best to add a couple very explicit examples to the docs.

Nielsvanpach: I see what you mean. I'll have another look at a different solution

Nielsvanpach: Updated:

 `getFileName` is removed from the FileNamer. It needs these 2 methods (which do not interfere each other):
- `getConversionFileName`
- `getResponsiveImageFileName`

src/Support/FileNamer/DefaultFileNamer.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Spatie\MediaLibrary\Support\FileNamer;
+
+use Spatie\MediaLibrary\Conversions\Conversion;
+
+class DefaultFileNamer extends FileNamer
+{
+    public function getConversionFileName(string $fileName, Conversion $conversion): string
```
freekmurze, line 9: maybe rename all these methods without the `get`

freekmurze (changes requested):
One nitpick!

freekmurze:
Thanks!

PR #2132, opened 2020-10-26
Set height attribute

rocramer:
The image tags do not contain a height attribute (only the width). This can lead to layout shifts while rendering. Google recommends adding the width and height attributes so modern browsers can calculate the aspect ratio automatically: https://web.dev/optimize-cls/#modern-best-practice


willemvb (approved):

sebastiandedeyne (approved):

freekmurze:
Thanks!

PR #2136, opened 2020-10-27
make media uuid creation optional

aksonnic:
Make the automatic uuid creation optional, and eliminate the need for a spurious DB column, for apps that don't need it. Default to creating uuids. Can be disabled with the config key `media-library.uses_media_uuids`, or by overriding `Media::usesUuids()`.


freekmurze:
Thank you for your work on this. 

This feels like a micro optimisation now to be. I'd rather have the current simplicity. If avoiding the uuid is crucial to your project, use a custom model where you make the necessary changes.

aksonnic:
You've broken every site that your community uses MediaLibrary on until they add a DB column that they aren't going to use. This is hardly a micro-optimization.

freekmurze:
This was not a breaking change as it was added in a new major version.

Issue #2121, opened 2020-10-21
Responsive image isn't used, when filename ends with "_"


bernhardh:
If you have images with filenames ending with "_" they will get rendered as "regular" images (they will use the `image.blade.php` instead of the `responsiveImage.blade.php` view), even if you use `withResponsiveImages()` method. The responsive files and the base64 encoded version are generated.

I have tracked it down and the error is because of the following:

The problem is with one line in the `Spatie\MediaLibrary\MediaCollections\Models\Media\RegisteredResponsiveImages` constructor:

```php
->filter(fn (ResponsiveImage $responsiveImage) => $responsiveImage->generatedFor() === $this->generatedFor);
```

where the `$responsiveImage->generatedFor()` looks like this:

```php
    public function generatedFor(): string
    {
        $propertyParts = $this->getPropertyParts();

        array_pop($propertyParts);

        array_pop($propertyParts);

        return implode('_', $propertyParts);
    }
```

The problem is, that if you image filename ends with "_", the result of the `generatedFor` method is `_YOURCOLLECTIONNAME` instead of `YOURCOLECTIONNAME`, so the filter in the `RegisteredResponsiveImages` constructor will not find the correct collection.

freekmurze:
I'd accept a PR that fixes this. 

@Nielsvanpach can you make sure this is fixed in v9?

freekmurze:
This one can be closed I guess.

PR #2140, opened 2020-10-30
find last occurence of starting characters

Nielsvanpach:
If a filename ends with an underscore (test_.jpg), the responsive images end up with four underscores in its name.
We used to trim at the first occurence of 3 underscores.This means the collection name starts with the last remaining underscore and results in responsive images not being rendered as expected.
I have fixed this by finding the position of the last occurence so the collection name will not start with an underscore in this case.

Fixes https://github.com/spatie/laravel-medialibrary/issues/2121


freekmurze:
Thanks!

PR #2146, opened 2020-10-31
Fix link to Laravel File Storage docs

ahallhognason:
The current link points to Laravel 5.6 which is not supported by this package.

This updates the link to point to Laravel on the latest version regardless of what version is the latest now (laravel docs automatically redirects when the version isn't included).


freekmurze:
Thanks!

PR #2151, opened 2020-11-02
Fix the version number for Version 9 in installation doc

prettifystudio:
Hello, I noticed in the documentations in the version 9 (master), The composer command refers to version 8.0.0 . 
I changed both of medialibrary and assumed that medialibrary-pro also has version 9.


freekmurze:
Thanks!

PR #2153, opened 2020-11-03
Add support for audio files

Synchro:
Having seen numerous requests in past issues to add support for thumbnails for audio files, and needing it myself, I thought I'd make one. It was quite straightforward since it uses the same ffmpeg config as is already used for video. It creates square thumbnails like this:

test

The foreground and background colours can be changed via config properties. The PR includes docs and tests, and a collection of sample files in various formats to test with.

The generation process is not ideal. ffmpeg generates transparent waveform images, but I could not find a way to use that as an overlay on an existing background colour using the Spatie Image library, so it uses a `background` manipulation to fill in the background colour, which results in some fringing around the edges of the waveform. However, it still looks reasonable, especially when reduced in size from the 1024x1024 native size it generates.


freekmurze:
I think this awesome functionality, really impressive.

However, I'm not going to merge this in, because I don't want to maintain this code. It would be nice if you would put this code in a separate package, so you can keep maintaining the code. I'd happily put a link to your package in our documentation.

Gummibeer:
@Synchro when you add this to your own package - could you make the size of the generated image configurable? 🤔
Most usecases will be something like SoundCloud or similar where the image is displayed in a much wider view.
Even better would be SVG output - but ffmpeg doesn't support SVG by its own. 😟

Synchro:
Dang. The problem with doing that is that it's quite difficult to maintain this as a separate package because media library doesn't have any kind of plug-in ability, so I'd need to maintain a complete fork and put it in packagist in order to make it usable. While I can see your case for not maintaining it, it would mean that I need to maintain the whole of the rest of ML in my own fork! Unless you can suggest a better way of doing that?

@Gummibeer – that's the sort of thing I was thinking about when I asked [this](https://github.com/spatie/laravel-medialibrary/discussions/2152), though I guess the simplest thing to do is to use config properties to set the size, which would be a simple change, but merging config properties into an existing ML installation would be tricky.

freekmurze:
I though we have a plug-in ability for this. You just need to add it to this array in the config: https://github.com/spatie/laravel-medialibrary/blob/2af7b291be56f3eaea1c7ee63bf1ff71a4ca9017/config/media-library.php#L92-L98

Synchro:
Ah, that's true. I'm not sure how to add to that at runtime though, nor how to install a config file to set my own params. Do you know of any other packages that extend ML in a similar way that I can use as a reference?

Gummibeer:
For the sizes: could you use the image variants?
So let's say I configure a `thumbnail` with `500x500` with `blue` background and a `full` with `1920x200` and `white` background.
I haven't worked enough in the last times with medialibrary to know it myself. But that would be MY favorite. This way it would also configure multiple images if needed in different aspect ratios instead of only scaling the first one. 🤔
Would result in a much higher server load but also result in much better image quality.

Regarding the different package problem: shouldn't it be enough if you provide the `ImageGenerator` class which the user should add to the `image_generators` config key? 🤔
And if you don't want to inject the config you have several options:
* ML config file - add it to installation step of your package
* own config file - not cool for two values which are only a plugin to another package

using service container primitive bindings - https://laravel.com/docs/8.x/container#binding-primitives as the generator is instantiated via the container.
https://github.com/spatie/laravel-medialibrary/blob/caa4e9b7bc3e2775dc0ee3c61a87e7030b2d73ae/src/Conversions/ImageGenerators/ImageGeneratorFactory.php#L10-L14

Gummibeer:
@Synchro your package should only add the generator class and have a proper README installation/configuration section.
You even don't need a service provider - except a deferred one if you want to use the primitive binding solution.

The whole process will be:
1. `composer require medialibrary medialibrary-audio-generator` will install both packages
2. `artisan vendor:publish medialibrary` will publish the original config
3. follow the audio-generator instructions to add `AudioImageGenerator::class` to config and do however you want to configure size, colors etc.

Synchro:
@Gummibeer I was confused by how thumbnail sizes work because none of the other image generators seem to specify a size - they just use whatever comes back from the source, e.g. video thumbnails use the full, original frame size of the video. Audio files obviously have no inherent size, so I had to pick one, and square is neutral! As far as I can see, ML then takes this file and does its resize-on-the-fly thing for each size a client wants, so I think you're asking for features that ML doesn't provide.

Gummibeer:
Yeah, thought so - but if you add the color configuration I would like to also be able to configure the size.
It can be 1024 square by default. But possibly I want a high-res square 2048 or a 16/9 aspect-ratio.
I would absolutely prefer the first (original config file) or primitive injection solution to configure everything. Possibly it could even use the primitive injection with the config by default. This way the user is able to don't adjust the original config but use the injection directly.

freekmurze:
@Synchro would this PR help? #2156

Gummibeer:
@freekmurze from my perspective this should solve all issues! 🚀

PR #2154, opened 2020-11-03
Add pdf to image as a dependency

mokhosh:
I don't know why it's not mentioned in the docs, but I had to do a lot of tinkering in the `artisan tinker` and follow the `media-library:regenerate` down the rabbit hole to find out that I have to install this package to get pdf thumbnails.


freekmurze:
Thanks!

PR #2156, opened 2020-11-04
Allow image generators to accept config

freekmurze:
When registering a custom image generator in the `media-library` config file, you can pass values like this

```php
// in config/laravel-medialibrary.php
'image_generators' => [
    // ...
    YourImageGenerator::class => ['myArgument' => 'value', 'myOtherArgument' => 'otherValue'],
], 
```

In your custom image generator you can accept those arguments in the constructor.

```php
class YourImageGenerator
{
    public function __construct(string $myArgument, string $myOtherArgument)
    {
        // do something with these arguments
    }
}
```

### DIFF
--- a/docs/converting-other-file-types/creating-a-custom-image-generator.md
+++ b/docs/converting-other-file-types/creating-a-custom-image-generator.md
@@ -7,7 +7,7 @@ If you want to generate a conversion for a file type that is not covered out of 
 
 In the following example we'll create a custom generator that can convert a Powerpoint to an image.
 
-## Creating the custom generator
+## Creating a custom image generator
 
 The first step for creating a custom generator is to create a class that extends `Spatie\MediaLibrary\Conversions\ImageGenerators\ImageGenerator`:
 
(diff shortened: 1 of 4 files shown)


Synchro:
This implementation turns named arguments into positional ones, which seems a bit uncomfortable. Given there's likely to be some diversity in what plugins might want, would it not be better to pass the whole array as a single param so that their names can be retained? Like this:

```php
public function __construct(array $myArguments)
{
    if (array_key_exists('myArgument', $myarguments)) {
        $this->myArg = $myArguments['myArgument'];
    } else {
        $this->myArg = env('MYARG', 'somevalue');
    }
```

freekmurze:
@Synchro The arguments will be passed by name, not by position. I've added a test `image_generators_will_receive_config_parameters_by_name` to make that clear.

freekmurze:
I'll tag this. @Synchro let me know if you need more assistance with that audio image package 🙂

Synchro:
Thanks for making this change - it means I don't have to make up my own config vars, and enabling the plugin will be a 1-line change in ML.

My biggest issue at the moment is how to test. With the earlier PR I could just piggy-back on your existing test setup, but at the moment it looks like I have to pull in and configure the whole of Laravel and media library just so I can call `config` and have values set for `media-library.ffmpeg_path`!

Synchro:
[Initial setup is done](https://github.com/Synchro/laravel-medialibrary-audio), tests not working yet!

freekmurze:
When it's ready to be used, send a PR to our docs that adds a link to your package 👍

PR #2171, opened 2020-11-12
Update media-library.php

Jon78:
This PR streamlines the default config with the config in spatie/laravel-image-optimizer, resulting in better compression for jpegs and happier people all around


freekmurze:
Thanks!

PR #2174, opened 2020-11-13
Fix missing test in MediaStreamTest.php

Arkhas:
While working on a new feature, I saw a missing `/** @test */` on the `media_with_zip_file_folder_prefix_property_saved_in_correct_zip_folder_and_correct_suffix`

And just adding it would makes the test not pass.

It seems that it might be just a work in progress left behind as the setup was not matching the expecting results

I run `php-cs-fixer fix tests/Support/MediaStreamTest.php` so it should be fine :)


freekmurze:
Thanks!

PR #2176, opened 2020-11-13
Allow customization of the file name of a Media when adding it to a zip

Arkhas:
Currently you can only specify the sub folder when adding a file to a zip.

I would allow the customization of the file name while doing it by adding a custom property : 

`zip_filename_suffix`

Update in the documentation : 

https://github.com/spatie/laravel-medialibrary/pull/2175

### DIFF
--- a/src/Support/MediaStream.php
+++ b/src/Support/MediaStream.php
@@ -70,7 +70,7 @@ class MediaStream implements Responsable
     {
         $headers = [
             'Content-Disposition' => "attachment; filename=\"{$this->zipName}\"",
-            'Content-Type' => 'application/octet-stream',
+            'Content-Type'        => 'application/octet-stream',
         ];
 
         return new StreamedResponse(fn () => $this->getZipStream(), 200, $headers);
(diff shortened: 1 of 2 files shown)


freekmurze:
Thanks for this, but I'm not going to merge this in. This PR brings some extra complexity that I'd not willing to take on for now.

Arkhas:
Then how can I customize a filename when adding it to a zip when the file is coming from S3 ? 

Right now, when zipping files coming from S3, you just get a messy folder with all file with random names instead of the `name` attribute.

A less complex solution would be to directly use the 'name' of the file but it may be a breaking change when adding file to a zip, what do you think ?

Arkhas:
I can also make this part : 

```
protected function getZipFileNameWithoutExtension(Collection $mediaItems, int $currentIndex): string
{
        return $mediaItems[$currentIndex]->hasCustomProperty('zip_filename_suffix') ? $mediaItems[$currentIndex]->getCustomProperty('zip_filename_suffix') : pathinfo($mediaItems[$currentIndex]->file_name, PATHINFO_FILENAME);
}
```

more simple like this : 

```
protected function getZipFileNameWithoutExtension(Collection $mediaItems, int $currentIndex): string
{
        return $mediaItems[$currentIndex]->zip_file_name ?: pathinfo($mediaItems[$currentIndex]->file_name, PATHINFO_FILENAME);
}
```

and the only thing to do is adding a getter to your Media model : 

```
public function getZipFileNameAttribute()
{
        return 'bar';
}
```

src/Support/MediaStream.php
```diff
@@ -98,39 +98,50 @@ public function getZipStream(): ZipStream
     protected function getZipStreamContents(): Collection
     {
         return $this->mediaItems->map(fn (Media $media, $mediaItemIndex) => [
-            'fileNameInZip' => $this->getZipFileNamePrefix($this->mediaItems, $mediaItemIndex).$this->getFileNameWithSuffix($this->mediaItems, $mediaItemIndex),
-            'media' => $media,
-        ]);
+        'fileNameInZip' => $this->getZipFileNamePrefix($this->mediaItems, $mediaItemIndex) . $this->getFileNameWithSuffix($this->mediaItems, $mediaItemIndex),
```
freekmurze, line 101: Could you make sure that there are no code style changes in this PR. Unnecessary changes like his make the PR harder to review.

freekmurze (changes requested):

freekmurze:
Closing due to inactivity.

PR #2180, opened 2020-11-15
Fix migrate:refresh issue media already exists

MohmmedAshraf:
This PR will fix the issue when running the `php artisan migrate:refresh` command.


freekmurze:
Thanks, but I'm not going to merge it as we don't use down migrations in any of our projects.

PR #2191, opened 2020-11-19
Update PerformConversionsJob.php to automatically delete job if model is missing

ractoon:
Automatically delete a queued job if the model is missing, e.g. if media is deleted before the job has a chance to process.

Utilizes the [Ignore Missing Models](https://laravel.com/docs/8.x/queues#ignoring-missing-models) behavior in Laravel.


freekmurze:
That's nice, thanks!

PR #2195, opened 2020-11-23
Add down() to create_media_table.php.stub

radmen:
The generated migration has no `down()` method. I don't know if this is intended. If not, this PR adds the missing `down()` method.


freekmurze:
We don't use down in our projects.

PR #2206, opened 2020-11-27
adds pdf-to-image back to required packages

thejacer87:


freekmurze:
I think it's better to send a PR to the docs.

PR #2219, opened 2020-12-01
Support path with uuid

masterix21:
It solves the problems of UUID change using the `TemporaryUpload` model from laravel-medialibrary-pro.


freekmurze:
I very much appreciate your help, but I'm not going to merge this in.

I think this adds complexity for something the vast majority of users don't need. 

You can add these changes to your own project, by using a custom Media model and applying these changes there.

masterix21:
I trust you, but I think the UUID part isn't only a part of my own project: it's a part of the package itself. Also, I think UUID should be unique and only one time generated for a model, and the part of package code that regenerates the UUID is no-sense.

If you agree with me, I will remove the UUID regeneration from your package, but before, I'd like to hear your opinion to avoid working on a refused PR. 

Thanks.

freekmurze:
Reconsidering this one...

src/MediaCollections/Models/Concerns/PathUuidSafe.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace Spatie\MediaLibrary\MediaCollections\Models\Concerns;
+
+use Illuminate\Support\Facades\Storage;
+use Spatie\MediaLibrary\MediaCollections\Models\Media;
+use Spatie\MediaLibrary\Support\PathGenerator\PathGeneratorFactory;
+
+trait PathUuidSafe
```
freekmurze, line 9: Please rename this to `SupportsUuidInPaths`

src/MediaCollections/Models/Concerns/PathUuidSafe.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    public static function bootPathUuidSafe()
+    {
+        if (! config('media-library.path_uses_uuid', false)) {
```
freekmurze, line 13: The `false` default can be removed, the config value is always present.

src/MediaCollections/Models/Concerns/PathUuidSafe.php
```diff
@@ -0,0 +1,31 @@
...
+            return;
+        }
+
+        static::updating(function (Media $media) {
```
freekmurze, line 17: Could you add a test for this behaviour?

freekmurze (changes requested):
Let's make a few final polishments before merging it in.

masterix21:
What do you think if I rename `media-library.path_uses_uuid` to `media-library.rename_path_on_media_saving`, and obviously, I rename all other parts of code?

Because I think it could be used any time, for any custom PathGenerator, for who uses the UUID or for who uses the collection, or any other part of the model that could change in time.

What's your idea?

freekmurze:
How about `moves_media_on_update`?

masterix21:
It sounds good.

masterix21:
or these?
- config `keep_consistent_path`
- trait `HasConsistentPath`

freekmurze:
Good suggestions, but I think I like `moves_media_on_update` most.

masterix21:
Ok 😉

src/MediaCollections/Models/Concerns/PathUuidSafe.php
masterix21: Renamed to `MovesOnUpdate`. Is it ok?

src/MediaCollections/Models/Concerns/PathUuidSafe.php
masterix21: Ok

src/MediaCollections/Models/Concerns/PathUuidSafe.php
masterix21: Yes, sure.

masterix21:
@freekmurze, please don't hate me. I have removed the trait and moved the code inside the observer, so I created a method inside the filesystem file.

freekmurze:
> please don't hate me. 

Never! ❤️

src/MediaCollections/Filesystem.php
```diff
@@ -229,6 +229,16 @@ public function syncFileNames(Media $media): void
         $this->renameConversionFiles($media);
     }
 
+    public function syncMediaPath(Media $media): void
+    {
+        $factory = PathGeneratorFactory::create();
+
+        $oldMedia = (clone $media)->fill($media->getOriginal());
+
+        $this->filesystem->disk($media->disk)
```
freekmurze, line 238: Maybe we should compare the old path and with the new path before calling `move`

freekmurze (changes requested):
Looks good. I do see one improvement.

src/MediaCollections/Filesystem.php
masterix21: Yes, it's good. So, we can remove the config `moves_media_on_update` and call the sync method when the paths are not equals. Is it right?

freekmurze: I would still only do the move if the `moves_media_on_update` is set to `true`. 

The base idea is that after upload, medialibrary won't any files, unless you explicitly enable it. Feel safer to me!

freekmurze (approved):

freekmurze:
Thanks!

lostincode:
@masterix21 @freekmurze So this solution is great but will not work with `s3` disk since the filesystem cannot move an entire directory. (took me awhile to figure that one out).

The universal solution for all disks is something like this:

```php

    public function updating(Media $media)
    {
        $oldMedia = (clone $media)->fill($media->getOriginal());

        if ($oldMedia->getPath() === $media->getPath()) {
            return;
        }

        $factory = PathGeneratorFactory::create();

        $oldPath = $factory->getPath($oldMedia);
        $newPath = $factory->getPath($media);

        $images = Storage::disk('s3')->allFiles($oldPath);

        foreach ($images as $img) {
            $moveTo = str_replace($oldPath, $newPath, $img);
            Storage::disk('s3')->move($img, $moveTo);
        }
    }

```

Since with `s3` disk you can only move a single file at a time and not the full directory. It seems like not many people use the `s3` driver with UUID in path so I'm just sharing a fix for anyone using this combo like me 😬

masterix21:
@lostincode great, thanks for your share.

dhcmega:
Hi
I just spent a few hours trying to understand why this didn't work:

```
    {
        return "media-library/{$account_id}/{$media->getKey()}/{$media->uuid}";
    }
```

It was because the model gets created with a different UUID than the one it received. But the file gets saved using the original one. When it tries to move it, the uuid is different (don't know why) and fails to find the file.
Also noticed that `getPath()` returns the path with `/` at the end and it is used in `Filesystem::getStream()` with an extra `/` when concatenating.

I don't like the idea of using `md5($media->id)` because every time the file is read, it will have to recalculate de md5.

Thanks

greeto:
What was the solution here?
users see this /media/2/my-image.png

I would like them to see /media/UUID/my-image.png

hakobfromlightin:
Seems like there is no solution for this bug. When using storage, the files are not moved. This is clearly a bug.

PR #2230, opened 2020-12-11
Support toArray method used by Inertiajs

koentjuh1:
Refactored MediaCollection.php, extracted the collection mapping and applied it to toHtml, jsonSerialize and added toArray method (used by Inertiajs eg.)


freekmurze:
Seems like the tests are failing for this one, could you take a look at that?

freekmurze:
Closing due to inactivity.

Issue #2226, opened 2020-12-09
[BUG] Race-condition for media with both queued and nonQueued conversions


naabster:
possibly connected to #2213 and #1304

Situation: I have a model with multiple conversions, one of them is nonQueued:

    public function registerMediaConversions(Media $media = null): void
    {
        $this->addMediaConversion('card')->nonQueued()->fit('crop', 736, 464);
        $this->addMediaConversion('thumb')->fit('crop', 368, 232)->sharpen(10);
        $this->addMediaConversion('medium')->width(1600)->height(1600);
    }

What we are seeing: Randomly, new uploaded media will have either 'thumb' and 'medium' or only 'card' in generated_conversions column. If you look at the file system, all three conversions are actually there, but the database is not up to date. This happens on a single queue worker locally as well as on Vapor based deployments. 

I think the Problem is a typical race condition: 

```
class FileManipulator
{
    public function createDerivedFiles(
        Media $media,
        array $onlyConversionNames = [],
        bool $onlyMissing = false
    ): void {
        if (!$this->canConvertMedia($media)) {
            return;
        }

        [$queuedConversions, $conversions] = ConversionCollection::createForMedia($media)
            ->filter(function (Conversion $conversion) use ($onlyConversionNames) {
                if (count($onlyConversionNames) === 0) {
                    return true;
                }

                return in_array($conversion->getName(), $onlyConversionNames);
            })
            ->filter(fn (Conversion $conversion) => $conversion->shouldBePerformedOn($media->collection_name))
            ->partition(fn (Conversion $conversion) => $conversion->shouldBeQueued());

        $this
            ->dispatchQueuedConversions($media, $queuedConversions, $onlyMissing)
            ->performConversions($conversions, $media, $onlyMissing);
    }
```

--> queued conversions are dispatched
--> nonQueued conversions are handled

It might happen, that the queued conversions are actually so fast, that they are colliding with the nonQueued conversion.
The `markAsConversionGenerated()` method on the media model reads the generated conversions, adds the new ones and saves to DB. 

Possible Solutions:
* We would need some sort of locking here (https://laravel.com/docs/master/queries#pessimistic-locking).
* Another approach could be to first handle nonQueued conversions and dispatch queued conversions after finishing them.

DaniTulp:
[...]

PR #2235, opened 2020-12-16
Fixed generated conversions race-condition

DaniTulp:
Fixed job overriding so nonQueued conversions get executed before the queued conversions get started. Adresses #2226. 

If I need to add tests I'd love some guidance, I tried to setup a failing test first but I had no luck recreating the behaviour inside a test.


freekmurze:
Good enough for now! Thanks!

PR #2242, opened 2020-12-29
Update handling-uploads-with-vue.md

trevorpan:
Add `!` to line 282 so button is disabled while form is _not ready_ to submit.

Without `!` the behavior disabled the button when ready to submit and **enabled** while _not ready_ to submit.


freekmurze:
Thanks!

trevorpan:
Eventually - I'll get in a **2** character PR!

PR #2277, opened 2021-02-04
[DOCS FIX] Fixed Media Library Pro Blade example

PovilasKorop:
Media Library Pro Blade example in the docs is not working, with error:

```
Argument 1 passed to App\Models\Listing::addFromMediaLibraryRequest() must be of the type array or null, object given
```

I've traced the reason and fixed the docs according to the example app code:
https://github.com/spatie/laravel-medialibrary-pro-app/blob/master/app/Http/Controllers/Blade/BladeAttachmentController.php#L23


freekmurze:
Thanks!

PR #2282, opened 2021-02-06
Allow conversions of any mime type to be rendered

chris-ware:
This allows conversions of any mime type to be rendered, rather than just limiting the rendering of media that is of an image based mime type. It still prevents rendering if the conversion is not available. It means that out of the box, no extra coding is required to render a selection of conversions of any form of file.

### DIFF
--- a/src/MediaCollections/HtmlableMedia.php
+++ b/src/MediaCollections/HtmlableMedia.php
@@ -45,7 +45,7 @@ class HtmlableMedia implements Htmlable
 
     public function toHtml()
     {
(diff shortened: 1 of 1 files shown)


freekmurze:
Can you add a test to make sure that this works correctly?

chris-ware:
I can have a go, but I haven't written tests before, as I haven't had the time to figure them out

freekmurze:
Closing this due to inactivity. Feel free to open up a new PR including tests.

PR #2297, opened 2021-02-17
Replacing an unused InteractsWithMedia trait

barnoma-soz:


freekmurze:
The Interacts one seems to be the right one
https://github.com/spatie/laravel-medialibrary/blob/master/src/InteractsWithMedia.php

PR #2302, opened 2021-02-23
"ignore" -> ".gitignore"

drbyte:
Upon first read I was confused by "Don't forget to ignore", until I realized it was referring to git. This PR simply aims to remove that ambiguity.


freekmurze:
Thanks Chris!

PR #2306, opened 2021-03-01
DiskDoesNotExist moved to it's own class

swilla:
In version 8, `diskDoesNotExist` was moved to its own exception class.


freekmurze:
Thanks!

PR #2322, opened 2021-03-08
Add info about setting the default value

squatto:
I just upgraded to v9 and added the `media.generated_conversions` field. The first time I updated an existing `Media` instance I got a JSON-validation constraint error because the newly-added field didn't have a value (it defaulted to an empty string). This PR adds info about and instructions for setting the default value to `[]` to help people avoid running into the same problem


freekmurze:
Thanks!

squatto:
My pleasure!

PR #2335, opened 2021-03-13
change return type of getMedia()

Danger-Mkh:
change return type from Colllection to MediaCollections\Models\Collections\MediaCollection
because InteractsWithMedia trait not the same as HasMedia interface

### DIFF
--- a/src/HasMedia.php
+++ b/src/HasMedia.php
@@ -38,9 +38,9 @@ interface HasMedia
      * @param string         $collectionName
      * @param array|callable $filters
      *
-     * @return \Illuminate\Support\Collection
+     * @return MediaCollections\Models\Collections\MediaCollection
      */
-    public function getMedia(string $collectionName = 'default', $filters = []): Collection;
+    public function getMedia(string $collectionName = 'default', $filters = []): MediaCollections\Models\Collections\MediaCollection;
 
     public function clearMediaCollection(string $collectionName = 'default'): HasMedia;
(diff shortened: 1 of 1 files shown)


freekmurze:
In rare circumstances, where people have override certain methods, narrowing the return type could be a breaking change.

As this change doesn't solve a real problem, I'm going to skip this for now.

Thanks for your work on this.

AlexanderPoellmann:
Hi @freekmurze, I don't follow your reply, but maybe I just misunderstand your point.

I too get an error as the method declaration of 'InteractsWithMedia->getMedia()' is not compatible with 'HasMedia->getMedia()', which in fact is a simple `Collection` on the interface while on the trait a `MediaCollection`.

Thanks in advance for some clarification ;-)

jonnywilliamson:
@freekmurze Does this not flag up with a red error on your phpStorm?

image

jonnywilliamson:
For anyone following. I found the mentioned bug track number here:

https://youtrack.jetbrains.com/issue/WI-53288
https://youtrack.jetbrains.com/issue/WI-57097

This package actually gets a specific mention. Hopefully fixed soon.

Thank you @freekmurze . I wasn't implying that your package was broken! Just curious if your system didn't highlight it like mine!

AlexanderPoellmann:
Thanks @freekmurze, and thanks @jonnywilliamson for the links.

PR #2336, opened 2021-03-14
Allow null in the FileAdder 'order' setter

JimCronqvist:
When disabling the automatic order property increment (media.order_column) that gets set on each insert, null is being passed automatically by the library when you perform `$media->copy(...)`. Causing an exception to be thrown.

The property is already nullable in the definition.

Btw: Thanks for a great library!


freekmurze:
Thank you!

PR #2342, opened 2021-03-18
Pure Flysystem AdapterInterface support for url generator

xtix:
Some file system adapters can implement pure AdapterInterface (without utilitary AbstractAdapter). AdapterInterface doesn't have `getPathPrefix()` method. We should support this case.


freekmurze:
Thanks!

PR #2351, opened 2021-03-24
Check for Lumen app when queueing responsive images.

brainlet-ali:


src/MediaCollections/Models/Observers/MediaObserver.php
```diff
@@ -84,4 +84,10 @@ private function isLaravel7orHigher(): bool
 
         return false;
     }
+
+    private function isLumen(): bool
```
freekmurze, line 88: Make this protected

src/MediaCollections/Models/Observers/MediaObserver.php
```diff
@@ -84,4 +84,10 @@ private function isLaravel7orHigher(): bool
...
+    {
+        return app() instanceof \Laravel\Lumen\Application;
```
freekmurze, line 90: Import the namespace

freekmurze (changes requested):

freekmurze:
Thanks!

Issue #2349, opened 2021-03-23
singleFile method not work correctly


HamidNE:
I have an album model with `FTP` file system to store media images, in order to add `singleFile` method to have a media for the model, I have this error.
It works fine if I do not use the `singleFile` method.

- laravel/framework: `v8.34.0`
- spatie/laravel-medialibrary: `v9.4.4`

### Model
```php
class Album extends Model implements HasMedia, Auditable {

   ...

    public function registerMediaCollections(): void
    {
        $this
            ->addMediaCollection('image')
            ->singleFile();
    }

    ...

}
```
### Controller
```php
class AlbumImageController extends Controller
{
    /**
     * Handle the incoming request.
     *
     * @param AlbumImageRequest $request
     * @param Album $album
     * @return \Illuminate\Http\JsonResponse
     * @throws \Spatie\MediaLibrary\MediaCollections\Exceptions\FileDoesNotExist
     * @throws \Spatie\MediaLibrary\MediaCollections\Exceptions\FileIsTooBig
     */
    public function __invoke(AlbumImageRequest $request, Album $album)
    {
        $album
            ->addMediaFromRequest('image')
            ->toMediaCollection('image');

        return response()->json([
            'data'    => $album->getFirstMediaUrl('image'),
            'message' => Lang::get('messages.album.image')
        ], 200);
    }
}
```
### Error
```json
{
    "message": "The file  cannot be opened.",
    "exception": "RuntimeException",
    "file": "/var/www/vendor/nyholm/psr7/src/Factory/Psr17Factory.php",
    "line": 46,
    "trace": [
        {
            "file": "/var/www/vendor/symfony/psr-http-message-bridge/Factory/PsrHttpFactory.php",
            "line": 114,
            "function": "createStreamFromFile",
            "class": "Nyholm\\Psr7\\Factory\\Psr17Factory",
            "type": "->"
        },
        {
            "file": "/var/www/vendor/symfony/psr-http-message-bridge/Factory/PsrHttpFactory.php",
            "line": 96,
            "function": "createUploadedFile",
            "class": "Symfony\\Bridge\\PsrHttpMessage\\Factory\\PsrHttpFactory",
            "type": "->"
        },
        {
            "file": "/var/www/vendor/symfony/psr-http-message-bridge/Factory/PsrHttpFactory.php",
            "line": 68,
            "function": "getFiles",
[...]

PR #2353, opened 2021-03-24
Check existence of a directory before removing it

HamidNE:
I have a collection of directories and I want to remove them recursively but in a special case of mine, maybe one of the directories contains the others as a nested directory and any effort to remove a child or nested directory will result in an error, so we have to check the existence of every directory before removing it to avoid error and crashe.

```
Illuminate\Support\Collection {#1870
  #items: array:3 [
    0 => "albums/774/"
    1 => "albums/774/conversions/"
    2 => "albums/774/responsive-images/"
  ]
}
```

related issue: #2349


freekmurze:
Thanks!

PR #2363, opened 2021-03-29
addMediaFromStream

Ahmed-Aboud:
some times the url is protected by auth and the return is 
`http-message/StreamInterface`
or its a return of `ZipArchive::getStream` which is resource

this pull request adds the ability to call `addMediaFromStream` and pass any type of stream
and behaving like `addMediaFromString` for file naming


freekmurze (approved):

freekmurze:
Thank you!

PR #2369, opened 2021-04-03
Return Fallback url or path when conversion is not ready yet

danyelkeddah:
current behavior when you try to access conversion using `getFirstMediaUrl` it will return correct path but the image may not be ready yet or out of the queue, this PR will check if the media `hasGeneratedConversion` before returning the path or URL and if connversion is not ready yet, it will return `FallbackMediaUrl` or `FallbackMediaPath`.

### DIFF
--- a/src/InteractsWithMedia.php
+++ b/src/InteractsWithMedia.php
@@ -312,6 +312,10 @@ trait InteractsWithMedia
             return $this->getFallbackMediaUrl($collectionName) ?: '';
         }
 
+        if ($conversionName !== '' && !$media->hasGeneratedConversion($conversionName)){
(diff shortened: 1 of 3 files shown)


danyelkeddah:
Related to #1897

freekmurze:
Thanks! Could you add a test for this?

danyelkeddah:
> Thanks! Could you add a test for this?

Tests are ready. 👍🏻

danyelkeddah:
> Thanks! Could you add a test for this?

@freekmurze Tests are ready. 👍🏻

freekmurze:
Thanks!

PR #2377, opened 2021-04-08
Fix #2372 the fallback should be the original image

### LINKED ISSUE #2372: Version 9.5.1 return always the FallbackUrl
After updating to version 9.5.1 the links to images converted to thumbail (`->getFirstMediaUrl('posts', 'thumb')`) no longer work. By downgrading to version 9.5.0 everything works normally again.

Tofandel:
#2369 Introduced some undesired behavior, it would be better to return the original image and not the fallback one until the conversions are generated

Partially fixes #2372


freekmurze:
Merging this in, as many users seem to be hindered by the changes made in #2369

PR #2382, opened 2021-04-10
Fixes adding Private Media from same S3 Disk

### LINKED ISSUE #2203: Media is not uploaded with proper custom headers on S3
When uploading a media that has custom headers for S3, if the remote file which is used to upload media is on the same disk then the file is copied directly. During this process, the custom headers expected in the new file are not inserted.

This is specifically happening in v8 of Media Library as the copy logic was not there till v7.
Check copyToMediaLibraryFromRemote() function in file src/MediaCollections/Filesystem.php line 52 of v8.

E.g.
1.Temp file is uploaded with private ACL.
2. Now when that file is used to upload to the new public location public ACL is provided through medialibrary.php config file.
3. Now maybe to improve performance, if both the files are on the same disk, they are just copied over and not re-retrieved and uploaded.
4. As the file is just copied over the extra custom headers are never applied.

ebbbang:
Fixes #2203

### DIFF
--- a/src/MediaCollections/Filesystem.php
+++ b/src/MediaCollections/Filesystem.php
@@ -49,17 +49,17 @@ class Filesystem
 
         $destination = $this->getMediaDirectory($media, $type).$destinationFileName;
 
-        if ($file->getDisk() === $media->disk) {
+        $diskDriverName = (in_array($type, ['conversions', 'responsiveImages']))
+            ? $media->getConversionsDiskDriverName()
+            : $media->getDiskDriverName();
+
+        if ($this->shouldCopyFileOnDisk($file, $media, $diskDriverName)) {
             $this->copyFileOnDisk($file->getKey(), $destination, $media->disk);
 
             return;
         }
 
         $storage = Storage::disk($file->getDisk());
-
-        $diskDriverName = (in_array($type, ['conversions', 'responsiveImages']))
-            ? $media->getConversionsDiskDriverName()
-            : $media->getDiskDriverName();
 
         $headers = $diskDriverName === 'local'
             ? []
@@ -75,6 +75,23 @@ class Filesystem
             $media->disk,
             $headers
         );
+    }
+
+    protected function shouldCopyFileOnDisk(RemoteFile $file, Media $media, string $diskDriverName): bool
+    {
+        if ($file->getDisk() !== $media->disk) {
+            return false;
+        }
+
+        if ($diskDriverName === 'local') {
+            return true;
+        }
+
+        if (count($media->getCustomHeaders()) > 0) {
+            return false;
+        }
+
+        return true;
     }
 
     protected function copyFileOnDisk(string $file, string $destination, string $disk): void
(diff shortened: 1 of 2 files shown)


src/MediaCollections/Filesystem.php
```diff
@@ -77,6 +77,11 @@ public function copyToMediaLibraryFromRemote(RemoteFile $file, Media $media, ?st
         );
     }
 
+    protected function shouldCopyFileOnDisk(RemoteFile $file, Media $media, string $diskDriverName): bool
+    {
+        return $file->getDisk() === $media->disk && ($diskDriverName !== 's3' || count($media->getCustomHeaders()) === 0);
```
freekmurze, line 82: This kind of conditional is unreadable to me. Could you split this up and use early returns?

freekmurze (changes requested):

src/MediaCollections/Filesystem.php
ebbbang: Done, please let me know if that helps. If not I'll be more than happy to improve it further.

For your reference, the conditions I want to meet are:
1. It only affects if the disks are the same.
2. It only affects S3 as far as I know.
3. And only when custom headers are involved which were not present on the source file, to begin with.

src/MediaCollections/Filesystem.php
```diff
@@ -77,6 +77,23 @@ public function copyToMediaLibraryFromRemote(RemoteFile $file, Media $media, ?st
         );
     }
 
+    protected function shouldCopyFileOnDisk(RemoteFile $file, Media $media, string $diskDriverName): bool
+    {
+        if ($file->getDisk() !== $media->disk) {
+            return false;
+        }
+
+        if ($diskDriverName !== 's3') {
```
freekmurze, line 86: Checking "s3" here feels a bit strange. Wouldn't it be better to check for `=== 'local'` here?

src/MediaCollections/Filesystem.php
freekmurze: Thanks!

src/MediaCollections/Filesystem.php
ebbbang: Yes, we could use ```=== 'local'``` here, but I did not know how that would affect other filesystem drivers performance-wise.

I assumed that copy would be most efficient for most drivers and even for 's3' if no custom headers are provided.

If you want or if I'm missing something then let me know and I can change it to ```=== 'local'```.

ebbbang: @freekmurze Please do let me know what you think?

freekmurze: Let's go with `=== 'local'` 👍

ebbbang: Done. I also squashed the commits in the process to keep it clean. Please let me know if more work is needed or can we merge it now?

freekmurze:
Thanks!

PR #2384, opened 2021-04-11
Alter the migration for upgrading to v9 to copy the custom properties properly

MaxKorlaar:
The suggested migration in the current docs didn't work for me, however by altering it I've got it to work now.

The migration now copies the content of the JSON field `generated_conversions` from the column `custom_properties` to the new column `generated_conversions`. You can optionally remove this field from `custom_properties` as well.

I've also added a `orWhereRaw` check because my MySQL didn't seem to think that a JSON field with the value `null` equals to a MySQL field's native `NULL`.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -10,12 +10,11 @@ Because there are many breaking changes an upgrade is not that easy. There are m
 ```php
 <?php
 
-<?php
-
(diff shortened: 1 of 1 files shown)


freekmurze:
thanks

PR #2389, opened 2021-04-12
Add `Collection` typehint to `addMultipleMediaFromRequest`

### LINKED ISSUE #2388: Bad type annotations on `addMultipleMediaFromRequest`
Hey, thanks for this library, super helpful!

I've noticed an issue with the `addMultipleMediaFromRequest` method on `InteractsWithMedia`, it returns a Collection but the PHPDoc specifies the return as an array of `FileAdder`, which makes sense, but trying to chain collection methods on to the end of that call causes intellisense to complain that it's not valid.

image

I've noticed the `addAllMediaFromRequest` method has the same PHPDoc annotation, but also has a PHP typehint, and that seems to be fine, so I think the solution is to just add the typehint to the method?

jakewtaylor:
Resolves #2388 

I've added the `Collection` type hint which makes intellisense happy when chaining methods

image

### DIFF
--- a/src/InteractsWithMedia.php
+++ b/src/InteractsWithMedia.php
@@ -120,7 +120,7 @@ trait InteractsWithMedia
      *
      * @return \Spatie\MediaLibrary\MediaCollections\FileAdder[]
      */
-    public function addMultipleMediaFromRequest(array $keys)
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks!

PR #2395, opened 2021-04-13
Don't decode base64 data thrice

abenerd:


src/InteractsWithMedia.php
```diff
@@ -215,17 +215,15 @@ public function addMediaFromBase64(string $base64data, ...$allowedMimeTypes): Fi
         }
 
         // strict mode filters for non-base64 alphabet characters
-        if (base64_decode($base64data, true) === false) {
+        if (false === ($binaryData = base64_decode($base64data, true))) {
```
freekmurze, line 218: This is next to unreadable for me. Let's just store the result in a dedicated variable on it's own line and use that in the checks.

abenerd: Ah yes, I was unsure whether I should keep this inlined or not.
Fixed it now, so that the assignment is its own declaration.

freekmurze:
Thanks!

PR #2403, opened 2021-04-19
Include extra_headers when adding media from same S3 Disk

ebbbang:
Bug: Extra Headers were not being applied when media was copied from the same S3 Disk.

Solution: Do not copy media from the same disk if ```extra_headers``` is provided.


freekmurze:
Thanks!

PR #2415, opened 2021-05-01
Enable Storage::fake() usage

edwardsteven:
I tried to use this library with Storage::fake() and found that I couldn't because the FileAdder is looking in the configuration to see if the disk is configured. For developer ease I have changed that check to actually attempt to get the FileSystem with Storage::disk, because file-configured, test-configured and Storage::fake disks will all work, and disks that don't exist will throw an Exception, which gets turned into a DiskDoesNotExist exception.

### DIFF
--- a/src/MediaCollections/FileAdder.php
+++ b/src/MediaCollections/FileAdder.php
@@ -373,9 +373,14 @@ class FileAdder
 
     protected function ensureDiskExists(string $diskName)
     {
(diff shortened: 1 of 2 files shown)


glamax13:
I would also be interested in this.
Is there any reason not to merge this?
Doesn't look like it's too invasive and it would help me test the package i'm working on...

I was actually going to make a PR for that but i found this issue.

PR #2418, opened 2021-05-04
Add a way to rename files globally

freekmurze:
This PR adds a method `originalFileName` on a `FileNamer`. that allow files to be renamed globally.

The method should return the desired file name without an extension.

In the config file you can specify which `FileNamer` class should be used.


benholmen:
Thank you @freekmurze!

PR #2427, opened 2021-05-06
Customize Filename in Zip using `zip_filename` custom property

ibrahim-mubarak:
This PR aims to allow setting custom filenames for media in zip files.
This is required in use cases where `zip_filename_prefix` is not sufficient.

- [x] Passed existing tests.
- [x] Added new tests.
- [x] Updated documentation.

The behavior right now is that filename will be in the following format
`zip_filename_prefix` `zip_filename`

`zip_filename` falls back to `$mediaItem->file_name` if the property does not exist
The file extension is not preserved and has to be set in `zip_filename`

### DIFF
--- a/docs/advanced-usage/using-custom-properties.md
+++ b/docs/advanced-usage/using-custom-properties.md
@@ -75,7 +75,8 @@ title: Special custom properties
 weight: 2
 ---
 
-## ZIP File Folders
+## File Structure customization in zip files
+### Folders and Subfolders
 
 The ZIP export stores all media files in the root folder of the ZIP file.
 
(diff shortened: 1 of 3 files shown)


ibrahim-mubarak:
There are certain questions that I have regarding this.

- Should extensions be preserved?
  -  If yes, should that be the default behavior?
  -  If yes, should that be achieved using a custom property such as `zip_preserve_extension` or `zip_skip_extension`
- Do we simplify `getZipFileNamePrefix`. This will be a breaking change with not much benefit.
  - Is there a way to gracefully deprecate custom properties?

freekmurze:
Thanks for you work on this!

I'm going to pass on this for the moment. I think most people don't need this, and I want to keep the package as simple as it is.

PR #2434, opened 2021-05-07
Added array type hint to $casts to match Laravel

kirkaracha:
In the latest versions of Laravel src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php the $casts variable has an array type hint. (Tested in 8.39.0)

When postAutoloadDump runs after installing spatie/laravel-medialibrary there is a Symfony\Component\ErrorHandler\Error\FatalError: 
Type of Spatie\MediaLibrary\MediaCollections\Models\Media::$casts must be array (as in class Illuminate\Database\Eloquent\Model)

Adding this type hint fixes the error.

### DIFF
--- a/src/MediaCollections/Models/Media.php
+++ b/src/MediaCollections/Models/Media.php
@@ -40,7 +40,7 @@ class Media extends Model implements Responsable, Htmlable
 
     protected $guarded = [];
 
-    protected $casts = [
+    protected array $casts = [
         'manipulations' => 'array',
         'custom_properties' => 'array',
         'generated_conversions' => 'array',


freekmurze:
This PR seems to break the tests, could you take a look at that?

src/MediaCollections/Models/Media.php
```diff
@@ -40,7 +40,7 @@ class Media extends Model implements Responsable, Htmlable
 
     protected $guarded = [];
 
-    protected $casts = [
+    protected array $casts = [
```
SimBig, line 43: as explained here you are not allowed to cast this property this way: https://stackoverflow.com/a/61302038

if you need to cast it for eg. php stan you could add a php doc block
```
    /**
     * @var array<string, string|class-string<\Illuminate\Contracts\Database\Eloquent\Castable>>
     */
     protected $casts = [
```

PR #2438, opened 2021-05-11
Added example migration

LowSociety:
Added example migration for new default collection name for responsive images.


freekmurze:
Thanks!

PR #2440, opened 2021-05-13
Inline setting the filname when adding media

abenerd:
This PR allows setting a custom filename when adding new Media from string or from a stream.
It was already possible to override the default filename set but with this change, the default can be overriden.
I don't consider the signature change a breaking change, as the argument has been added to the end and has a default value, so even people using PHP 8 named arguments should not have any problems.


freekmurze:
Thanks for your work on this. I'm thinking the extra `usingFileName` call people have to do to name the file is convenient enough.

PR #2447, opened 2021-05-21
[v7] PHP8 Support

erikn69:
Testing v7 on Php 8
```
PHPUnit 9.5.4 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.0.5
```

On `P7.4 - L8.* - prefer-lowest` is the only error, on my localhost pass all test, here is adding a letter "S" to table's name
`Illuminate\Database\QueryException: SQLSTATE[HY000]: General error: 1 no such table: medias (SQL: select max("order_column") as aggregate from "medias")`


freekmurze:
Feel free to require higher version to let all tests pass.

erikn69:
@freekmurze inflector dependency to 1.4.1 minimum version

AlexVanderbist:
Thanks! Will tag this as a new feature version on v7

PR #2449, opened 2021-05-24
Additional condition for upgrade migration added

4unkur:
The old code generates an error when trying to run the migration on the media table with empty `custom_properties` column (i.e. `[]`)


freekmurze:
Thanks!

PR #2453, opened 2021-05-27
Add getLastMediaUrl methods

mokhosh:
We have had the great `getFirstMedia` methods which are very helpful. Sometimes we need to get the newest media in a gallery of media. So I added them.


freekmurze:
I'm going to pass on this one, in order to keep the package the smallest it can be.

Thanks for your work on this.

PR #2463, opened 2021-06-06
Fix method name in preview manipulation example

BrianLangevin:
The function `TemporaryUpload::manipulatePreview` does not exist. It should be `TemporaryUpload::previewManipulation`.


freekmurze:
Thanks!

PR #2475, opened 2021-06-18
Add getPathRelativeToRoot method to Media model

jandreasn:
This PR adds a `getPathRelativeToRoot()` method to the Media model, which for example enables test assertions like this:

```
Storage::fake('avatars');

$this->patch('/avatar', [
    'avatar' => UploadedFile::fake()->image('avatar1.jpg'),
]); 

$this->assertNotNull($media = $user->getFirstMedia('avatars'));
Storage::disk('avatars')->assertExists($media->getPathRelativeToRoot());
```

Since this changes the UrlGenerator interface it will have to be targeted for the next major version. I will add tests and upgrade instructions if you are willing to accept this addition.

### DIFF
--- a/src/MediaCollections/Models/Media.php
+++ b/src/MediaCollections/Models/Media.php
@@ -83,6 +83,13 @@ class Media extends Model implements Responsable, Htmlable
         return $urlGenerator->getPath();
     }
 
+    public function getPathRelativeToRoot(string $conversionName = ''): string
+    {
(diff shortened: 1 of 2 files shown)


freekmurze:
Thanks you for this.

Closing for now, but I've added a label to remind me to take a look at this and merge this in when creating the next major version of the medialib.

PR #2486, opened 2021-06-23
Update CleanCommand.php

prullenbak:
Add the option to skip 'deprecated conversions' when cleaning the media directory.

I asked a question about this, but I cant find it anymore...strange. Anyway, we add some custom conversion files after media-library has done it's job. When we then run this command, media-library identifies these files as deprecated conversions, but they aren't. This is not a bug, it's quite logical that it works this way. But with this option, there is a way to leave these files intact and still remove all the orphaned directories. It should help us, and maybe others to have this option.

I would also update tests, but I couldn't find any. If there is more to be done to make this work, please let me know!


freekmurze:
Thanks!

huubk:
This update resulted in an error while updating the package in my project. See: #2489

PR #2490, opened 2021-06-24
Include `url` to the `disks` config

3ub41r:
This config example, in particular, caused some confusion when I installed laravel-medialibrary for the first time. Thanks for all the great contributions! 🥰


freekmurze:
Thank you!

PR #2527, opened 2021-07-28
Allow keepOriginalImageFormat() method to work with uppercase file extensions

intrepidws:
Currently, the keepOriginalImageFormat() method will not work with a file that has an uppercase file extension. Instead it simply defaults to a JPG. This PR fixes that.


freekmurze:
Very nice catch! Thank you!

PR #2529, opened 2021-07-30
Add extra comment to max_file_size

lakuapik:
I dont know if this will be useful to others or not, but for me it is useful to make me think once instead of thinking twice by calculating those numbers and reading full comment.

feel free to accept or drop this PR :v: :smiley: .


freekmurze:
Thanks!

Issue #2534, opened 2021-08-06
SQLSTATE[42S01]: Base table or view already exists: 1050 Table 'media' already exists


bhavingajjar:
php artisan migrate:refresh
when I run after install media library getting "SQLSTATE[42S01]: Base table or view already exists: 1050 Table 'media' already exists" this error

bhavingajjar:
https://github.com/spatie/laravel-medialibrary/pull/2535 fixed by this solution

erikn69:
You have to implement `down` method on your aplication, not on the package

PR #2535, opened 2021-08-06
#2534 Base table or view already exists: 1050 Table 'media' already e…

bhavingajjar:
…migration exists issue on migration refresh fixed


freekmurze:
Thanks, passing on this, as we don't use down migrations in our own projects.

PR #2537, opened 2021-08-09
Update Upgrading.md from v7 to v8

juliangeissler:
Added additional info for upgrading the models from v7 to v8


freekmurze:
Thanks!

PR #2557, opened 2021-08-22
Add custom header support for temporary uploads

mbardelmeijer:
Added to support https://github.com/spatie/laravel-medialibrary-pro/pull/300


freekmurze:
Thank you!

PR #2575, opened 2021-09-03
[V9] Make sure only the extension is removed on DefaultFileNamer

erikn69:
According docs we can use a custom filename with `usingFileName` and a custom sanitizer with `sanitizingFileName`, but when you add a name with slash `FileNamer` cuts that part beacuse `pathinfo($fileName, PATHINFO_FILENAME)` remove all and let only file name, this PR fix that without change old behaivor

```php
$yourModel
   ->addMedia($pathToFile)
   ->usingFileName('other/otherFileName.jpg')
      return strtolower(str_replace(['#',  '\\', ' '], '-', $fileName));
   })
   ->toMediaCollection();
```
Test added

### DIFF
--- a/src/Support/FileNamer/FileNamer.php
+++ b/src/Support/FileNamer/FileNamer.php
@@ -9,7 +9,9 @@ abstract class FileNamer
 {
     public function originalFileName(string $fileName): string
     {
-        return pathinfo($fileName, PATHINFO_FILENAME);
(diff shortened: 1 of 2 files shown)


freekmurze:
Thanks!

francoism90:
@erikn69 Could you please check #2586? It seems to cause some issues on importing files.

Thanks. :)

PR #2576, opened 2021-09-04
Add simpler way to prefix the root directory where media are saved

kevinpurwito:
Currently, if we use S3 bucket for our media, the `DefaultPathGenerator` will save all our media in the root directory of the S3 bucket. 

For example:
```
your-bucket/
___1/
______filename.png
___2/
______filename.png
```

If we want to add a prefix/root directory where we want to save all our media, we need to implement our own `PathGenerator`.

For example:
```
your-bucket/
___media/
______1/
_________filename.png
______2/
_________filename.png
```

This PR made it simpler (IMO) to add this prefix, by simply using a new env `MEDIA_PREFIX`, which defaults to empty so that it doesn't change any existing behavior at all. 

This way, the general users who want to add their own root directory to store the media - such as myself, can easily do so without implementing our own `PathGenerator`.


freekmurze:
Seems nice! Thanks! 

Could you add a test that proves that this works?

kevinpurwito:
OK, I'll add a test later today or tomorrow.

kevinpurwito:
Hi, I've added the tests to make sure it works properly. 

Also, since I'm developing in Windows, I took the liberty to make the tests in `GetPathTest.php` to be Windows-friendly as well (I hope you don't mind).

image

freekmurze (changes requested):

kevinpurwito:
Noted. I'll update later as you suggested.

kevinpurwito:
Hello, I've made the changes you suggested. Thank you for your kind review.

tests/TestCase.php
```diff
@@ -316,4 +316,13 @@ protected function fileExistsInZipRecognizeFolder($zipPath, $filename): bool
 
         return false;
     }
+
+    protected function makePathOsSafe($path)
```
freekmurze, line 320: Add parameter type hint + return type hint here

.gitignore
```diff
@@ -7,3 +7,4 @@ phpunit.xml
 .env
 .phpunit.result.cache
 .php-cs-fixer.cache
+.idea
```
freekmurze, line 10: Remove this, it belongs in a global gitignore

config/media-library.php
```diff
@@ -8,6 +8,12 @@
      */
     'disk_name' => env('MEDIA_DISK', 'public'),
 
+    /*
```
freekmurze, line 11: Move this option to the bottom of the config file.

freekmurze (changes requested):
Looks good, just a few nitpicks left

kevinpurwito:
> Looks good, just a few nitpicks left

Hi, I've made the changes you requested

freekmurze:
Thanks!

PR #2580, opened 2021-09-09
Include --force param to image_optimzer . jpegoptim

ibrainventures:
I wonder why 20% of my conversions are not progressive. 
Jpegoptim is handling this by the --force param.
If the resulting image is only 1 Bit bigger than it is non-progressive
saved.


freekmurze:
Thanks!

PR #2583, opened 2021-09-14
Add down method in migration stub

luilliarcec:
I don't know if they forgot to add it or it was something intentional. However, there I leave my small contribution.


erikn69:
It was intentional, spatie packages normally don't have a down method

PR #2605, opened 2021-10-01
Feature/add media deleted event

### LINKED ISSUE #2574: Adding a MediaDeleted event
It would be great to send an event when a particular media entry is deleted similar to MediaAdded event

Jhnbrn90:
## Intro
As proposed in #2574 this PR adds support for dispatching a `MediaHasBeenDeleted` event, whenever a `Media` entity is deleted.

## What has been done
- The `MediaHasBeenDeleted` event was added.
- The event is dispatched from the `removeAllFiles()` call in `FileSystem`: 
  Since the `MediaObserver` model observer hooks into the `deleted()` lifecycle hook of the specified `Media` model, following the calls eventually lead to a call to the `removeAllFiles()` method on the `Filesystem` class. From here, the `MediaHasBeenDeleted` event is dispatched.
- Updated the `EventTest` testcase:
  - Extracted common logic to a `addMediaTo()` method
  - Since the `MediaHasBeenDeleted` event is only fired after the model event "deleted" has been fired, the test exclusively fakes the `MediaHasBeenDeleted` event.
- Updated the "Consuming Events" section in the documentation:
  - Fixed a typo in the "MediaHasBeenAdded" section
  - Added section for "MediaHasBeenDeleted" event 

closes #2574.

### DIFF
--- a/docs/advanced-usage/consuming-events.md
+++ b/docs/advanced-usage/consuming-events.md
@@ -6,7 +6,7 @@ weight: 8
 The media library will fire the following events that your handlers can listen for:
 
 ### MediaHasBeenAdded
-This event is fired after the a file has been saved to disk.
+This event is fired after a file has been saved to disk.
 
 The event has a property `media` that holds the `\Spatie\MediaLibrary\Models\Media`-object of which the file has been stored.
 
--- a/docs/advanced-usage/consuming-events.md
+++ b/docs/advanced-usage/consuming-events.md
@@ -10,6 +10,11 @@ This event is fired after a file has been saved to disk.
 
 The event has a property `media` that holds the `\Spatie\MediaLibrary\Models\Media`-object of which the file has been stored.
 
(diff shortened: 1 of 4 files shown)


freekmurze:
Thanks for this. 

Right now, a model deleted event is fired after a media model is deleted. What is the use case for having an event after all files have been deleted?

Jhnbrn90:
Hey @freekmurze , 

In all honesty, I didn't and can't really think of a good use case for this additional event. I was browsing some issues in response to the Hacktoberfest mail Spatie sent out and figured I'd PR an implementation of this feature request (without thinking of an actual use case 🤔). 

Maybe @sicsol (who opened the issue), can you help us explain your use case for a separate event here?

Otherwise, feel free to close this PR of course.

freekmurze:
Thanks for the work, but passing on it now in the interest of keeping the package as simple as it can be.

PR #2609, opened 2021-10-06
Fix missing image preview URLs in Inertia Projects

dustinboling:
When using Inertia.js with Laravel, preview_url is not being populated causing missing thumbnails in the Media Library Pro Vue components. This is because the Inertia::render() method checks if a prop is arrayable. If so, it calls the toArray method. This PR adds the toArray method for proper mapping of media objects when used with Inertia.js.


freekmurze:
This PR has been reverted in 9.8.2

PR #2636, opened 2021-10-20
Add 'toInlineResponse' to Media

mbardelmeijer:
Follow-up of #1358

### DIFF
--- a/docs/advanced-usage/outputting-media.md
+++ b/docs/advanced-usage/outputting-media.md
@@ -1,0 +1,34 @@
+---
+title: Outputting media
+weight: 5
+---
+
+If you want to output your `$media` instance to the browser, you may use the `toResponse` & `toInlineResponse` methods.
+
+```bash
+use Illuminate\Http\Request;
+use Spatie\MediaLibrary\MediaCollections\Models\Media;
+
+public function download(Request $request, Media $media)
+{
+    return $media->toResponse($request);
+}
+```
+
+By using `toResponse`, your browser is instructed to download the file with the `attachment` Content-Disposition.
+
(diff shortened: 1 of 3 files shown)


rcerljenko:
@mbardelmeijer @freekmurze  would be great if you can optionally pass a `conversion` name to these functions. eg:
```php
return $media->toResponse($request, 'thumb');
```

and how about responsive images?

docs/advanced-usage/outputting-media.md
```diff
@@ -0,0 +1,34 @@
+---
+title: Outputting media
+weight: 5
+---
+
+If you want to output your `$media` instance to the browser, you may use the `toResponse` & `toInlineResponse` methods.
+
+```bash
```
freekmurze, line 8: This should be ```php I guess

freekmurze (changes requested):

freekmurze:
> would be great if you can optionally pass a conversion name to these function

Good idea!, could you take care of this @mbardelmeijer 

Responsive images are out of scope for now.

mbardelmeijer:
@freekmurze @rcerljenko I've checked if it would be possible to add conversion support to these methods, but ran into that the `stream` method from the `Filesystem` doesn't support passing a conversion name. 

We can get the file via the default methods, but then we don't have the memory optimized stream function to use.

We can add a 'getConversionStream' method for instance in the `Filesystem`, and a `conversionStream` to the Media object, but I do think this would grow the scope of the PR too much. Thoughts?

rcerljenko:
@mbardelmeijer yes it adds more work.. let @freekmurze decides

freekmurze:
Let's keep it simple for now.

Thanks @mbardelmeijer

PR #2652, opened 2021-10-29
Add down method in migration Schema::dropIfExists('media');

saber13812002:
Add down method in migration Schema::dropIfExists('media'); when php artisan migrate:rollback run in laravel project that medialibrary installed


freekmurze:
We don't use down migrations in our projects, see other duplicate PRs.

PR #2670, opened 2021-11-11
Reinsert down()

danidou:
The down() method in the project's media table migration stub has been dropped at some point, which breaks rollbacks.  I'm simply proposing to add it back.


freekmurze:
We don't use `down` in our packages/project, see previous PRs.

PR #2681, opened 2021-12-01
Add option to disable loading of dragula via cdn in medialibrary-pro

lukasleitsch:
This PR adds a new option to disable the loading of dragula via a cdn.

See also the PR https://github.com/spatie/laravel-medialibrary-pro/pull/349


freekmurze:
Because not too many people will need this option, I prefer to have it hidden.

PR #2683, opened 2021-12-02
Check for lower and uppercase image extensions but always return the original extension for conversions

timvanuum:
Uploading images with capital JPG didn't work. 
This seems to fix the issue.

Originally fixed in: #2527


freekmurze:
Could you add a test that would break without this code?

freekmurze:
Thanks!

PR #2688, opened 2021-12-08
Add starting-from-id option to regenerate command

billypoke:
I have found myself wanting to regenerate recent images, but with potentially thousands of ids to pass as a command option, I thought adding a date option could be useful

### DIFF
--- a/src/Conversions/Commands/RegenerateCommand.php
+++ b/src/Conversions/Commands/RegenerateCommand.php
@@ -18,7 +18,7 @@ class RegenerateCommand extends Command
 
     protected $signature = 'media-library:regenerate {modelType?} {--ids=*}
     {--only=* : Regenerate specific conversions}
-    {--after-date= : Regenerate media after a certain date}
+    {--starting-from-id= : Regenerate media with an id equal to or higher than the provided value}
     {--only-missing : Regenerate only missing conversions}
     {--with-responsive-images : Regenerate responsive images}
     {--force : Force the operation to run when in production}';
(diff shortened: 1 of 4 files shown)


freekmurze:
Could you refactor this option to `starting-from-id`. This way you still get the behaviour your want, but it's slight more clear and a little bit more flexible.

freekmurze:
Could you also update the readme with an example?

billypoke:
@freekmurze refactored and documented in the readme. Let me know if you have any other changes

Thanks for taking the time 😄

billypoke:
looks like failing tests now. Seems to be failing on `main` as well

src/Conversions/Commands/RegenerateCommand.php
```diff
@@ -75,16 +76,25 @@ public function handle(MediaRepository $mediaRepository, FileManipulator $fileMa
     public function getMediaToBeRegenerated(): Collection
     {
         $modelType = $this->argument('modelType') ?? '';
+
         $mediaIds = $this->getMediaIds();
+        $noIds = count($mediaIds) === 0;
```
freekmurze, line 81: This whole bit feel unreadable to me. 

Try to start with the special cases (so starting with determining if you should start from an id or model.

Instead of using `&&`, introduce new helper functions where you don't use `&&` but work with early returns.

Also make it possible to get models that start with an id greater than some id.

src/Conversions/Commands/RegenerateCommand.php
```diff
@@ -75,17 +77,22 @@ public function handle(MediaRepository $mediaRepository, FileManipulator $fileMa
...
 
-        if (! count($mediaIds)) {
-            return $this->mediaRepository->getByModelType($modelType);
+        $mediaIds = $this->getMediaIds();
+        if (count($mediaIds) > 0) {
+            return $this->mediaRepository->getByIds($mediaIds);
         }
 
-        return $this->mediaRepository->getByIds($mediaIds);
+        return $this->mediaRepository->all();
```
billypoke, line 95: I think this is what you were suggesting? Handle the special cases first, then fall back to ids or all media

freekmurze: This seems better now.

A case that isn't handled well now is we you pass both `modelType` and `startingFromId`. Right now that `startingFromId` will be ignored.

billypoke: This should work without refactoring the existing methods to pass the builder instead of the result and chaining everything depending on the args and options. 

I didn't think passing `ids` and `modelType` at the same time would make a ton of sense as a use case as the ids may not belong to the passed type.

freekmurze:
Thanks!

PR #2694, opened 2021-12-13
Fix for deleted files on copy media to model

jon-sm:
Greetings!

We have been losing files since upgrading to 9.10.0 when we copy media items to models, apparently introduced in #2638.

It's my first contribution to spatie, any feedback is really appreciated.


freekmurze:
Thanks!

PR #2701, opened 2021-12-15
Update .gitattributes

PaolaRuby:
Make the package **4.66 MB** (4,886,528 bytes) lighter on vendor

https://github.com/spatie/spatie.be/blob/4d8d699c1230888af4ded98814733033b765aa95/app/Console/Commands/ImportDocsFromRepositoriesCommand.php#L119-L120
With `git pull origin ${branch} && cp -r docs/* ../../../docs/{$repository['name']}/{$alias}` it would copy the docs despite the `export-ignore`, right?

I did test it, i get the docs despite the `export-ignore`

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,11 +3,16 @@
 # https://www.kernel.org/pub/software/scm/git/docs/gitattributes.html
 
 # Ignore all test and documentation with "export-ignore".
+/.github            export-ignore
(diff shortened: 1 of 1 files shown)


freekmurze:
Nice! And thank you for being so thoughtful checking the code at spatie.be

PR #2704, opened 2021-12-17
add index to order_column migration

petsoukos:
This PR adds an index to the order_column migration.

It significantly speeds up adding media with large media tables (>10m entries)

### DIFF
--- a/database/migrations/create_media_table.php.stub
+++ b/database/migrations/create_media_table.php.stub
@@ -24,7 +24,7 @@ class CreateMediaTable extends Migration
             $table->json('custom_properties');
             $table->json('generated_conversions');
             $table->json('responsive_images');
-            $table->unsignedInteger('order_column')->nullable();
+            $table->unsignedInteger('order_column')->nullable()->index();
 
             $table->nullableTimestamps();
         });


PaolaRuby:
Closes #2705

freekmurze:
Thanks!

mikemand:
For those of us that have already run the migration (in my case several years ago), you can run this SQL command to do the same thing:

```sql
ALTER TABLE `media` ADD INDEX `media_order_column_index` (`order_column`);
```

`media_order_column_index` is the name of the index, and I mimicked Laravel's way of naming indexes as best as I could; feel free to adjust as you prefer.

Or, you could create a new migration to alter the table that way, but I went ahead and just did it through SQL because I have long since [cleaned up my migrations](https://laravel.com/docs/8.x/migrations#squashing-migrations).

I'm only up to 87,000 rows, so nowhere near the 14 million that @petsoukos was having problems with, but every little bit helps. Thank you @petsoukos!

petsoukos:
@mikemand You're welcome. Hopefully, others may not experience slow downs if they get to a ridiculous amount of media entries, like me.

patrickomeara:
For those that want the migration:

```php
    public function up()
    {
        Schema::table('media', fn (Blueprint $table) => $table->index('order_column');
    }
```

PR #2726, opened 2022-01-11
Fix tests

fabio-ivona:
This PR will fix the "beforeEach already exist" error in pest test suite

Can't test the code, though (have not an S3 for testing)

my only fear is that you couldn't call Pest's `getS3BaseTestDirectory()` directly from `S3TestPathGenerator.php`, but the main issue is that `S3IntegrationTest` class was autoloaded by composer in `S3TestPathGenerator.php` and it looked into `S3IntegrationTest.php` triggering beforeEach() a second time


freekmurze:
Thank you!

PR #2737, opened 2022-01-19
Make fallback url/path have different defaults (in tests)

danielsimkus:
I was looking at the tests surrounding fallbackUrl as part of https://github.com/spatie/laravel-medialibrary/discussions/2734 and just noticed that I could slightly improve them by using a different default image name.

This prevents the tests from passing if we accidentally return the fallbackPath on the fallbackUrl method


freekmurze:
Nice, thanks!

PR #2756, opened 2022-02-03
Fix `jsonSerialize` return type

orkhanahmadov:
Currently `jsonSerialize` returns collection from but method expects array.


freekmurze:
Thanks!

PR #2784, opened 2022-02-16
fix: explicitly load media

jaulz:
Currently, if lazy loading is disabled there is a dump because `media` is eager loaded. This PR fixes it and loads `media` explicitly if it wasn't loaded yet.

It would be great if you could also apply the same for the 9.* release. Thanks!


freekmurze:
Thanks!

freekmurze:
Feel free to send the same PR to the `v9` branch.

LufoX11:
Could you please extend this fix to v9 as well for users with PHP < 8?

PR #2785, opened 2022-02-16
fix: explicitly load media

jaulz:
See https://github.com/spatie/laravel-medialibrary/pull/2784


freekmurze:
Thanks!

jaulz:
Thanks to you 😉

PR #2788, opened 2022-02-18
Add collection template to collections

bastien-phi:
As Laravel 9 provides generics for the collections, this PR adds templates for MediaCollection and ConversionCollection.

PHPStan baseline was also regenerated


freekmurze:
Thanks!

PR #2797, opened 2022-02-21
Tiff support

Synchro:
Further to the discussion in #2794, this PR adds TIFF support to Media Library.

I''m not entirely happy with the implementation, and I think it would be much better to obtain the list of supported file extensions and MIME types from spatie/image rather than maintaining a separate set of them in here. However, what's here is quite workable.

This is the final PR in a chain [starting in glide](https://github.com/thephpleague/glide/pull/345), going [via spatie/image](https://github.com/spatie/image/pull/159) before ending up here, all because I needed to use TIFF in MLP! It's the first time I've had a PR chain that long!


Synchro:
Eep, did I make the PR against the wrong branch?

Synchro:
I'll resubmit this against main, think my upstream clone was out of sync

PR #2799, opened 2022-02-21
Tiff support

Synchro:
Further to the discussion in #2794, this PR adds TIFF support to Media Library. This is my second attempt at this PR, having made it against the wrong branch last time!

I'm not entirely happy with the implementation, and I think it would be much better to obtain the list of supported file extensions and MIME types from spatie/image rather than maintaining a separate set of them in here. However, it's workable for now.

This is the final PR in a chain [starting in glide](https://github.com/thephpleague/glide/pull/345), going [via spatie/image](https://github.com/spatie/image/pull/159) before ending up here, all because I needed to use TIFF in MLP! It's the first time I've had a PR chain this long!


freekmurze:
Very nice work! Thanks for all the PRs you made in this chain 👏 

> I think it would be much better to obtain the list of supported file extensions and MIME types from spatie/image rather than maintaining a separate set of them in here. However, it's workable for now.

Yeah, I agree, it could be nicer, but this list won't change too much in the near future.

freekmurze:
Could you create a similar PR on the `main` branch for v10 of the package?

Synchro:
> Could you create a similar PR on the `main` branch for v10 of the package?

I can, however, I will need to do a similar shuffle in spatie/image to port the TIFF support to the 2.x branch first.

Synchro:
Actually I'm a bit confused. ML v10 depends on spatie/image 2.x, but spatie/image doesn't seem to have a v2 branch: the PR I did there was against `main`, but it was released as 1.11.0 yesterday; the latest 2.x release is from 2 months ago, but it was also released from `main`. Where should I make a PR to get these patches into 2.x?

freekmurze:
> but spatie/image doesn't seem to have a v2 branch

The main branch is essentially v2. If see that we've made a wrong tag yesterday on spatie/image. We've removed 1.11.0 and replaced it with 2.2.2

To make v1 of spatie/image support TIFFs, send a PR to the v1 branch.

My apologies for the confusion.

PR #2803, opened 2022-02-23
Change to forceDelete in move method of Media

ziming:
Some of us subclass the Media model class to add the SoftDelete trait.

This lead to the delete() in the move method doing a soft delete instead. Which can cause some issues. like if we copy the uuid of the earlier record to the new record. Since uuid is defined as unique in the migrations.


freekmurze:
Thanks!

PR #2813, opened 2022-03-01
Use newer syntax for accessors

phh:
In [Laravel V8.x](https://laravel.com/docs/8.x/eloquent-mutators#defining-an-accessor) you had to define accessors as a method on the class named `get{Attribute}Attribute` which felt a bit "magic". In [newer versions](https://laravel.com/docs/master/eloquent-mutators#defining-an-accessor) you are able to use a newer syntax.

This is pretty much converting all "older" accessors to use the new syntax instead.

### DIFF
--- a/src/MediaCollections/Models/Media.php
+++ b/src/MediaCollections/Models/Media.php
@@ -6,6 +6,7 @@ use DateTimeInterface;
 use Illuminate\Contracts\Support\Htmlable;
 use Illuminate\Contracts\Support\Responsable;
 use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Casts\Attribute;
 use Illuminate\Database\Eloquent\Model;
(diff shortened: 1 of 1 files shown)


freekmurze:
Very nice, thank you!

phh:
Will attempt to fix the PHPStan errors later this evening.

freekmurze:
I'll take care of it, no worries!

phh:
You should be able to add a docblock to the model. Something like:

```
/*
 * @property-read string $extension
 * /
```

Will give it a try later :)

phh:
@mgralikowski You are not supposed to use `getExtensionAttribute()` directly. However this is an accessor so it should be access like `->extension`.

mgralikowski:
Sure, I know what is the idea behind eloquent getters but it is not forbidden to use the full name, I use a "getNameAttribute" very often on the backend side, especially as they are auto-suggested by IDE by default. Short "name" I use in views, methods like pluck etc.

PR #2814, opened 2022-03-01
add property-read for accessors

phh:
Follow up of https://github.com/spatie/laravel-medialibrary/pull/2813


freekmurze:
Thanks!

PR #2817, opened 2022-03-02
Fix phpstan errors in ci

phh:
Since CI is failing I thought I rather just fix it if you don't mind :)

I have:

- Removed no longer failing errors
- Added a few `@return` and `@var` type hints to make PHPStan happy again
- Removed code for the `MediaObserver` `isLumen()` and `isLaravel7orHigher()` since L7 is no longer supported

Types and generics are still not my strong side so if anything is not as expected please let me know!

At least now CI is no longer complaining without just suppressing the errors :)

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -176,11 +176,6 @@ parameters:
 			path: src/Conversions/ConversionCollection.php
 
 		-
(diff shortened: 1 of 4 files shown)


freekmurze:
Thank you!

PR #2823, opened 2022-03-04
[10.x] Added string support for id in deleteMedia()

rcerljenko:
Currently, method `deleteMedia()` in `InteractsWithMedia` trait has a typed param `$mediaId` as an `int` or `Media` object.
That's a problem if you extend a default `Media` model and have a `string` type primary key (e.g.: UUID).
This PR resolves that by adding a string type support for `$mediaId` param.


rcerljenko:
@freekmurze could you check phpstan.. it doesn't seem to have any relation with what I did...

freekmurze:
Thanks!

Issue #2773, opened 2022-02-11
[v10] `Storage::fake()` isn't working correctly anymore


LucaPuddu:
The disk root is not correct anymore when testing with `Storage::fake()`

My config:
```php
'disks' => [
        'test' => [
            'driver' => 'local',
            'root' => storage_path('app/test'),
            'url' => env('APP_URL').'/test',
            'visibility' => 'public',
        ],
]
```
Steps to reproduce:
```php
Storage::fake('test');

$file = UploadedFile::fake()->createWithContent('file.txt', 'lorem ipsum');
$media = $model->addMedia($file)->toMediaCollection(diskName: $this->disk);

// v10
Storage::disk('test')->assertExists('1/file.txt', 'lorem ipsum'); // OK!
echo $media->getPath(); // /var/www/html/storage/app/test/1/file.txt
file_exists($media->getPath()); //false

// v9
Storage::disk('test')->assertExists('1/file.txt', 'lorem ipsum'); // OK!
echo $media->getPath(); // /var/www/html/storage/framework/testing/disks/test/1/file.txt
file_exists($media->getPath()); //true
```

freekmurze:
It seems like all our own tests still pass. 

My guess is that this is a problem caused by Laravel, and not by our package.

dsErick:
I've just updated Laravel to version 9 and MediaLibrary to version 10, and I'm getting the same issue as @LucaPuddu.
The file uploading is working, but on tests the `getPath()` ignores the `Storage::fake()`.

prullenbak:
I also get the same error. I tried to find out what the issue was, and I found a big change in the DefaultURLGenerator's getBasePath() method. In medialibrary 9, it was quite complex. in medialibrary 10, it simply looks into the filesystem config file for the root of your disk. So what happens, is it ignores the fact that you're using a fake disk. 

So I get the exact same result as @LucaPuddu, and I added something

```Storage::fake('test');

$file = UploadedFile::fake()->createWithContent('file.txt', 'lorem ipsum');
$media = $model->addMedia($file)->toMediaCollection(diskName: 'test');

// v10
Storage::disk('test')->assertExists('1/file.txt', 'lorem ipsum'); // OK!
echo $media->getPath(); // /var/www/html/storage/app/test/1/file.txt
echo storage_path(); // /var/www/html/storage
file_exists($media->getPath()); //false

// v9
[...]

PR #2831, opened 2022-03-10
Update DefaultUrlGenerator.php

prullenbak:
update getRootDisk() to find the correct root path when using laravels storage fake.

attempt to fix 
https://github.com/spatie/laravel-medialibrary/issues/2773


freekmurze:
Thank you!

PR #2835, opened 2022-03-13
Add methods to get first-available conversion url

gdebrauwer:
When a new media model has been saved, the conversions are not instantly available as they need to be generated by a job on the queue. That means you have to fall back to the original URL in the meantime to ensure the user does see something instead of an error because the conversion image can not be found. Currently you have to that manually with the following code:
```php
$media->getFullUrl($media->hasGeneratedConversion('small') ? 'thumbnail' : '');
```
That is a bit cumbersome. To fix that, this PR adds a method that returns the URL (or path) of the first available conversion. If no conversion exists, then it returns the URL (or path) of the original file.
```php
$media->getUrl(['small', 'medium', 'large']);
$media->getFullUrl(['small', 'medium', 'large']);
$media->getPath(['small', 'medium', 'large']);
```


freekmurze:
Nice, can you also add examples of these in the docs?

PR #2837, opened 2022-03-13
Update DefaultUrlGenerator.php

matt127127:
Remove base path, get url directly from disk


matt127127:
Fix for the updated code from laravel.

freekmurze:
Thank you!

PR #2843, opened 2022-03-16
Added setConnection in FileAdder processMediaItem

davidrushton:
Thanks for the package.

I'm using it within a SaaS app where there is a single Media model that uses the landlord or tenant connection depending on the context it is used in.

A model's queried media contains the correct results since Eloquent's `hasMany` sets the media model to use the same database connection name as the parent model (https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php#L753-L760)

However when using `$model->addMedia($path)->toMediaCollecton($collection)` I was running into:

```
Illuminate\Database\QueryException

  SQLSTATE[3D000]: Invalid catalog name: 1046 No database selected (SQL: select max(`order_column`) as aggregate from `media`)
  ```
  
I believe this is because the `FileAdder` "news up" a media model without specifying the database connection of the parent. 

This PR updates the `processMediaItem` method to automatically set the media model connection to the same as the parent model if that is set.

### DIFF
--- a/docs/advanced-usage/customising-database-connections.md
+++ b/docs/advanced-usage/customising-database-connections.md
@@ -10,7 +10,8 @@ The built-in model (`Spatie\MediaLibrary\MediaCollections\Models\Media`) will us
 If you need to change this database, connection you can create a custom model and set the `$connection` property (https://laravel.com/docs/8.x/eloquent#database-connections). See <a href="https://docs.spatie.be/laravel-medialibrary/v9/advanced-usage/using-your-own-model">Using your own model</a> for more information.
 
 ```php
-<?
+<?php
+
 namespace App\Models;
 
 use Spatie\MediaLibrary\MediaCollections\Models\Media as BaseMedia;
@@ -27,7 +28,8 @@ class Media extends BaseMedia {
 The `Spatie\MediaLibrary\InteractsWithMedia` trait defines a `MorphMany` relationship to the media model. Eloquent automatically uses the database connection of your parent model when querying the database. In the example below, the user media results will use the `tenant` database connection rather than the application's default connection.
 
 ```php
-<?
+<?php
+
 namespace App\Models;
 
 use Spatie\MediaLibrary\HasMedia;
--- a/docs/advanced-usage/customising-database-connections.md
+++ b/docs/advanced-usage/customising-database-connections.md
@@ -1,0 +1,54 @@
+---
+title: Customising Database Connections
+weight: 13
+---
+
+### Media model connection
+
+The built-in model (`Spatie\MediaLibrary\MediaCollections\Models\Media`) will use the default database connection set for your application.
+
+If you need to change this database, connection you can create a custom model and set the `$connection` property (https://laravel.com/docs/8.x/eloquent#database-connections). See <a href="https://docs.spatie.be/laravel-medialibrary/v9/advanced-usage/using-your-own-model">Using your own model</a> for more information.
+
+```php
+<?
+namespace App\Models;
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by freekmurze** [comment]
Could you solve this by using a custom media model?

https://github.com/spatie/laravel-medialibrary/blob/faf29b67e8c388da2154238c678e0ed63990376a/config/media-library.php#L31

**Comment by davidrushton** [comment]
@freekmurze Unfortunately not. I already do use a custom model but for simplicity it is the same model for both landlord and tenant (both databases have media). For example my default `connection` is `tenant` and a tenant model looks like:

```php
<?php

class User extends Model implements HasMedia {

	use InteractsWithMedia;

}
```

but I also use media on a landlord model (in this case the landlord database tenants table)

```php
<?php

class Tenant extends Model implements HasMedia {

	protected $connection = 'base';
	
	use InteractsWithMedia;

}
```

Fetching media from either correctly sets the connection when using Eloquent's `hasMany` method, but when trying to save new media it is always using the default `tenant` connection rather than landlord.

freekmurze:
Ok, could you document this?

davidrushton:
@freekmurze You mean adding a new md file in the documentation - e.g. advanced-usage/multiple-database-connections.md ?

freekmurze:
@davidrushton Yes! 👍

davidrushton:
@freekmurze Added https://github.com/davidrushton/laravel-medialibrary/blob/feature/model-connection-name/docs/advanced-usage/customising-database-connections.md - hope that is okay.

freekmurze:
Thank you!

freekmurze:
Could you also send this to the v10 branch?

jeff-h:
@davidrushton Am I understanding correctly that this PR aims to ensure that the `media` model will be written to whichever db connection the associated parent model is using? If so, I have two questions/comments:

Firstly @freekmurze, would you accept a PR to [the documentation](https://github.com/spatie/laravel-medialibrary/blob/main/docs/advanced-usage/customising-database-connections.md) to clarify that:

* you will need to create the `media` table in every relevant database?
* you will need to customise the filesystem path media is written to, as each table will share the same set of IDs (eg out-of-the-box every db's media table's first entry will look for the file insde `public/storage/1`)

Secondly, I believe the following code will always attempt to refer to the models in the default DB (due to the use of `static`), which is incorrect. This code would somehow need to configure the connection to the parent model at the time.

https://github.com/spatie/laravel-medialibrary/blob/85b2fb49b20ecd4103ae8f3ef2d9e7af396c804d/src/MediaCollections/Models/Concerns/IsSorted.php#L16-L21

https://github.com/spatie/laravel-medialibrary/blob/85b2fb49b20ecd4103ae8f3ef2d9e7af396c804d/src/MediaCollections/Models/Concerns/IsSorted.php#L34-L45

jeff-h:
(`trait HasUuid` also has the same static problem as above)

davidrushton:
@jeff-h Yes this PR was to force the FileAdder to use the model's correct database connection.

I see what you mean about the `::static` calls. I think you could submit a PR that does something like the following for those traits and calls.

```php
return (int) static::on($this->getConnectionName())
                        ->where('model_type', $this->model_type)
                        ->where('model_id', $this->model_id)
                        ->max($this->determineOrderColumnName());
```

@freekmurze might be able to confirm if that's the best way to handle when the media model connection name is changeable.

jeff-h:
That looks like a nice fix for non-static functions such as `getHighestOrderNumber`. However, I think it may be quite impossible on the static functions mentioned above, at least without changing the parameters of those methods to include the "parent" model (or just its connection, perhaps)?

PR #2850, opened 2022-03-21
Add explanation for processing responses

SahinU88:
Not sure if I missed it somewhere else, but I guess the behaviour is by design like this, but basically we have to call the `toMediaCollection` method to ensure that the database is up to date and that the `model_type` is updated. I was using it with the temporary uploads functionality (which is enabled by default I think) and didn't call the method `toMediaCollection`  because I only have a default one, but noticed that I couldn't retrieve any media-files.

After a bit of reading source code and debugging I noticed that we have to call this, so that the `PendingMediaItem` is converted to the associated model.

Hope my explanation makes sense.

so my code was basically:
```php
$model
    ->addFromMediaLibraryRequest($request->attachments)
    ->usingName(fn(MediaLibraryRequestItem $item) => strtolower($item->name));
```

and had to change it to
```php
$model
    ->addFromMediaLibraryRequest($request->attachments)
    ->usingName(fn(MediaLibraryRequestItem $item) => strtolower($item->name))
    ->toMediaCollection();
```

I am using medialibrary v10 and pro v2 with the blade directives.

Hope this helps

### DIFF
--- a/docs/handling-uploads-with-media-library-pro/processing-uploads-on-the-server.md
+++ b/docs/handling-uploads-with-media-library-pro/processing-uploads-on-the-server.md
@@ -259,7 +259,7 @@ class StoreLivewireCollectionCustomPropertyRequest extends FormRequest
 
 ## Processing responses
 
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks!

PR #2856, opened 2022-03-22
Update percentage value in the config file

ClouddJR:
The `FileSizeOptimizedWidthCalculator` uses the following line when generating widths for responsive images:

`$predictedFileSize *= 0.7;`

I updated the comment in the config file to match this logic.


freekmurze:
Thank you!

Issue #2329, opened 2021-03-11
Rendering a conversion only works if the original media is also renderable


squatto:
I have a model that has a media collection for PDFs. I've registered a simple 150x200 "thumbnail" conversion to the collection. When a PDF is added to the collection, the thumbnail is generated without any problems. However, I am unable to render the "thumbnail" conversion.

This is in my view:

```
@if ($media->hasGeneratedConversion('thumbnail'))
  {{ $media('thumbnail')->attributes(['class' => 'img-thumbnail bg-transparent']) }}
@endif
```

`$media->hasGeneratedConversion('thumbnail') === true` because, as I already mentioned, the conversion ran just fine:

![image](https://user-images.githubusercontent.com/748444/110861220-0caa2e80-827b-11eb-82e1-c72cd47fadf8.png)

The problem is that `$media('thumbnail')` returns an empty string because of the conditional in `\Spatie\MediaLibrary\MediaCollections\HtmlableMedia::toHtml()`:

```php
public function toHtml()
{
    if (! (new Image())->canHandleMime($this->media->mime_type)) {
        return '';
    }

    // ...
}
```

It is checking the mime type of the _original_ media, not the _converted_ media. The original media is a PDF and `\Spatie\MediaLibrary\Conversions\ImageGenerators\Image` only handles image mime types, so `toHtml()` returns an empty string.

I put together a simple example you can run to recreate it. You just need to change the IDs in the `find()` calls.

```php
$files = [
    'image' => Media::find(2550), // change this to a `media.id` for an image
    'pdf'   => Media::find(2551), // change this to a `media.id` for a PDF
];

foreach ($files as $file => $media) {
    dump($file, [
        'mime_type'       => $media->mime_type,
        'can_handle_mime' => (new Image())->canHandleMime($media->mime_type),
        'thumbnail_url'   => $media->getUrl('thumbnail'),
        'thumbnail_html'  => $media->img('thumbnail')->toHtml(),
    ]);
}
```

Here is my output:

![image](https://user-images.githubusercontent.com/748444/110863287-b4c0f700-827d-11eb-868b-878b0eccb50d.png)

Note that on "pdf" there _is_ a "thumbnail_url" value but there _is not_ a "thumbnail_html" value.

For further confirmation, I made the conditional return the string "nope":

![image](https://user-images.githubusercontent.com/748444/110863583-200ac900-827e-11eb-8140-5cad32e9b3c5.png)

[...]

PR #2864, opened 2022-03-30
Fix Htmlable render media

Balsakup:
Resolve issue : https://github.com/spatie/laravel-medialibrary/issues/2329  
  
Use ImageGeneratorFactory to generate htmlable media instead of using directly hard coded `Image` generator


freekmurze:
The tests fail for this one, could you take a look at that?

freekmurze:
Thank you!

PR #2890, opened 2022-04-18
Added support for passing attributes to toHtml(), added attribution sanitization

Braunson:
- Added support to pass attributes array to `Media.php` `toHtml()` method.
- Added sanitization of attributes to the same method

**Reference**: https://github.com/spatie/laravel-medialibrary-pro/discussions/423

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,6 +31,7 @@
         "illuminate/database": "^9.0",
         "illuminate/pipeline": "^9.0",
         "illuminate/support": "^9.0",
+        "illuminate/view": "^9.0",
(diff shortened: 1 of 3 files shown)


freekmurze:
Could you add a test that proves that this works?

Braunson:
@freekmurze I'm working on it, trying to add a test and run it locally but I'm getting `could not find driver`, any idea? I see `sqlite` is set in the `TestCase` file.

Braunson:
Any ideas @freekmurze?

freekmurze:
No idea really, it seems to be working here locally.

PR #2904, opened 2022-05-03
Prevent error when extracting an image from a video

prullenbak:
Prevent error when the video is exactly as long (in seconds) as the timestamp set in extractVideoFrameAtSecond()


prullenbak:
It was a coincidence, but in our case we got errors because a client was uploading a very short video that was exactly as long as our "extractVideoFrameAtSecond"...This prevents that.

freekmurze:
Thanks!

PR #2914, opened 2022-05-11
Added support to conditionally enable or disable queue on media conversion

aroutinr:
This PR adds two new methods to the Conversion class to allow to enable or disable queued conversion execution based on a condition.

For example, in my case I have a model that can be used with from the web and from the artisan console. I want to be able to process the conversions from the web without queuing the conversion, to do so I must do something like this:

```
    public function registerMediaConversions(Media $media = null): void
    {
        if (app()->runningInConsole()) {
                ->height(350);
        } else {
                ->height(350)
                ->nonQueued();
        }
    }
```

It would be much cleaner if you could define if you want to queue or not with the same method.

```
    public function registerMediaConversions(Media $media = null): void
    {
            ->height(350)
            ->nonQueuedIf(!app()->runningInConsole());
    }
```

In case the application is not running in the console, the condition will be true and the conversion will be done without using the queue.

### DIFF
--- a/docs/api/defining-conversions.md
+++ b/docs/api/defining-conversions.md
@@ -28,20 +28,20 @@ public function addMediaConversion(string $name): \Spatie\MediaLibrary\Conversio
  * @param string $collectionNames,...
  */
 public function performOnCollections($collectionNames): self
-``` 
+```
 
 ### queued
 
-```php 
+```php
 /*
  * Mark this conversion as one that should be queued.
  */
- public function queued(): self
+ public function queued(bool $performOnQueue = true): self
 ```
 
 ### nonQueued
 
-```php 
+```php
 /*
  * Mark this conversion as one that should not be queued.
  */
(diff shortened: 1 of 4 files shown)


freekmurze:
Nice idea! Instead of adding extra methods, we could just pass a boolean to `queued`

Usage:

```php
->queued(false)
```

Implementation

```php
 public function queued($performOnQueue = true): self
    {
        $this->performOnQueue = performOnQueue;
         return $this;
     }
```

That feels much more lightweight to me. Could you implement it like this, add tests and docs?

Thank you!

aroutinr:
Yes, youre right, that feels so much better! 

I think the `nonQueded()` its a good helper method, and I belive you wont be agree to remove it but with this new implementation, the `nonQueded()` method can also be removed and instead just use `queued(false)`

Let me know what you think and I can update the PR with that change too.

freekmurze:
Let's indeed just focus on adding that boolean to `queued` and leave the rest as it is.

PR #2925, opened 2022-05-19
Fix `media-library:clean` command when using custom path generators

### LINKED ISSUE #2922: Command `media-library:clean` does not use custom path generators
Hello,

The `media-library:clean` command uses the default `Spatie\MediaLibrary\Support\PathGenerator\DefaultPathGenerator` instead custom path generators that can be configured with the `media-library.custom_path_generators` configuration key (it's [hard-coded in the command file](https://github.com/spatie/laravel-medialibrary/blob/main/src/MediaCollections/Commands/CleanCommand.php#L47)). Due to that, if we use a custom path generator for models and configure a custom path for conversions, the deprecated images inside this custom directory won't be cleaned by the command (since the command do not look in the correct directory because it uses the default path generator included in the package).

As a simple example, let's assume I added the media relationship on the default User model and defined the following custom path generator:

```PHP
<?php

namespace App;

use Spatie\MediaLibrary\Support\PathGenerator\DefaultPathGenerator;

class UserPathGenerator extends DefaultPathGenerator
{
    /**
     * @inheritDoc
     */
    public function getPathForConversions(Media $media): string
    {
        return $this->getBasePath($media) . '/my-custom-path/';
    }
}
```

and configuration in the `config/media-library.php` file:

```PHP
/*
 * Here you can specify which path generator should be used for the given class.
 */
'custom_path_generators' => [
    \App\Models\User::class => \App\UserPathGenerator::class,
],
```

Media conversions will be correctly stored in a `/my-custom-path` sub-directory, the `media-library:regenerate` command will correctly regenerate conversions in this directory, but the `media-library:clean` command won't see deprecated conversions in this sub-directory but will instead look in the `/conversions` sub-directory.

I don't know how the dynamic part of the `media-library.custom_path_generators` is handled inside the package but it should probably also be used in the `media-library:clean` command.

YellowFish085:
This PR fixes #2922 that I opened myself (this is a personal account, the account used to open the PR is my professional account).

Two modifications were done in this PR:

#### 1.

The path generator used by the `media-library:clean` command is now dynamically created in the `deleteConversionFilesForDeprecatedConversions` method for each processed media using the [PathGeneratorFactory](https://github.com/spatie/laravel-medialibrary/blob/main/src/Support/PathGenerator/PathGeneratorFactory.php).

This modification allows custom path generators configured via the `media-library.custom_path_generators` to actually be used by the command and not only the [DefaultPathGenerator](https://github.com/spatie/laravel-medialibrary/blob/main/src/Support/PathGenerator/DefaultPathGenerator.php) included in this package, and consequently for conversions saved in custom path to be properly deleted.

#### 2.

Following this change, another issue came to light where the original image would be deleted if the path for conversions is configured to be the same as the original image path.

A new check was added to the `deleteConversionFilesForDeprecatedConversions` method to ensure that the original image is not mistakenly considered as a deprecated conversion and deleted.

---

Side note: This is my first PR to open source project and I'm not really familiar with tests, so I hope the changes included in this PR are satisfying enough.

### DIFF
--- a/src/MediaCollections/Commands/CleanCommand.php
+++ b/src/MediaCollections/Commands/CleanCommand.php
@@ -111,6 +111,7 @@ class CleanCommand extends Command
 
         collect($currentFilePaths)
             ->reject(fn (string $currentFilePath) => $conversionFilePaths->contains(basename($currentFilePath)))
+            ->reject(fn (string $currentFilePath) => $media->file_name === basename($currentFilePath))
             ->each(function (string $currentFilePath) use ($media) {
                 if (! $this->isDryRun) {
                     $this->fileSystem->disk($media->disk)->delete($currentFilePath);
--- a/src/MediaCollections/Commands/CleanCommand.php
+++ b/src/MediaCollections/Commands/CleanCommand.php
@@ -14,7 +14,7 @@ use Spatie\MediaLibrary\MediaCollections\Exceptions\DiskDoesNotExist;
 use Spatie\MediaLibrary\MediaCollections\MediaRepository;
 use Spatie\MediaLibrary\MediaCollections\Models\Media;
 use Spatie\MediaLibrary\ResponsiveImages\RegisteredResponsiveImages;
-use Spatie\MediaLibrary\Support\PathGenerator\DefaultPathGenerator;
+use Spatie\MediaLibrary\Support\PathGenerator\PathGeneratorFactory;
 
 class CleanCommand extends Command
 {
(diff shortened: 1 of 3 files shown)


freekmurze:
Thank you!

petsoukos:
Will this make it into v9 or older versions dont get support ?

freekmurze:
We don't actively support the older versions anymore, but we would still accept and merge PRs for bug fixes.

petsoukos:
:disappointed: 
I wish I knew Pest so I could do a PR for v9. I'm tied into v9 without the possibility to upgrade to Laravel 9 for the foreseeable future. For now I will have to fiddle with it in the vendors directory since I rarely need this feature.

Issue #2942, opened 2022-05-30
Custom model-based path generator doesn't work when using morph map


aprokopenko:
**Problem**
We're using v9.12.3 and found a bug in `PathGeneratorFactory`:

It has an undocumented feature to provide `config('media-library.custom_path_generators')` and specify custom generators based on media model class:
https://github.com/spatie/laravel-medialibrary/blob/9.12.3/src/Support/PathGenerator/PathGeneratorFactory.php#L30

Best practices recommend using Relation::morphMap() for storing morph model type. In this case, the code provided doesn't work, because `$media->model_type` is not a class. Furthermore, if I specify config using morph keys, instead of the class name, then `is_a()` doesn't compare this as equal, because the morph key is not a class.

Will it be possible to fix this (or we can create a pull request for this)?

**Possible solutions:**
a. add OR case of simple string compare of `$media->model_type` and `$modelClass`
b. add OR case of `is_a((string)Relation::getMorphedModel($media->model_type), $modelClass, true)`

Please let me know what is the preferred fix for this. Ideally if you fix this by your own, but we can do a Pull Request as well.

freekmurze:
I'd appreciate a PR for this. Let's use approach B. 👍

aprokopenko:
PR is done: https://github.com/spatie/laravel-medialibrary/pull/2947

aprokopenko:
@freekmurze , For some reason all tests failed in the created PR, however, we didn't update anything related to errors inside the tests. Is it okay?
We connected the PR diff as composer patch in our project and the package and patch works fine.

PR #2947, opened 2022-06-03
Fixed: Custom model-based path generator doesn't work when using morph map

jc-atsinya:
Fix for issue #2942


PaolaRuby:
Look like this break the tests

freekmurze:
@aprokopenko Could you target v10 first. If it works there, it can be backported to v9

PR #2950, opened 2022-06-13
add default image to getFirstUrl as general 

husseinzaher:
Pleas tell me about this update


freekmurze:
Thanks for your PR, but we're not going to merge this in. You can override the function on your models in your project.

PR #2955, opened 2022-06-17
Update preparing-your-model.md

timgavin:
When implementing the `registerMediaConversions()` method, it is not clear that additional models need to be imported. Hopefully this PR makes it a little more clear for users with an IDE that does not automatically import the required models.


freekmurze:
Thanks!

PR #2958, opened 2022-06-21
Add down function to media migration file

Bashar-Modallal:
While making migrate:refresh command I got an error said that media table still exists in database. so I added down function to migration file.


freekmurze:
We don't use down migration in our projects and won't be adding this (see the other PRs on this)

Of course, you can add the `down` migration to your published migration.

PR #2964, opened 2022-06-24
Add @template annotation to FileAdder

sforward:
This pull request adds the type hint for the Media model in the FileAdder class. This makes it possible to avoid false-positives from IDEs and static analysis tools when extending the Media model.


freekmurze:
Thanks!

PR #2970, opened 2022-06-29
Upgrade example using JSON_EXTRACT

### LINKED ISSUE #2969: MariaDb issue when upgrading from v8 to v9
I ran into an issue when upgrading from v8 to v9 when using a Mariadb database. I was getting 
`Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '>"$.generated_conversions", `media`.`updated_at` = ? where (`generated_conver...' at line 1 (SQL: update `media` set `generated_conversions` = custom_properties->"$.generated_conversions", `media`.`updated_at` = 2022-06-29 08:04:38 where (`generated_conversions` is null or `generated_conversions` =  or JSON_TYPE(generated_conversions) = 'NULL') and JSON_LENGTH(custom_properties) > 0)` 
when I added the AddGeneratedConversionsToMediaTable migration. 

The solution in the end was to edit the migration file, line 31.

Change
`'generated_conversions' => DB::raw('custom_properties->"$.generated_conversions")`
To
`'generated_conversions' => DB::raw('JSON_EXTRACT(custom_properties, "$.generated_conversions")')`

Think this could possibly be added to the docs as anyone using a MariaDb database will run into this issue.

erikn69:
Closes #2969
>Change
>'generated_conversions' => DB::raw('custom_properties->"$.generated_conversions")
To
'generated_conversions' => DB::raw('JSON_EXTRACT(custom_properties, "$.generated_conversions")')
>
>Think this could possibly be added to the docs as anyone using a MariaDb database will run into this issue.

[JSON_EXTRACT Tested on Mysql, MariaDB](https://paiza.io/projects/x6dxlM9iaITCzP6720halw)

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -41,7 +41,7 @@ class AddGeneratedConversionsToMediaTable extends Migration {
             })
             ->whereRaw("JSON_LENGTH(custom_properties) > 0")
             ->update([
-                'generated_conversions' => DB::raw('custom_properties->"$.generated_conversions"'),
+                'generated_conversions' => DB::raw("JSON_EXTRACT(custom_properties, '$.generated_conversions')"),
                 // OPTIONAL: Remove the generated conversions from the custom_properties field as well:
                 // 'custom_properties'     => DB::raw("JSON_REMOVE(custom_properties, '$.generated_conversions')")
             ]);


freekmurze:
Thanks!

PR #2972, opened 2022-07-01
clearMediaCollection() documentation clarification

chimit:
`$yourModel->clearMediaCollection()` doesn't delete all media in all collections.

Is there any shorthand method to delete all media in all collections?


freekmurze:
Thanks!

PR #2985, opened 2022-07-22
Added the required arguments for the getMedia method

titonova:
Kept wondering why the `$model->getMedia()` method wasnt working, when it clearly had media in the db. Realized i had to specify the collection name


cesarreyes3:
> Realized i had to specify the collection name

only when you are using a different one than the default, so `$model->getMedia()` is ok

freekmurze:
I think the current example is ok.

PR #2992, opened 2022-07-29
Order media relationship

Propaganistas:
Any reason for not ordering the `media()` relationship by default?
Especially because the order does get applied in `loadMedia()`:

https://github.com/spatie/laravel-medialibrary/blob/19576a5fab5f173b7edcd8444e851b36203b6695/src/InteractsWithMedia.php#L529-L541

### DIFF
--- a/src/InteractsWithMedia.php
+++ b/src/InteractsWithMedia.php
@@ -57,7 +57,7 @@ trait InteractsWithMedia
 
     public function media(): MorphMany
     {
-        return $this->morphMany(config('media-library.media_model'), 'model');
+        return $this->morphMany(config('media-library.media_model'), 'model')->orderBy('order_column');
     }
 
     /**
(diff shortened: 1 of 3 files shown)


freekmurze:
Seems like the tests are failing. Could you take a look at that?

Propaganistas:
Done

freekmurze:
Thanks!

rmsj:
`public function media(): MorphMany
    {
        return $this->morphMany(config('media-library.media_model'), 'model')->orderBy('order_column');
    }`

This has introduced a but where the query that eloquent runs now looks like this:

`select "file_name" from "media" where "media"."model_type" = ? and "media"."model_id" = ? and "media"."model_id" is not null and "collection_name" = ? order by "order_column" asc, "size" desc`

Propaganistas:
For future reference: this PR has been reverted because it breaks custom ordering.

If media will always be sorted by the order column, then this behavior can be easily achieved by adding a global scope to the Media model:

```php
// In AppServiceProvider.php - boot()
Media::addGlobalScope('ordered', function($query) {
    $query->ordered();
});
```

PR #3000, opened 2022-08-11
Add a note about limitation of generators

Orrison:
Add a note about limitation of generators to deter users from going down the path of attempting to use this functionality for anything other than images.


freekmurze:
Thanks!

PR #3002, opened 2022-08-18
feat: add ability to replace original in conversion

### LINKED ISSUE #1526: How to optimize and replace the original image when uploading to media?
On the $post,there will be request featured containing image of the post,
i have used conversion to create the thumb of the image but i want to optimize the image also original one.
i am using
`$post->addMediaFromRequest('featured')->toMediaCollection('featured');`
Where on the Post model
` $this->addMediaConversion('thumb')
            ->width(env('THUMB_WIDTH',150))
            ->height(env('THUMB_HEIGHT',150))
            ->performOnCollections('featured');`

I find the thumb image is being optimized but not original one.
I could not even find on the documentation too.

Tofandel:
This is a simple feature to allow conversions to be performed on the original image

Fixes #1526

### DIFF
--- a/docs/converting-images/defining-conversions.md
+++ b/docs/converting-images/defining-conversions.md
@@ -165,3 +165,24 @@ public function registerMediaConversions(Media $media = null): void
 ```
 
 Be aware that this can lead to a hit in performance. When processing media the media library has to perform queries to fetch each separate model.
+
+
+## Converting the original image
+
+If one of your collection will always have the same size or format, it is a good idea to run some conversions directly on the original image to preserve space on your server
+
+Luckily it is possible using `replaceOriginal()` on the conversion, just keep in mind you will lose the original image and only the optimized image will be stored
+
+```php
+// in your model
+public function registerMediaConversions(Media $media = null): void
+{
+    $this->addMediaConversion('original')
+          ->width(500)
+          ->height(500)
+          ->format('jpg')
+          ->quality(80)
+          ->replaceOriginal()
+          ->performOnCollections('images');
+}
+```
--- a/src/Conversions/Actions/PerformConversionAction.php
+++ b/src/Conversions/Actions/PerformConversionAction.php
@@ -40,9 +40,18 @@ class PerformConversionAction
             );
         }
 
-        app(Filesystem::class)->copyToMediaLibrary($renamedFile, $media, 'conversions');
+        $fs = app(Filesystem::class);
+        if ($conversion->shouldReplaceOriginal()) {
+            $fs->removeFile($media, $fs->getMediaDirectory($media) . $media->file_name);
+            $fs->copyToMediaLibrary($renamedFile, $media);
 
-        $media->markAsConversionGenerated($conversion->getName());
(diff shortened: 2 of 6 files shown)


freekmurze:
Thank you for your work on this, but I'm going to pass.

The philosophy behind media library is that we never touch the original image. Any changes to an image should be done before adding it to the media library. Our image and image optimizer packages allow for easy manipulation of any image.

Tofandel:
Well that's a bit problematic because the media library basically creates this image if you import it from a url or a string, so how are we supposed to optimize it before that, except by reinventing the whole downloader logic?

michaellindahl:
I would definitely agree that this adds value and I've run into a situation where I've needed to do this as well. However, I also understand that this could add complexity to the media library package. We'd want to make sure that re-running media conversions didn't cause this to be re-ran on the original image. 

At first my solution involved copying the logic from [`DefaultDownloader`](https://github.com/spatie/laravel-medialibrary/blob/main/src/Downloaders/DefaultDownloader.php) to download the file.
But with the help of [`invade`](https://github.com/spatie/invade) we can intercept the image prior to handing it off to the media library:

```php
/** @var \Spatie\MediaLibrary\MediaCollections\FileAdder */
$fileAdder = $this->addMediaFromUrl($url);
$filePath = invade($fileAdder)->pathToFile;

/** @var \Spatie\Image\Image */
$image = Image::load($filePath);
$image->optimize();
$image->save();

/** @var \Spatie\MediaLibrary\MediaCollections\Models\Media */
$media = $fileAdder->toMediaCollection();
```

Tofandel:
I do agree, as is, this PR doesn't have the best design, I'm thinking we can either filter the conversions based on the flag and run this on the tmp file before it's even added in the media library

Or add a different method to add manipulations (to differentiate from conversions) so we don't need to filter at all. It can still use the conversion class under the hood for its perfect Api (or extend it)

With this design the original would never make it to the final file system and only the manipulated image and we're sure the conversion will run only once

@freekmurze would you be willing to add this feature with a better  cleaned up design?

Tofandel:
Thanks for the suggestion with invade, will use as a dev workaround

While it may do in a testing environnement, it's really not something I would ship to production as that means accessing private variables which are not prone to breaking changes requirements

So at the very least some events are needed in the library (I would say around [there](https://github.com/spatie/laravel-medialibrary/blob/main/src/MediaCollections/FileAdder.php#L446)) if we want just a chance at implementing optimization

I'm a big fighter for this as disk space is a big way to reduce CO2 emissions (between the multiple drives, the backups, backups of backups and cpu usage of those as well as network load) and this library just doesn't make it easy

Tofandel:
@freekmurze Any new thoughts? This is clearly something the users of the lib want

Should we go with an event or a better design of the feature?

ipontt-neering:
Since this is "so easy" but no examples are available, here's an example.
```php
$file = request()->file('image');

Image::load($file)->format('webp')->save($file->getRealPath());

$media = $model->addMediaFromRequest('image')
        ->usingFileName(pathinfo($file->getClientOriginalName(), PATHINFO_FILENAME). '.webp')
        ->toMediaCollection('image');
```

PR #3003, opened 2022-08-19
Update CleanCommand to delete depreciated responsive images

michaellindahl:
When running `php artisan media-library:clean` we should be clearing responsive images for conversions that are no longer used, as well as for active conversions that are no longer using responsive images. Fixes  #2397


freekmurze:
Could you also update the tests, so we don't break this functionality in the future?

michaellindahl:
@freekmurze  Added a unit test for this and also deleted the duplicate unit test / file. Based on the structure of the existing CleanConversionsTest this seemed like the right thing to do.

freekmurze:
Thanks!

PR #3004, opened 2022-08-19
Update README.md to reference pest

michaellindahl:
Minor, but useful.
```
Fatal error: Uncaught Pest\Exceptions\InvalidPestCommand: Please run `./vendor/bin/pest` instead of `/vendor/bin/phpunit`. in /Users/michaellindahl/Projects/laravel-medialibrary/vendor/pestphp/pest/src/TestSuite.php:114
```


freekmurze:
Thanks!

PR #3005, opened 2022-08-19
Add compatibility for `migrate:rollback`

jordanhavard:
When publishing the migrations. In the event someone wants to rollback, this change will prevent an error from occurring if someone rollsback and then re-migrates


freekmurze:
Thanks, but I'm going to pass. We don't use down migrations in any of our packages. Take a look at the numerous previous similar PRs.

Issue #3020, opened 2022-09-02
Regression ? in the MediaStream class


FabienLucini:
Hello, this used to be called with the 'static' keyword but now it's called with the 'self' keyword.
In this scenario, we can't overwrite the differents method of this class if we make a new class that extends this one

https://github.com/spatie/laravel-medialibrary/blob/6f82e03359dbe7add037e9e8c5e3904522155f28/src/Support/MediaStream.php#L20

See previous version : https://github.com/spatie/laravel-medialibrary/blame/47807503f79bd5a4385b70a6053fb886dfa344d1/src/Support/MediaStream.php

Is there any specific reason or would it be possible to put back the 'static' keyword ?

Many thanks

FabienLucini:
Solution for now is to extend also the create method as well in our own class : 
```
class MediaStream extends \Spatie\MediaLibrary\Support\MediaStream
{
    /**
     * On surcharge cette méthode pour avoir le nom du fichier en mode name+extension
     * Et non en mode file_name (qui est chez nous un hash illisible)
     * Ce que la lib ne permet pas de faire par défaut.
     */

    public static function create(string $zipName): self
    {
        return new static($zipName);
    }
```

PR #3021, opened 2022-09-02
Change the create method to use 'static'

FabienLucini:
It is better to use the `static` keyword instead of `self` in order to let the methods of this class beeing override when the class is extended 
see discussion here #3020


freekmurze:
Thanks!

Issue #3015, opened 2022-08-29
Latest optimisation is sort order query breaks media copy for single file collections


victor-priceputu:
1. Create a model
2. Attach a media to the model, after registering the media collection as a single file.
3. Attach another media to the model (basically have 1 media item with order 2)
4. Move another media from another model to this one

Expected: the new media would belong to the model
Actual: the old media belongs to the model

This can be replicated with other steps as well, for example:
Have a model with 2 media items (which have order 1 and 2)
Move a media item from another model (that has the order 1)

The copied media has order 1, not 3. Biggest issue is in Single model collections.

Because copy persist the order of the media, of that order number is lower than the latest entry you have for the model, then the order of the collection is broken. When using single file collection it completely breaks the copy functionality.

Issue introduced with the optimisation in https://github.com/spatie/laravel-medialibrary/pull/2840

freekmurze:
Could you PR a tests that proves that this is broken?

victor-priceputu:
I will make one as soon as I can spare some time

victor-priceputu:
PR created with the test, that is failing locally. Thank you for looking into this

PR #3025, opened 2022-09-06
#3015 - Add test for single file collections not working properly when copying media from another model.

victor-priceputu:
Added test to replicate issue presented in https://github.com/spatie/laravel-medialibrary/issues/3015

### DIFF
--- a/tests/Feature/FileAdder/MediaConversions/MediaCollectionTest.php
+++ b/tests/Feature/FileAdder/MediaConversions/MediaCollectionTest.php
@@ -225,6 +225,30 @@ test('if the single file method is specified it will delete all other media and 
     expect($model->getMedia('images'))->toHaveCount(1);
 });
 
+test('can copy the media to already populated single file collection', function () {
+    $testModel = new class () extends TestModelWithConversion {
+        public function registerMediaCollections(): void
+        {
+            $this
+                ->addMediaCollection('images')
+                ->singleFile();
+        }
+    };
+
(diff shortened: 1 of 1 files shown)


freekmurze:
That is indeed failing, could you also add the fix to this PR?

victor-priceputu:
I will try to make an efficient fix. I am thinking of checking if we can take the max from the already-loaded media collection (if possible) and in case we cannot, to make the DB call for the sort, to try and find an optimal solution.

will have to make some local testing with bigger collections to see which way is better rounded for more cases when it comes to the amount of data

patinthehat:
@victor-priceputu Are you still able to work on adding a fix for this issue? If you don't have time right now no problem - I can look into implementing a fix in a separate PR if necessary.  Thanks!

victor-priceputu:
@patinthehat unfortunately the quickest I can work on it properly is in one month.
I just updated my fork and I cannot make the fix in just a few minutes I have available at the moment (there are some tests failing for me and I have to check them out as well)

patinthehat:
@victor-priceputu Thanks for the update.  Please post a comment once you've submitted a PR or let us know if you don't have time and we'll create an issue for this that someone else can pick up. Thanks!

victor-priceputu:
Hey, I tried to do this over the holiday but had no time, and do not see any spare time any time soon. If anyone wants to pick this up, please feel free to do so with my thanks

freekmurze:
Closing due to inactivity.

PR #3026, opened 2022-09-06
Add getPathRelativeToRoot as a public method in the Media model

emielmolenaar:
Add `getPathRelativeToRoot` as a public method in the Media model for convenience. For example, we use this to get the contents of a conversion via the filesystem (not via a URL) to perform actions based on this conversion. 

The method `getPath` doesn't work for us because we have configured a root path for the S3 disks where the media is stored.


freekmurze:
Could you add a test for that method?

emielmolenaar:
Done!

freekmurze:
Thank you!

PR #3028, opened 2022-09-10
Allow the Github actions to run locally via act

emielmolenaar:
1. Invoke `apt-get` with `-y` because `apt-get` would wait for input when running the actions locally.
2. Add some docs about act.


freekmurze:
Thanks!

PR #3030, opened 2022-09-13
Fixed variable type at responsive width calculator

boryn:
As we do the multiplication by 0.7, the $predictedFileSize may not always be integer


patinthehat:
@boryn To help move this PR forward, could you add more of a description of the functionality of this PR along with adding unit tests as appropriate? Thank you!

freekmurze:
Closing due to inactivity.

boryn:
In the `calculateWidths()` in the `FileSizeOptimizedWidthCalculator()` there is:

```php
$predictedFileSize = $fileSize;
// ...
$predictedFileSize *= 0.7;
```

which means, that for example with the initial filesize of 121121, the `$predictedFileSize` multipilicated by 0.7 will give a result of 84784.7 which IS NOT an integer. That's why `finishedCalculating()` should accept float as `$predictedFileSize`.

PR #3035, opened 2022-09-19
Add media-library:transform-path command to migrate to a new path generator

tibbsa:
This PR adds a new command to facilitate migrating existing media objects into a new path generation scheme by moving all existing objects from the "old" path generator into the schema provided by a "new" path generator.

Use case: I recently encountered a project which used the default `media/id` folder structure, which is ordinarily fine, except that there were many hundreds of thousands of images being added.  While not technically hitting a file system limitation, anyone who inadvertently tried to list all the files in that directory was in for an unpleasant wait.  However, changing to a sharded directory structure meant not only changing to a new path generator, but migrating all of the existing data into the new format. 

[A Laracasts discussion](https://laracasts.com/discuss/channels/general-discussion/spatie-media-library-can-i-add-a-custom-path-generator-then-update-all-old-media-to-reflect-the-new-structure-easily) put me on the right track, but I realized that solution would not handle conversions, particularly if on a separate disk, nor warn about potential conflicts, etc.

This moves both primary media and conversions, and will work even if conversions are on a different disk. 

NOTE: There is an unrelated but necessary and included change to `TestPathGenerator`, which was adding an extra forward slash to conversion/responsive image paths leading to errors in my testing.  The fix does not appear to have broken any of the other existing tests.

### DIFF
--- a/docs/advanced-usage/migrating-between-directory-structures.md
+++ b/docs/advanced-usage/migrating-between-directory-structures.md
@@ -1,0 +1,46 @@
+---
+title: Migrating between custom directory structures
+weight: 5
+---
+
+If you wish to adopt or change your custom directory structure but already have `Media` objects created, 
+it is necessary not only to update the `path_generator` config setting, but also to move the 
+existing media files (and potentially any associated conversions) to match the new folder structure.
+
+Begin by [creating a custom path generator to meet your requirements](using-a-custom-directory-structure.md), but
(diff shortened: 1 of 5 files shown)


freekmurze:
Hi, thank you for this, but I don't want to maintain this code.

Maybe you can put it in a gist / blog post of your own, and I'll happily link to it from our docs.

Issue #3045, opened 2022-09-30
Call to undefined method FFMpeg\\Media\\Audio::frame()


rcerljenko:
Hi,

This error happens when trying to generate a thumb from uploaded video file:
```
Call to undefined method FFMpeg\\Media\\Audio::frame()

on

.../vendor/spatie/laravel-medialibrary/src/Conversions/ImageGenerators/Video.php:27
```
And indeed even intelli complains:

https://ibb.co/613qgfK

rcerljenko:
ok strange thing happened
`open()` method returns either Video or Audio object depending on the processed file.
now our frontend team uploaded audio only .mp4 file which has only audio track without video and this function then returns an Audio object BUT it still tries to generate a thumb from it

rcerljenko:
i'll provide a PR for that

PR #3046, opened 2022-09-30
Fix for audio only "video" files

rcerljenko:
Fix for https://github.com/spatie/laravel-medialibrary/issues/3045


freekmurze:
Thanks Ricardo!

PR #2208, opened 2020-11-27
update docs to inform users to enable pdf-to-image

thejacer87:
update docs to make sure people know to enable pdf-to-image

https://github.com/spatie/laravel-medialibrary/issues/2207

PR #3047, opened 2022-09-30
Hint to additional installation instructions

driesvints:
Builds further upon https://github.com/spatie/laravel-medialibrary/pull/2208 but in a place where I really believe this should be hinted at. Spent a bit too much time figuring this out 😅


freekmurze:
Thanks!

PR #3057, opened 2022-10-12
Add getLastMedia getLastMediaUrl getLastTemporaryUrl getLastMediaPath methods.

ChenRuihua:
Add getLastMedia getLastMediaUrl getLastTemporaryUrl getLastMediaPath methods.


freekmurze:
Thank you, but I'd like to keep this package as simple as it can be.

Add these methods in your own project.

PR #3067, opened 2022-10-21
Throw error if disk cannot be accessed

Silver343:
This PR changes the behaviour when a filesytem cannot be accessed as seen in #3010, it will throw and exception and not create a new media record.

### DIFF
--- a/src/MediaCollections/FileAdder.php
+++ b/src/MediaCollections/FileAdder.php
@@ -3,6 +3,7 @@
 namespace Spatie\MediaLibrary\MediaCollections;
 
 use Closure;
+use Illuminate\Support\Facades\DB;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Facades\Storage;
 use Illuminate\Support\Traits\Macroable;
@@ -431,6 +432,8 @@ class FileAdder
 
         $this->checkGenerateResponsiveImages($media);
 
+        DB::beginTransaction();
+
         if (! $media->getConnectionName()) {
             $media->setConnection($model->getConnectionName());
         }
@@ -470,6 +473,8 @@ class FileAdder
                 $model->clearMediaCollectionExcept($media->collection_name, $collectionMedia->reverse()->take($collectionSizeLimit));
             }
         }
+
+        DB::commit();
     }
 
     protected function getMediaCollection(string $collectionName): ?MediaCollection
--- a/src/MediaCollections/FileAdder.php
+++ b/src/MediaCollections/FileAdder.php
@@ -3,12 +3,12 @@
 namespace Spatie\MediaLibrary\MediaCollections;
 
 use Closure;
-use Illuminate\Support\Facades\DB;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Facades\Storage;
 use Illuminate\Support\Traits\Macroable;
 use Spatie\MediaLibrary\Conversions\ImageGenerators\Image as ImageGenerator;
 use Spatie\MediaLibrary\HasMedia;
+use Spatie\MediaLibrary\MediaCollections\Exceptions\DiskCannotBeAccessed;
 use Spatie\MediaLibrary\MediaCollections\Exceptions\DiskDoesNotExist;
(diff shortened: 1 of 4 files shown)


src/MediaCollections/FileAdder.php
```diff
@@ -431,6 +432,8 @@ protected function processMediaItem(HasMedia $model, Media $media, self $fileAdd
 
         $this->checkGenerateResponsiveImages($media);
 
+        DB::beginTransaction();
```
freekmurze, line 435: I wonder if this isn't breaking. What would happen if the user already started a transaction?

patinthehat: @freekmurze I had the same thought, perhaps there's a better way to approach this functionality without using transactions.  

@Silver343, what are your thoughts on this?

Silver343: @freekmurze @patinthehat I believe Laravel handles nested transactions by using a  [transaction counter](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Concerns/ManagesTransactions.php), but I may be misunderstanding it.

If the functionality could be created without using transactions that may be preferable.

src/MediaCollections/Filesystem.php
```diff
@@ -129,26 +131,38 @@ public function copyToMediaLibrary(string $pathToFile, Media $media, ?string $ty
...
+            $success = $this->filesystem
                 ->disk($diskName)
                 ->put($destination, $file);
 
             fclose($file);
 
+            if (!$success) {
+                Throw DiskCannotBeAccessed::create($diskName);
+
+                DB::rollback();
```
Silver343, line 143: Not sure if the exception or rollback should come first.

patinthehat:
@Silver343 Maybe a better approach is to perform a check to see if the disk can be accessed first, and immediately fail if necessary instead of using a transaction. Thoughts?

patinthehat:
@Silver343 I've refactored your code to remove the use of database transactions.  Are you okay with the changes I made?

ping @freekmurze

src/MediaCollections/Filesystem.php
patinthehat: I refactored and removed this; see comment below.

src/MediaCollections/FileAdder.php
patinthehat: I refactored and removed this; see comment below.

Silver343:
looks good to me, not sure if the change in return type to add and add remote would be considered breaking changes.

patinthehat:
@freekmurze I believe this PR is now ready for review.

tests/Feature/FileAdder/IntegrationTest.php
```diff
@@ -594,3 +596,22 @@
 
     expect($result->getStatusCode())->toEqual(200);
 });
+
+it('will throw and exception and not create a record in database if file cannot be added', function () {
+
+    $this->expectException(DiskCannotBeAccessed::class);
```
freekmurze, line 602: Instead of using `expectException`, tack on `throws` at the end of the test (see Pest docs for more info).

tests/Feature/FileAdder/IntegrationTest.php
```diff
@@ -594,3 +596,22 @@
...
+
+    config()->set('filesystems.disks.minio_disk', [
```
freekmurze, line 604: Is this a deliberately wrongly configured disk?

Maybe name it `invalid_disk` for clarity.

freekmurze (changes requested):

tests/Feature/FileAdder/IntegrationTest.php
patinthehat: Done, nice catch.

tests/Feature/FileAdder/IntegrationTest.php
patinthehat: Done.

patinthehat:
@freekmurze To keep this PR moving along and get it merged/closed ASAP, I've pushed the changes you requested to the unit test on behalf of the original PR author (since they were minor).

Would you please re-review at your convenience? Thanks! :+1:

freekmurze:
Thanks!

PR #3082, opened 2022-11-04
Fix typo in property name in Media model

pdziewa:
Hi!

I just found a little typo in one property of Media model, while sending mails with `Model::shouldBeStrict(true)`.

Have a nice day! 👋


freekmurze:
Thank you!

PR #3085, opened 2022-11-07
Added option to set custom queue connection

rahulhaque:
Hello Spatie,

For convenience, I usually setup a custom queue connection that uses separate jobs table for heavy and high volume tasks like image processing. For this purpose, I've added a option to set up custom queue connection in the config and updated the documentation where necessary. I thought of adding some tests. However, Laravel itself validates empty string passed to its job's `onConnection` and `onQueue` methods. So, I didn't bother doing the same. All the other test cases passed except `create a derived version for an image without an extension` which was not related to this. I will have a look when possible. Let me know what you think of it and will update if any changes needed.

### DIFF
--- a/config/media-library.php
+++ b/config/media-library.php
@@ -15,6 +15,12 @@ return [
     'max_file_size' => 1024 * 1024 * 10, // 10MB
 
     /*
+     * This queue connection will be used to generate derived and responsive images.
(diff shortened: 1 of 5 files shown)


freekmurze:
Thanks!

PR #3091, opened 2022-11-09
fix #3090 - branch v9 PHP 7.4 version of the fix

### LINKED ISSUE #3090: MediaObserver updating fails if media file_name changes but path doesn't
As noted in the discussion https://github.com/spatie/laravel-medialibrary/discussions/3089 , if you set `moves_media_on_update` to true, and then update a media's file_name, but its generated path doesn't change the `MediaObserver` exceptions with:

```
League\Flysystem\FileExistsException
File already exists at path: <path>
```
on line 29 of the MediaObserver
```
        if (config('media-library.moves_media_on_update')) {
            $filesystem->syncMediaPath($media);            <---------- here
        }
```
This shouldn't be erroring out because a sync method shouldn't complain it can't move a file that doesn't need moving.

The example code that causes this to occur: 
```
        $avatar = $this->addMediaFromBase64($base64Avatar)

        $ext = MimeTypes::getDefault()->getExtensions($avatar->mime_type)[0] ?? null;
        $filename = pathinfo($avatar->file_name, PATHINFO_FILENAME);

        if ($ext) {
            $avatar ->file_name = "{$filename}.{$ext}";
            $avatar ->save();                             <--- crash at this point
        }
```

SlyDave:
Fixes syncMediaPath not checking paths have changed correctly.

Push to v9 branch to fix this for the version of medialibrary that supports PHP7.4

### DIFF
--- a/src/MediaCollections/Filesystem.php
+++ b/src/MediaCollections/Filesystem.php
@@ -245,7 +245,7 @@ class Filesystem
 
         $oldMedia = (clone $media)->fill($media->getOriginal());
 
-        if ($oldMedia->getPath() === $media->getPath()) {
+        if ($factory->getPath($oldMedia) === $factory->getPath($media)) {
             return;
         }
 


freekmurze:
Thanks!

PR #3098, opened 2022-11-17
[Micro Optimization] Use slice instead of reverse->take

Tofandel:
This does exactly the same but avoiding an array reverse


freekmurze:
Thanks!

PR #3109, opened 2022-11-23
Issue delete command once per unique directory

alisaleem27:
If a custom PathGenerator has been set which stores conversions and/or responsive images in the same directory, there is no point making the delete call to that directory multiple times.


freekmurze:
Thanks!

PR #3120, opened 2022-12-05
Allow to clean orphaned media when the `registerMediaConversionsUsingModelInstance` property is set to `true` on the model

tomcoonen:
This is a fix for the `media-library:clean` command.

Without the check if the model exists, the `$model->mediaConversions = []` would fail because `$model` would be `null`.

When `$model->registerMediaConversionsUsingModelInstance` is false, the `$media->model` check will not execute and not result in an extra query.


freekmurze:
Thank you!

PR #3135, opened 2022-12-23
guessing file extension from add streamed media.

jangidgirish:
guessing media extension name in `addMediaFromStream` and `addMediaFromString` methods.


freekmurze:
I'm going to pass on this for now.

PR #3137, opened 2022-12-26
Add findByUuidOrFail

francoism90:
It would be great if you could use this to fail whenever an UUID does not exists (anymore). :)

Please let me know if I need to add tests.


freekmurze:
Use a custom model, and put this method on your custom model.

PR #3139, opened 2022-12-30
Fix data corruption bug when storage disk becomes invalid

Pilskalns:
This fixes a bug when a storage disk becomes invalid. The [FileAdder.php#L450](https://github.com/spatie/laravel-medialibrary/blob/main/src/MediaCollections/FileAdder.php#L450) would "redo" already saved records in DB, unfortunately, the produced SQL delete query scopes to **all media belonging to a model** (not by id).

While `$model->media()->delete($media->id)` call looks ok, for some reason it actually ignores specific media id and scopes to all.

The was already a test case for this scenario, but it would run on the assumption there are no other media attached. So it would produce a false positive outcome.

Sample delete query **before**:
```
DELETE FROM `media`
WHERE `media`.`model_type` = 'App\\Models\\User' and `media`.`model_id` = 2 and `media`.`model_id` IS not NULL;
```

Sample delete query **after**:

```
DELETE FROM `media`
WHERE `id` = 16
```

### DIFF
--- a/src/MediaCollections/FileAdder.php
+++ b/src/MediaCollections/FileAdder.php
@@ -447,7 +447,7 @@ class FileAdder
         }
 
         if (! $addedMediaSuccessfully) {
-            $model->media()->delete($media->id);
+            $media->delete();
 
             throw DiskCannotBeAccessed::create($media->disk);
         }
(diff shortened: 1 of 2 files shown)


freekmurze:
Thanks!

PR #3142, opened 2023-01-01
Remove failed DB record even when using SoftDeletes

Pilskalns:
This is a tiny improvement on top of #3139. Basically, on the backtrack scenario, the failed DB record should be removed even when using SoftDeletes (i.e. custom Media model).

SoftDeletes are for _intended delete but keep in the trashcan scenario_. In this specific case, it is cleaning up after a failure.


freekmurze:

Pilskalns:
Ok, will do.

freekmurze:
Thanks!

PR #3143, opened 2023-01-04
Ability to use morphed map model with custom path generator

petyots:
This add the ability to to use morphed map model enforced by Eloquent enforce morph map
```php 
Relation::enforceMorphMap([
    'user' => User::class,       
 ]);
```

Now you can do:
```php
'custom_path_generators' => [
        'user' => \App\Domain\Media\Support\UserPathGenerator::class,
    ],
```


freekmurze:
Nice one, thanks!

PR #3144, opened 2023-01-04
Fix GitHub workflow status "test" badge

stevebauman:


freekmurze:
Thanks, we'll be fixing all our badges in bulk soon!

stevebauman:
Ah ok sounds good, thanks @freekmurze! 🙏

PR #3145, opened 2023-01-05
Prevent implicit conversion from float to int in FileSizeOptimizedWidthCalculator

QuentinGab:
Hi,

Super simple PR to prevent  implicit conversion from float to int and so, prevent a warning to be log.
Example of a warning:
> Implicit conversion from float 7616.5962599999975 to int loses precision in ...\vendor\spatie\laravel-medialibrary\src\ResponsiveImages\WidthCalculator\FileSizeOptimizedWidthCalculator.php on line 46


freekmurze:
Thanks!

boryn:
Good, it's finally solved, but IMHO solution presented here https://github.com/spatie/laravel-medialibrary/pull/3030 would be better - that is accepting float at `finishedCalculating()` instead of (int) casting the `$predictedFileSize`

PR #2907, opened 2022-05-04
Update create_media_table.php.stub according Laravel 9.x

erikn69:
https://github.com/spatie/laravel-medialibrary/blob/a1b16d72ea7ad07867b21e7c06b7a4eabdf5a72b/composer.json#L29-L33

PR #3149, opened 2023-01-06
fix: publish migration file only if it does not exist

Maggomann:
With the adaptation of the migration file in the pull request #2907, the check in the `MediaLibraryServiceProvider` whether the class `CreateMediaTable` already exists is no longer effective, because the class is omitted in the above pull request, so that a new migration file is always created when the publication is executed, regardless of whether a migration file already exists.

This pull request now provides a method that checks whether the `CreateMediaTable` class exists or whether a migration with the string `_create_media_table` exists in the migration directory. If either case occurs, a new migration file is not created.  

The error occurred when testing a private package and using the `Illuminate\Foundation\Testing\LazilyRefreshDatabase` trait.

### DIFF
--- a/src/MediaLibraryServiceProvider.php
+++ b/src/MediaLibraryServiceProvider.php
@@ -3,6 +3,7 @@
 namespace Spatie\MediaLibrary;
 
 use Illuminate\Support\ServiceProvider;
+use Illuminate\Support\Str;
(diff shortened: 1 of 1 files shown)


Maggomann:
I'm closing this again. I think the problem sits 30 cm in front of the screen :-P

PR #3156, opened 2023-01-12
feat: add uuid phpdoc block on Media class

kyryl-bogach:
This will allow clients of this tool be able to infer this property and its type.


freekmurze:
Thanks!

PR #3164, opened 2023-01-18
Method for reversing migrations

eimantaaas:
Adding `down()` method to migrations stub


freekmurze:
We don't use `down` migrations in any of our projects. Add this in the migration you publish.

PR #3187, opened 2023-02-12
update config instructions for custom_path_generators

lukas-pierce:
It is not obvious that I need to use a string if I have declared `Relation::morphMap([...])`


freekmurze:
Thanks!

PR #3191, opened 2023-02-21
Fixed custom path generators for Media with morphed map relations

aprokopenko:
Hi, we already reported this issue for v9 version and created the Pull Request for v9. However, you asked then to create the PR for v10 first.

Finally, I had time to do that.

**Issue:**
When using custom path generators for models and Media model has morph map strings, not Model names it doesn't work.

Furthermore, I found the related discussions and I guess you had similar PR in the past:
https://github.com/spatie/laravel-medialibrary/discussions/3034

### DIFF
--- a/src/Support/PathGenerator/PathGeneratorFactory.php
+++ b/src/Support/PathGenerator/PathGeneratorFactory.php
@@ -22,7 +22,12 @@ class PathGeneratorFactory
         $defaultPathGeneratorClass = config('media-library.path_generator');
 
         foreach (config('media-library.custom_path_generators', []) as $modelClass => $customPathGeneratorClass) {
-            if (is_a($media->model_type, $modelClass, true)
+            if (
+                // model doesn't have morphMap
+                is_a($media->model_type, $modelClass, true)
+                // config is set via morphMap alias
+                || $media->model_type === $modelClass
+                // config is set via morphMap class name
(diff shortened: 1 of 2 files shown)


aprokopenko:
I fixed unit tests and add a new one, to the option case, which this PR is a fix for

src/Support/PathGenerator/PathGeneratorFactory.php
```diff
@@ -21,7 +22,14 @@ protected static function getPathGeneratorClass(Media $media)
         $defaultPathGeneratorClass = config('media-library.path_generator');
 
         foreach (config('media-library.custom_path_generators', []) as $modelClass => $customPathGeneratorClass) {
-            if (is_a($media->model_type, $modelClass, true) || $media->model_type === $modelClass) {
+            if (
+                // model doesn't have morphMap
+                is_a($media->model_type, $modelClass, true)
```
freekmurze, line 27: Could you extra these conditionals to a function that used early returns. Give the new function a nice name.

freekmurze (changes requested):

src/Support/PathGenerator/PathGeneratorFactory.php
aprokopenko: @freekmurze , I did it, could you check please is the name is okay to you?

freekmurze:
Thanks!

PR #3192, opened 2023-02-21
Include media object as optional param

613-msechrest:
Adding the media object as a parameter to the convert method.
- This allows us to utilize the media objects parameters to leverage how the media should be processed from a custom image manipulator.

Current TODOs
- Rather than forcing the media object to be included in the convert method, write a way for it to be an optional parameter.
-     Idea - Create a new Conversion method that allows us to include the media object for only specified conversions. For example how the withResponsiveImages() (or any other options that can be conversion specific) method is defined.


613-msechrest:
Apologies, meant to submit this PR to my local fork. Please ignore.

PR #3204, opened 2023-03-08
fix: widen version constraint for illuminate/conditionable

christian98:
The PR (https://github.com/spatie/laravel-medialibrary/pull/3162) introduced `illuminate/conditionable` as a dependency, but with a small missing character.

The tight version prevented us from upgrading  `spatie/laravel-medialibrary`.


christian98:
I think this is now done by 4fabfc58e08ee53586a613b18b8879049132e526. So, closing this.

PR #3219, opened 2023-03-28
Set model relation on InteractsWithMedia@loadMedia()

cosmastech:
If ever media needs to reference its model (for instance, within a Resource), it will have to make a separate DB query to get the relationship (and this is especially annoying if you disable lazy loading).

When `loadMedia()` is called, it will set the model property for all Media within the Collection.


freekmurze:
Very neat, thank you!

PR #3220, opened 2023-03-28
Limit which media is loaded and hydrated

cosmastech:
Thank you for this wonderful package. We leverage it for a product I work on where we have over 25mil rows in our media table.

We do a good job of segregating models and collections so that there aren't too many records for a given model, but even still, if we are just loading a thumbnail for a gallery preview or paginating a response, MediaLibrary is hydrating all of the media models just to filter by the collection_name.

I don't expect this PR to be accepted in its current state, as it's breaking some tests, but more wanted to start a conversation around how the library could account for this (or how I could better utilize the library to solve for the problems we are seeing).


freekmurze:
As this is not fully completed, I'm not going to merge it, but I'll keep in mind that we could improve on this when creating a new major version of the package.

Issue #3224, opened 2023-03-29
Release 10.7.14: 502 Bad Gateway


SebastianSchoeps:
I receive a very strange 502 Bad Gateway that I cannot explain after upgrading to 10.7.14 when calling `getFirstMediaUrl`, but it seems to be a bit random when it appears. I don't know how to provide more information as I cannot really isolate the issue - I could present it in a call if you are interested. It seems to happen in the very basic `Arr::first` method of Laravel.
It happens across different systems.

For now I will stick to release 10.7.13. This issue is more of an information to others that have the same problem.

freekmurze:
I'm not seeing the error, it would be very handy if any of you would share the actual error message + stack trace here.

steveallinson:
Same here. Getting 502 errors when calling getMedia/getFirstMediaUrl in certain parts of our app. Broken since updating to 10.7.14

HolgerHatGarKeineNode:
```
2023/03/29 09:29:51 [error] 3404896#3404896: *108034 recv() failed (104: Connection reset by peer) while reading response header from upstream, client: X.X.X.X, server: xx.xx.xx, request: "GET /xx HTTP/2.0", upstream: "fastcgi://unix:/var/run/php/php8.2-fpm-einundzwanzig.sock:", host: "portal.einundzwanzig.space", referrer: "https://portal.einundzwanzig.space/news?l=de"
```

Nginx is loosing connection to php completely. With Laravel Sail all browsers loosing connection immediately.

Somehow it's destroying the php process. Searching logs for more error logs now.

freekmurze:
I've just release 10.7.15 that reverts the changes of 10.7.14

In that previous version a potential N+1 probably was solved by explicitly setting a related model on a `Media` object. My hunch is that there's something going wrong when Laravel / PHP tries to serialise such a media object when queuing stuff. In our tests we don't queue stuff, and that might be the reason why our testsuite didn't catch it.

My apologies for any inconvenience. Do let me know if 10.7.15 works for you.

SebastianSchoeps:
Works for me. Thanks very much for the very fast reaction!

cosmastech:
Could you help me set up a test case for this? Which Laravel version are you using, is this from a web request or is there possibly dispatching a job?

SebastianSchoeps:
Laravel: 10.5.0
Livewire: 2.12.3

It fails at a very simple Livewire component. I stripped everything out of it and it still fails.
[...]

PR #3226, opened 2023-03-29
Set Model Relationship && Fix serialization

cosmastech:
It looks like serialization/deserialization never would've worked if media was being added to a collection (without first refreshing the model).

The properties set by the InteractsWithMedia trait contain closures.

This MR add back in the set relation functionality, along with a fix for InteractsWithMedia@__sleep() so that those properties are not getting serialized. This should save on data pushed to the queue as well. 😄


cosmastech:
https://github.com/spatie/laravel-medialibrary/issues/3224#issuecomment-1488950912

Thanks to @HolgerHatGarKeineNode for testing this.

@freekmurze let me know if there's anything else you'd like to see done with this PR to confirm it's working properly

cosmastech:
Closing this. The serialization issue is addressed https://github.com/spatie/laravel-medialibrary/pull/3258

michaelbaril:
Hi @cosmastech , I was about to create a similar PR and then I stumbled upon yours. I don't understand why you closed it though? The other PR you're referring to addresses the serialization issue, but doesn't set the model relation (which is what I'm interested in). 
Thanks!

cosmastech:
> Hi @cosmastech , I was about to create a similar PR and then I stumbled upon yours. I don't understand why you closed it though? The other PR you're referring to addresses the serialization issue, but doesn't set the model relation (which is what I'm interested in). Thanks!

Hello @michaelbaril

If you're interested in adding this functionality, you can extend the MediaRepository class and bind the new implementation into the container.

This is untested, but something like:

```php
namespace App;


use Illuminate\Support\Collection;
use Spatie\MediaLibrary\MediaCollections\MediaRepository;

class MyMediaRepository extends MediaRepository
{
    public function getCollection(
        HasMedia $model,
        string $collectionName,
        array|callable $filter = []
    ): Collection {
        return $this->applyFilterToMediaCollection($model->loadMedia($collectionName), $filter)
            ->each(fn(Model&HasMedia $media) => $media->setRelation('model', $model));
    }
}
```

The binding I'm less sure about. I don't think you could put the binding within AppServiceProvider, since the MediaLibraryServiceProvider is loaded afterwards. 

You could extend the package's [MediaLibraryServiceProvider](https://github.com/spatie/laravel-medialibrary/blob/main/src/MediaLibraryServiceProvider.php) and override the register method so it's relying on your custom MyMediaRepository.

---

Another option would be to create your own InteractsWithMedia trait.

```php
namespace App\Concerns;

use Spatie\MediaLibrary\MediaCollections\MediaCollection;

trait InteractsWithMediaTrait
{
    use InteractsWithMedia {
        getMedia as baseGetMedia;
    }
    
    public function getMedia(string $collectionName = 'default', array $filters = []): MediaCollection
    {
        return $this->baseGetMedia($collectionName, $filters)
            ->each(fn(Model&HasMedia $media) => $media->setRelation('model', $this));
    }
}
```

Then replace your models which employ `InteractsWithMedia` to use `InteractsWithMediaTrait` instead.
---

As for the reason I closed it... I didn't assume it was going to get merged. It broke the package for a handful of people, so I assume @freekmurze would be less inclined to merge this change in again, even after fixing the serialization issue. The benefit it provides is only if the client code is fetching the media's model, which can be worked around other ways.

Another thought I had just now is that I'm not sure how it might work in an environment like Octane tbh. Would hate for this to lead to memory leaks.

Feel free to open your own PR for this of course!

michaelbaril:
Hey @cosmastech , thanks for your fast and detailed answer! Actually I already did it using the 2nd approach you suggest (overriding the trait), but I still thought of doing a PR because I felt it should be the standard behavior. When using `$registerMediaConversionsUsingModelInstance = true`, it avoids unnecessary queries.

Anyway, I'll probably open my own PR and see what the Spatie guys think.

Thanks again!

PR #3227, opened 2023-03-30
Only offer publishing when running in console

erikn69:
https://github.com/spatie/laravel-package-tools/blob/d7ea4f74ffe1f7863f01bf5ba4b6d64847dacf8b/src/PackageServiceProvider.php#L58-L69


freekmurze:
Thanks! I'd also accept a PR that would use our package service provider.

PR #3228, opened 2023-03-30
Avoid duplicates on migration files

erikn69:
After anonymous migrations `CreateMediaTable` always does not exist, this duplicate media migration file

### DIFF
--- a/src/MediaLibraryServiceProvider.php
+++ b/src/MediaLibraryServiceProvider.php
@@ -2,6 +2,8 @@
 
 namespace Spatie\MediaLibrary;
 
+use Illuminate\Filesystem\Filesystem as IlluminateFilesystem;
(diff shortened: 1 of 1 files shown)


freekmurze:
For most people the current code works, so I'm going to leave it like it is for now.

erikn69:
@freekmurze actually i was the one who broke this functionality when i did upgrade to anonimous migrations on #2907
Also, your package service provider do that too, look 
https://github.com/spatie/laravel-package-tools/blob/d7ea4f74ffe1f7863f01bf5ba4b6d64847dacf8b/src/PackageServiceProvider.php#L167-L173

erikn69:
>Thanks! I'd also accept a PR that would use our package service provider.

Maybe better that, I'll try

PR #3235, opened 2023-04-06
Allow custom responsive url

francoism90:
I would like to customize the responsive image URL. Currently, it is only possible to use a full directory root, however I would like use something like this:
```php
<?php

namespace Support\MediaLibrary\UrlGenerator;

use DateTimeInterface;
use Illuminate\Support\Facades\URL;
use Illuminate\Support\Str;
use Spatie\MediaLibrary\Support\UrlGenerator\BaseUrlGenerator;

class DefaultUrlGenerator extends BaseUrlGenerator
{
    public function getUrl(): string
    {
        return URL::signedRoute(
            'api.media.show',
            [
                'media' => $this->media,
                'conversion' => $this->conversion?->getName(),
                'version' => $this->media?->updated_at?->timestamp,
            ]
        );
    }

    {
        return URL::temporarySignedRoute(
            'api.media.show',
            $expiration,
            [
                'media' => $this->media,
                'conversion' => $this->conversion?->getName(),
                'version' => $this->media?->updated_at?->timestamp,
            ]
        );
    }

    public function getResponsiveImageUrl(string $fileName): string
    {
        return URL::signedRoute(
            'api.media.responsive',
            [
                'media' => $this->media,
                'conversion' => $fileName, // offers route completion as well
                'version' => $this->media?->updated_at?->timestamp, // allows passing version (or any other custom route args)
            ]
        );
    }

    public function getBaseMediaDirectoryUrl(): string
    {
        return $this->getDisk()->url('/');
    }

    {
        return $this->getRootOfDisk().$this->getPathRelativeToRoot();
    }

    protected function getRootOfDisk(): string
    {
        return $this->getDisk()->path('/');
    }
}
```

Is it possible to improve the test as well? It would be great if you could use `->withResponsiveImages()` on a test model, and validate if the image has been created. :)

Thanks!

### DIFF
--- a/src/ResponsiveImages/ResponsiveImage.php
+++ b/src/ResponsiveImages/ResponsiveImage.php
@@ -45,7 +45,7 @@ class ResponsiveImage
 
         $urlGenerator = UrlGeneratorFactory::createForMedia($this->media, $conversionName);
 
-        return $urlGenerator->getResponsiveImagesDirectoryUrl().rawurlencode($this->fileName);
+        return $urlGenerator->getResponsiveImageUrl($this->fileName);
     }
 
     public function generatedFor(): string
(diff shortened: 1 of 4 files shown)


francoism90:
@freekmurze Could you please take a look? :)

If tests aren't possible, I'lll remove it. It would be great, as it helps generating custom URLs.

freekmurze:
I'm going to pass on this for now, because it might be breaking for some, but I'll revisit when creating a new major version of the package.

Issue #3081, opened 2022-11-04
WebP responsive images


veneliniliev:
when a **webp** image is added no **wepb** responsive images are generated for `media_library_original`

the problem may be here:
`if ($this->generateResponsiveImages && (new ImageGenerator())->canConvert($media)) {`

`use Spatie\MediaLibrary\Conversions\ImageGenerators\Image as ImageGenerator;`
**ImageGenerator** instance always is an `Image` conversion and returns `false` for **webp** images

is there a way to have responsive images in **webp** format?

for the moment I run the job (_GenerateResponsiveImagesJob_) manually but it is not a good solution :(

patinthehat:
@veneliniliev Would it be possible to provide a failing unit test to show that this is in fact a bug, or possibly a sample project/repository that demonstrates the issue described? Thanks!

yannkuesthardt:
Can confirm this issue. In the file "FileAdder.php" (Spatie\MediaLibrary\MediaCollections) in the function "processMediaItem" the following if statement is used to check if the file can be made responsive (this is used whenever "->toMediaCollection()" is called on a new image). 

`if ($this->generateResponsiveImages && (new ImageGenerator())->canConvert($media)) {`

The problem is, that ImageGenerator is a renamed import for
`Spatie\MediaLibrary\Conversions\ImageGenerators\Image`
which will of course return false when passing a **.webp** file to
 `canConvert()`
 
 My suggestion for a simple fix is changing the if statement to the following
 `if ($this->generateResponsiveImages && ((new ImageGenerator())->canConvert($media) || (new WebpGenerator())->canConvert($media))) {`
 where "WebpGenerator" is a renamed import for
 `use Spatie\MediaLibrary\Conversions\ImageGenerators\Webp`

freekmurze:
Could you PR that change and add a test to prove that it works?

yannkuesthardt:
I will do that, but I won’t have time till mid/end January

stuartcusackie:
Also seeing this issue. By the way, manually running he GenerateResponsiveImagesJob doesn't work for me. Responsive variants are generated but they are using the original format of the image.

websitevirtuoso:
I also have problem with this.
![image](https://user-images.githubusercontent.com/5112918/230823685-64be4a81-f225-417d-b3cd-d72e3708c6ae.png)

Last image it webp and lib couldn't create responsive images

websitevirtuoso:
[...]

PR #3242, opened 2023-04-17
Add webp extension. Fix: files with webp doesn't process responsive images

websitevirtuoso:
Fix for 
https://github.com/spatie/laravel-medialibrary/issues/3081


freekmurze:
Thanks!

PR #3243, opened 2023-04-17
Optimisation for setNewOrder

linaspasv:
Reordering on larger collections are quote slow because we are iterating over each media item and updating them individually. My PR solves the performance issues by introducing the CASE statement to update multiple media items in a single query.

I see around 6-7x speed improvement when reordering collections with 30 media items, so there should be greater improvements with even larger collections. At the moment setNewOrder executes around 60 queries (SELECT, UPDATE) for 30 media items, and after the PR it should be reduced to a single query for the update.


freekmurze:
I'm going to leave it like it is for now, as people might rely on the model events that are being fired now.

PR #3245, opened 2023-04-18
Simplify the supportedMimeTypes

niladam:
This PR is just a simplification of the supportedMimeTypes version.

It makes use of the already built supportedExtensions, and simply rejects the `jpg` since it doesn't have a mime-type (only `image/jpeg` being valid) and transforms it to the same format (`image/jpeg`, `image/png`..etc)


freekmurze:
I'm going to leave it like it is for now. Thanks!

PR #3248, opened 2023-04-20
feat: add user agent to stream context in DefaultDownloader

rolfvandekrol:
Some random image sources (including https://picsum.photos ) refuse to serve images when there is no user agent header in the request. See https://laracasts.com/discuss/channels/laravel/why-spatielaravel-medialibrary-raise-error-on-uploding-image-uploaded-with-faker-images for more context.

This PR adds a user agent to all requests in the `DefaultDownloader` to solve this problem.


freekmurze:
Thanks!

PR #3252, opened 2023-04-22
Feature/new custom delete fix #3190 

### LINKED ISSUE #3190: Losing images when using custom media path generator
Using a custom path generator, I have implemented a storage structure to match a structure I was already using.

So my media is saved (in s3) as 

```[BUCKET]/media/2023/02/new-red-car.jpg```

and conversions

```[BUCKET]/media/2023/02/c/new-red-car-large.jpg```
```[BUCKET]/media/2023/02/c/new-red-car-medium.jpg```
```[BUCKET]/media/2023/02/c/new-red-car-small.jpg```

The problem here is that, due to the way the media delete observer works, it uses getPath() and getConversionPath() etc and simply removes those base folders.

The expectation would be that the ```removeAllFiles``` functionality would remove the image only, and its conversions only. 

In reality, the ```media/2023/02/``` folder is being deleted in its entirety - meaning it just wipes a month worth of other media unrelated to the item that was deleted.

### PR DESCRIPTION
**Pull request by benjam-es**
This adds a test, using a custom path generator which uses a shared folder (i.e. not individual /1/ folder per image)

Once the test was failing, I extracted out the 'removeFile' and 'removeAllFiles' logic into a FileRemover class using a FileRemover Factory, to work in the same way you currently swap out path generators.

I then added a new FileBasedFileRemover class, which fixes the failing test created to demonstrate #3190.

This PR allows:
 - File removal logic can be swapped out
 - Existing file removal logic is maintaned in a DefaultFileRemoval class - ensuring everything works as before
 - New optional FileBasedFileRemover class allow us to use a custom directory path generator, which *does not* delete other files with is
 
 These are split into 3 commits to see each part.

### DIFF
--- a/src/Support/FileRemover/FileBaseFileRemover.php
+++ b/src/Support/FileRemover/FileBaseFileRemover.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace Spatie\MediaLibrary\Support\FileRemover;
+
+use Illuminate\Contracts\Filesystem\Factory;
+use Spatie\MediaLibrary\MediaCollections\Filesystem;
+use Spatie\MediaLibrary\MediaCollections\Models\Media;
+use Spatie\MediaLibrary\Support\FileRemover\FileRemover;
+use Spatie\MediaLibrary\Support\FileRemover\DefaultFileRemover;
+
+class FileBaseFileRemover extends DefaultFileRemover implements FileRemover
+{
+    public function __construct(private Filesystem $mediaFileSystem, private Factory $filesystem)
+    {}
+
+    public function removeAllFiles(Media $media): void
+    {
+        $this->removeFile($this->mediaFileSystem->getMediaDirectory($media). $media->file_name, $media->disk);
+
+        $this->removeConvertedImages($media);
+    }
+
+    public function removeConvertedImages(Media $media): void
+    {
+        $conversionsDirectory = $this->mediaFileSystem->getConversionDirectory($media);
+        $mediaDirectory = $this->mediaFileSystem->getMediaDirectory($media);
+
+        collect($media->getMediaConversionNames())->each(function ($conversionName) use ($media) {
+            $this->removeFile(
+                path: $media->getPathRelativeToRoot($conversionName),
+                disk: $media->conversions_disk
+            );
+
+            $this->mediaFileSystem->removeResponsiveImages($media, $conversionName);
+        });
+    }
+
+    public function removeFile(string $path, string $disk): void
+    {
+        $this->filesystem->disk($disk)->delete($path);
+    }
+}
--- a/src/Support/FileRemover/FileBaseFileRemover.php
+++ b/src/Support/FileRemover/FileBaseFileRemover.php
@@ -10,7 +10,7 @@ use Spatie\MediaLibrary\Support\FileRemover\DefaultFileRemover;
 
 class FileBaseFileRemover extends DefaultFileRemover implements FileRemover
 {
-    public function __construct(private Filesystem $mediaFileSystem, private Factory $filesystem)
+    public function __construct(protected Filesystem $mediaFileSystem, protected Factory $filesystem)
     {}
 
     public function removeAllFiles(Media $media): void
--- a/src/Support/FileRemover/FileBaseFileRemover.php
+++ b/src/Support/FileRemover/FileBaseFileRemover.php
@@ -22,9 +22,6 @@ class FileBaseFileRemover extends DefaultFileRemover implements FileRemover
 
     public function removeConvertedImages(Media $media): void
     {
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: src/Support/FileRemover/FileRemover.php [hunk 1]
```diff
@@ -0,0 +1,31 @@
...
+
+    /*
+     * Remove all converted files relating to the media model.
+     */
+    public function removeResponsiveImages(Media $media, string $conversionName): void;
+
+    /*
+     * Remove all responsive image files relating to the media model.
+     */
+    public function removeFile(string $path, string $disk): void;
```
freekmurze, line 26: This name is very confusing. Shouldn't this be named `removeResponsiveImages`?

src/Support/FileRemover/FileRemover.php
```diff
@@ -0,0 +1,31 @@
...
+
+    /*
+     * Remove all files relating to the media model.
+     */
+    public function removeAllFiles(Media $media): void;
+
+    /*
+     * Remove all converted files relating to the media model.
+     */
+    public function removeResponsiveImages(Media $media, string $conversionName): void;
```
freekmurze, line 21: Shouldn't this me named `removeConvertedFiles`?

src/Support/FileRemover/FileRemover.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    /*
+     * Create a new file remover instance, using the media library file system and illuminate filesystem.
+     */
+    public function __construct(Filesystem $mediaFileSystem, Factory $filesystem);
+
+    /*
+     * Remove all files relating to the media model.
+     */
+    public function removeAllFiles(Media $media): void;
```
freekmurze, line 16: Rename this to `removeOriginalFiles`

freekmurze (changes requested):

src/Support/FileRemover/FileRemover.php
benjam-es: This name matches the existing method name inside the Filesystem class 
```php
public function removeAllFiles(Media $media): void
```

src/Support/FileRemover/FileRemover.php
benjam-es: This matches the existing method from within the Filesystem class
```php
public function removeResponsiveImages(Media $media, string $conversionName = 'media_library_original'): void
```

Fixed the comments, which were incorrect

src/Support/FileRemover/FileRemover.php
benjam-es: Fixed the code comments - copy and paste mistake :)

Log1x:
This will be great. I was extremely confused why my entire media library was disappearing... 😩

elmudometal:
It will take a long time for this to be solved?

benjam-es:
@freekmurze I believe the requests were addressed in April. Let me know if you need anything else.

freekmurze:
This seems pretty nice! 

Could you also write docs on this?

src/Support/FileRemover/FileBaseFileRemover.php
```diff
@@ -0,0 +1,42 @@
...
+    public function removeAllFiles(Media $media): void
+    {
+        $this->removeFile($this->mediaFileSystem->getMediaDirectory($media). $media->file_name, $media->disk);
+
+        $this->removeConvertedImages($media);
+    }
+
+    public function removeConvertedImages(Media $media): void
+    {
+        $conversionsDirectory = $this->mediaFileSystem->getConversionDirectory($media);
```
marijnhurkens, line 25: These ars are unused right?

src/Support/FileRemover/FileBaseFileRemover.php
```diff
@@ -0,0 +1,42 @@
...
+        $mediaDirectory = $this->mediaFileSystem->getMediaDirectory($media);
+
+        collect($media->getMediaConversionNames())->each(function ($conversionName) use ($media) {
```
marijnhurkens, line 28: Shouldn't this use `getGeneratedConversions()` instead of `getMediaConversionNames()`?

benjam-es:
Awaiting @freekmurze approval I believe. I am happy to write docs on it

src/Support/FileRemover/FileBaseFileRemover.php
benjam-es: From memory, names are used to get to the specific file names individually

benjam-es:
@freekmurze I have added a new docs MD file to detail the new functionality

benjam-es:
It looks like these tests are failing due to this: https://github.com/spatie/laravel-medialibrary/issues/3417

freekmurze:
Could you rebase with `main`? That should make the tests pass.

freekmurze:
Thanks!

PR #3255, opened 2023-04-25
Update installation-setup.md

PatryQHyper:
Specifying version will return requirements problem:

Problem 1
 - Root composer.json requires spatie/laravel-medialibrary 10.0.0 -> satisfiable by spatie/laravel-medialibrary[10.0.0].
 - spatie/laravel-medialibrary 10.0.0 requires illuminate/bus ^9.0 -> found illuminate/bus[v9.0.0, ..., v9.52.7] but these were not loaded, likely because it conflicts with another require.

### DIFF
--- a/docs/installation-setup.md
+++ b/docs/installation-setup.md
@@ -8,7 +8,7 @@ Media Library can be installed via Composer:
 If you only use the base package issue this command:
 
 ```bash
-composer require "spatie/laravel-medialibrary:^10.0.0"
+composer require spatie/laravel-medialibrary
 ```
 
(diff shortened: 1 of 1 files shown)


PatryQHyper:
Without specifying version, it just installs fine

freekmurze:
I've just tested this, and the current command works just fine. I'm thinking it probably doesn't work for you because you don't meet the requirements of v10, composer will probably just pull in an older version that works for your environment.

PatryQHyper:
10.9.1 is version that is installed without specified command. My env must be fine, because it's a clean Laravel project

freekmurze:
I don't know why it's not working at your end 😬

jonhassall:
Same issue on a clean install of Laravel on Windows.

PR #3258, opened 2023-05-02
Fix InteractsWithMedia serialization

cosmastech:
Fixes `Exception : Serialization of 'Closure' is not allowed` when attempting to serialize a Model that employs InteractsWithMedia.


freekmurze:
Thanks!

PR #3286, opened 2023-06-12
feat: add support for * ( ANY ) media collection

huzaifaarain:
There should be an option to fetch all associated media records regardless of the collection they are associated with.

We can pass the asterisk or "star" symbol (*)  to function for loading all media that belongs to the caller model.
 
`$model->getMedia('*')`

### DIFF
--- a/docs/basic-usage/retrieving-media.md
+++ b/docs/basic-usage/retrieving-media.md
@@ -9,6 +9,12 @@ To retrieve files you can use the `getMedia`-method:
 $mediaItems = $yourModel->getMedia();
 ```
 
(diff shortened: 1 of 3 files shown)


freekmurze:
Could you add tests for this and mention this in the documentation?

huzaifaarain:
> Could you add tests for this and mention this in the documentation?

Sure.

huzaifaarain:
> Could you add tests for this and mention this in the documentation?

Hi @freekmurze 

Please review the PR, I have added the docs and the test.

freekmurze:
Thanks!

PR #3289, opened 2023-06-14
feat: add file name to addMediaFromString method

hmreumann:
This PR adds the ability to set a specific file name instead of the default "text.txt" when adding a media from a string

```
$media = $model->addMediaFromString($imageContent, 'profile_photo.jpg')->toMediaCollection('image');
```

### DIFF
--- a/src/InteractsWithMedia.php
+++ b/src/InteractsWithMedia.php
@@ -170,7 +170,7 @@ trait InteractsWithMedia
      *
      * @param string string
      */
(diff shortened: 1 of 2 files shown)


freekmurze:
I think you can already achieve this with:

```php
 $model
   ->addMediaFromString($imageContent)
   ->usingFileName('profile_photo.jpg')
   ->toMediaCollection('image');
```

PR #3299, opened 2023-06-29
FIX create migration

wamesro:
After running the php artisan `vendor:publish command ...`, the new migration file was still being created

the migration template was changed to an anonymous class without a classname


freekmurze:
Thanks!

Issue #3282, opened 2023-06-07
Regenerating images does not update the Media model


joostvanhoof:
Laravel: `10.13.2`
MediaLibrary: `10.9.4`
PHP: `8.2`

In a model, in the `registerMediaConversions` function I changed:

```php
$this->addMediaConversion('large')
  ->keepOriginalImageFormat()
  ->performOnCollections('header')
  ->width(1400);
```

to this:

```php
$this->addMediaConversion('large')
  ->format('webp') // this was updated
  ->performOnCollections('header')
  ->width(1400);
```

If I then run:

`php artisan media-library:regenerate`

The new webp images are generated correctly. However: the Media model does not seem to be updated. The `mime_type` column still says `image/jpeg` and the `updated_at` datetime stamp is also not updated.

The functions to display the image in the blade file, like `getFirstMediaUrl` etc, all do work as expected. 

This seems like a bug to me as the model/database information and the actual files are now out of sync.

The other issue I see is that if after this regeneration I want to run `media-library:clean`, it would not know to delete the old jpegs/pngs and keep the webp images.

freekmurze:
I'd accept a PR that fixes this in the regenerate command 👍

davidoskay:
Same after making some manipulations on image:
`                $media = $estate->addMedia($file)->toMediaCollection('gallery');
                Image::load($media->getPath())
                    ->fit(fit: Fit::Max, desiredWidth: 1920, desiredHeight: 1080)
                    ->quality(85)
                    ->save();`

Now I wanted to get filesize but it is also wrong not refreshed.

PR #3316, opened 2023-07-15
Always set the updated_at column when marking conversions are finalized

cosmastech:
Resolves part of the issue identified in https://github.com/spatie/laravel-medialibrary/issues/3282


freekmurze:
Thanks!

PR #3319, opened 2023-07-21
Automatically set model relationship on media

michaelbaril:
The purpose of this PR is to automatically set the `model` relationship on the Media model. This avoids unnecessary queries, especially when [`$registerMediaConversionsUsingModelInstance = true`](https://spatie.be/docs/laravel-medialibrary/v10/converting-images/defining-conversions#content-using-model-properties-in-a-conversion) is used.

@cosmastech already created [a similar PR](https://github.com/spatie/laravel-medialibrary/pull/3219) a while ago. It was merged, but then rolled back because of issues with serialization.

Since then, @cosmastech has fixed the serialization issue in [this PR](https://github.com/spatie/laravel-medialibrary/pull/3258).

### DIFF
--- a/src/InteractsWithMedia.php
+++ b/src/InteractsWithMedia.php
@@ -546,8 +546,11 @@ trait InteractsWithMedia
 
         $collection = new MediaCollections\Models\Collections\MediaCollection($collection);
 
+        $modelWithoutMedia = (clone $this)->unsetRelation('media');
+
         return $collection
             ->filter(fn (Media $mediaItem) => $collectionName !== '*' ? $mediaItem->collection_name === $collectionName : true)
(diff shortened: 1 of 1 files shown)


freekmurze:
Thank you, hopefully no problems this time around 👍

keatliang2005:
This cause infinite loop when model have attribute append, then serialize on response

```php
class Example extends Model implements HasMedia
{
    use InteractsWithMedia;

    protected function exampleImageUrl() : Attribute
    {
        $file = $this->getMedia('example_image_url')->last();

        return Attribute::make(
            get: fn() => $file ? ['url' => $file->getUrl()] : null
        );
    }
}
```

```php
$model = Example::find(1);
$model->append('example_image_url');
return response->json($model);
```

PR #3320, opened 2023-07-22
added AVIF support

0xb4lint:
- AVIF support based on https://github.com/spatie/image-optimizer/pull/198
- AVIF image generator
- AVIF test


freekmurze:
Try requiring the latest version of the image package in composer.json. That should fix the tests

freekmurze:
I quickly took care of this myself thanks!

ariaieboy:
Thank you so much for this PR @0xb4lint 
I have a quick question for you did you test this new optimizer method in production?
because I've been using Avif format for a while with spatie media library but the generated avif format was much bigger than Webp and sometimes bigger than the source image. 
Is the generated avif files smaller than webp files?

0xb4lint:
I've noticed the same, that's why I made the AVIF optimizer first: https://github.com/spatie/image-optimizer/pull/198
Based on my research in this topic: https://github.com/feat-agency/avif

The main problem with the large file size coming from the Intervention Image's default 90 quality setting, which results a much higher AVIF than WEBP. 

So either you decrease the AVIF conversion quality or use the AVIF optimizer, which will convert to a lot smaller file, while respecting the DSSIM value I've researched.

PR #3326, opened 2023-07-26
Handle null extensions and mimetypes with ImageGeneratorFactory

cosmastech:
In our test suite @inflcr, some of our factories were lacking mime_types and filenames.

This causes and exception to be thrown when attempting to access the Media attribute accessor `type()` (because `ImageGenerator@canHandleExtension()` and `ImageGenerator@canHandleMime()` both require a string).

I have fixed that in this PR by exiting early in their respective factory methods when null is passed in.


freekmurze:
Thanks!

PR #3341, opened 2023-08-10
PhpStan Workflow

erikn69:
[laravel-medialibrary/actions/runs/phpstan.yml](https://github.com/erikn69/laravel-medialibrary/actions/runs/5823518168/job/15790583424)


freekmurze (changes requested):
Seems very nice!

config/media-library.php
```diff
@@ -48,6 +48,7 @@
      * The fully qualified class name of the model used for temporary uploads.
      *
      * This model is only used in Media Library Pro (https://medialibrary.pro)
+     * @phpstan-ignore-next-line
```
freekmurze, line 51: I don't like `@phpstan-ignore-next-line` in a config file, could you add the issues that this generates in the phpstan base line.

Also add all other issues that you're now solving with `@phpstan-ignore-next-line` to the baseline 👍

freekmurze (changes requested):

PR #3347, opened 2023-08-21
Fixed error when config is cached

Roboroads:
This is one I came across when deploying to production after adding media library. I had an error because my config was cached, media-library wasn't in the cache yet, I coudn't config:clear or config:cache because you cannot run `null::observe()`, which kept the app crashing before the command could be run.

Having a sensible default fixes this issue. (or manually `rm bootstrap/cache/config.php`, which was my solution for now)


freekmurze:
Thanks!

PR #3357, opened 2023-08-28
Adds `FileAdder@setFileSize()`

cosmastech:
This PR makes it possible to set the filesize via FileAdder. My thinking is that we may already have this information (such as from the headers of a request) and having to either call `filesize()` or `Storage::size()` can be avoided.

Also, I refactored it a bit so file size is only fetched one time.

### DIFF
--- a/src/MediaCollections/FileAdder.php
+++ b/src/MediaCollections/FileAdder.php
@@ -48,6 +48,8 @@ class FileAdder
 
     protected string $diskName = '';
 
+    protected ?int $fileSize = null;
+
     protected string $conversionsDiskName = '';
 
     protected ?Closure $fileNameSanitizer;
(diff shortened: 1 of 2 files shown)


cosmastech:
After diving deeper into this, I don't think we need the setFileSize method.

However, I do think the refactor to only determine the filesize once is beneficial, as it cuts down on a repeated call to S3.

freekmurze:
Thanks!

francoism90:
Why would you want to do this? What if the file size differs when the upload has been finished?

Doing the extra call, would help debugging these issues.

rodrigopedra:
I am sorry if this looks like a rant by any means, I value all the hard work put into it, and into the other many Spatie packages, as open source for the community.

And am very grateful for all your hard work, and how it impacted on my family's life.

I just would like to point out, somehow this PR was missed on the changelog, and one of my apps had a subclass with a private `$fileSize`, so it started failing for one of my clients.

I know I shouldn't be subclassing, and should have more tests, this is a feature that is rarely used on a rather stable project.

Likewise, I could easily fix the issue upon a client's report, and no one was upset, but I thought on letting you know.

ping @freekmurze (just for awareness, thank you for all the hard work)

rodrigopedra:
I actually wanted to mean release notes instead of the changelog.

Yet, everyone is happy, and no puppies were harmed during the fix 😃

thanks again 🙏

PR #3361, opened 2023-08-31
Make `Filesystem@copyToMediaLibraryFromRemote()` more extensible

cosmastech:
We have a situation where we have two separate AWS buckets and we need to be able to copy data from one to the other. They are both registered in the filesystem. One is named `s3` and the other is named `s3_upload`, which serves as a temporary storage for user uploaded files.

Once a document is uploaded and the filekey is passed in a request, our code relies on InteractsWithMedia@addMediaFromDisk() to copy the data from `s3_upload` to its permanent home within `s3`.

Unfortunately, after some playing around this week, I see that we are actually streaming the file through our server rather than relying on the AWS to copy the object for us.

Luckily, Filesystem is always built from the container. So there's two changes that need to take place.

```php

class CustomFilesystem extends Spatie\MediaLibrary\MediaCollections\Filesystem
{
    protected function shouldCopyFileOnDisk(RemoteFile $file, Media $media, string $diskDriverName): bool
    {
        if ($file->getDisk() === 's3_upload' && $media->disk === 's3') {
            return true;
        }
        return parent::shouldCopyFileOnDisk($file, $media, $diskDriverName);
    }
}
```

It would look like we could just change `Filesystem@copyFileOnDisk()`, but unfortunately to use the S3Client how we want, we need to know: origin bucket, origin file key, destination bucket, and destination file key.

The parameters passed to this method currently are origin file key, destination file key, and the destination disk.

What's proposed here is simply a "go-between" method that can be more easily overridden. I've named it `prepareCopyFileOnDisk` which is truly not a great name, but 🤷 .

I think a better addition might be to create some sort of comparator class that would handle this. Maybe even a configuration to say "you should stream from one disk to the other."

Look forward to hearing your thoughts on this @freekmurze and will gladly PR any ideas/suggestions you may want to throw out here. 👍 

edit: I realize that the idea of a separate class to handle this is a nice idea, but it would be a BC to remove or move the methods in the Filesystem class 😓

### DIFF
--- a/src/MediaCollections/Filesystem.php
+++ b/src/MediaCollections/Filesystem.php
@@ -54,6 +54,11 @@ class Filesystem
         return true;
     }
 
+    public function prepareCopyFileOnDisk(RemoteFile $file, Media $media, string $destination): void
+    {
+        $this->copyFileOnDisk($file->getKey(), $destination, $media->disk);
+    }
+
     public function copyToMediaLibraryFromRemote(RemoteFile $file, Media $media, ?string $type = null, ?string $targetFileName = null): void
     {
         $destinationFileName = $targetFileName ?: $file->getFilename();
@@ -65,7 +70,7 @@ class Filesystem
             : $media->getDiskDriverName();
 
         if ($this->shouldCopyFileOnDisk($file, $media, $diskDriverName)) {
-            $this->copyFileOnDisk($file->getKey(), $destination, $media->disk);
+            $this->prepareCopyFileOnDisk($file, $media, $destination);
 
             return;
         }


freekmurze:
Thank you for this.

I'll merge this in for now, but I'll probably change things up when working on the next major version.

padre:
@cosmastech, I have been using a similar solution in my project for a long time (when I also realised that this package was actually streaming the file through our server rather than relying on the AWS to copy the object for us).

But it is still slow with large files, could you please tell me how long it takes you?

Some examples:

- 4021435391 bytes (3.75 GB) 57.137s consumed
- 3882712998 bytes (3.62 GB) 54.581s consumed
- 3850437023 bytes (3.59 GB) 53.605s consumed

Thanks!

cosmastech:
> @cosmastech, I have been using a similar solution in my project for a long time (when I also realised that this package was actually streaming the file through our server rather than relying on the AWS to copy the object for us).
> 
> But it is still slow with large files, could you please tell me how long it takes you?
> 
> Some examples:
> 
> * 4021435391 bytes (3.75 GB) 57.137s consumed
> * 3882712998 bytes (3.62 GB) 54.581s consumed
> * 3850437023 bytes (3.59 GB) 53.605s consumed
> 
> Thanks!

@padre Not entirely sure to be honest. I actually just started implementing the changes to our app this morning. Locally, I see about a 90% reduction in time when copying from one bucket to another. However, I was testing with a 110 MB file, not gigabytes. I believe at those filesizes, the AWS S3 SDK will be uploading via multipart. I admittedly do not know how this is handled, but from the scan I took earlier, it still shouldn't be streaming the file across your server.

Are you seeing a decrease in time from using this package's default copy-by-stream for a bucket to bucket file copy?

Also, are you confident that you're calling S3Client@copy() in your extended Filesystem class?

cosmastech:
Just ran a quick test, and a 768 MB file took 34.64s.

padre:
Thank you very much.

Don't you think it is too much for time for a file that is in s3?

cosmastech:
Sorry didn't see your edit there @padre
> Thank you very much.

You're welcome!

> Don't you think it's too long for a file that is in s3?

I don't know. Is it longer than I would like? Yes. Is there any other kind of alternative? 🤷 The only other thought I would have is running it async.

This was just a local test (running on valet). I believe results *may* look pretty different on production, as the default package's default copy method even runs much quicker there.

Per chance, have you reached out to AWS or are there any docs that tell us what the "normal" transfer speed should be?

Issue #3367, opened 2023-09-04
Problems with parallel testing and temporary folder


korridor:
First of all, thanks for the great package! 

We use the package in multiple projects. For one of the projects we also use parallel testing in Laravel like this: https://laravel.com/docs/10.x/testing#running-tests-in-parallel

This creates the following exception in some tests:
```
ErrorException: unlink(/tmp/49e39689192636d5d0bc23f681f36c1f): No such file or directory

/var/www/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php:254
/var/www/vendor/spatie/image/src/GlideConversion.php:130
/var/www/vendor/spatie/image/src/Image.php:122
/var/www/vendor/spatie/laravel-medialibrary/src/ResponsiveImages/ResponsiveImageGenerator.php:92
/var/www/vendor/spatie/laravel-medialibrary/src/ResponsiveImages/ResponsiveImageGenerator.php:62
/var/www/vendor/spatie/laravel-medialibrary/src/Conversions/Actions/PerformConversionAction.php:40
/var/www/vendor/spatie/laravel-medialibrary/src/Conversions/FileManipulator.php:71
/var/www/vendor/laravel/framework/src/Illuminate/Collections/Traits/EnumeratesValues.php:236
/var/www/vendor/spatie/laravel-medialibrary/src/Conversions/FileManipulator.php:70
/var/www/vendor/spatie/laravel-medialibrary/src/Conversions/FileManipulator.php:39
/var/www/vendor/spatie/laravel-medialibrary/src/MediaCollections/Filesystem.php:52
/var/www/vendor/spatie/laravel-medialibrary/src/MediaCollections/FileAdder.php:444
/var/www/vendor/spatie/laravel-medialibrary/src/MediaCollections/FileAdder.php:428
/var/www/vendor/spatie/laravel-medialibrary/src/MediaCollections/FileAdder.php:278
/var/www/vendor/spatie/laravel-medialibrary/src/MediaCollections/FileAdder.php:293
...
```

I debugged the problem and I'm pretty use that the problem is that the parallel tests in Laravel clear the system temporary folder after tests. Since all parallel processes use the same temporary folder, some testing processes fail with the given exception.

I fixed this problem by setting the temporary directory of `spatie/image` to the temporary folder of `spatie/laravel-medialibrary`.
It does not have to be this specific folder, I think any folder that is not cleared by the parallel testing system of Laravel would also solve this error.

[...]

PR #3368, opened 2023-09-04
Use temporary folder of spatie/laravel-medialibrary for spatie/image as well

korridor:
Fix for issue #3367


freekmurze:
Thanks!

PR #3371, opened 2023-09-07
add feature: getLastMedia

aabadawy:
As the `getFirstMedia` feature is popular to use in many cases, I think also one of the most popular cases to retrieve the **last media.** 
for example: 
- Get the profile picture.
- Get the uploaded CV.
- Get the project proposal doc .
- I think in the previous examples, usually the default media is the **last** one, not the first **one**.
so I had added `getLastMedia`, `getLastMediaUrl`, `getLastMediaPath`, and `getLastMediaTemporaryUrl`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ phpunit.xml
 .php-cs-fixer.cache
 phpstan.neon
 tests/Support/temp/
+/.idea
--- a/src/InteractsWithMedia.php
+++ b/src/InteractsWithMedia.php
@@ -10,6 +10,7 @@ use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\Validator;
 use Illuminate\Support\Str;
+use InvalidArgumentException;
 use Spatie\MediaLibrary\Conversions\Conversion;
 use Spatie\MediaLibrary\Downloaders\DefaultDownloader;
 use Spatie\MediaLibrary\MediaCollections\Events\CollectionHasBeenCleared;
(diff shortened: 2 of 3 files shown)


freekmurze:
To keep the package simple, I'm going to hold off on this one for now.

ssheduardo:
But is there any way to get the last item added to a collection?
When I upload an image I want to return the information of the uploaded item, how can I do it?

aabadawy:
@ssheduardo, after uploading media using `toMediaCollection()` method this return the uploaded media model, which you can use, to access the media information. 

Also, you can retrieve the **last** uploaded media attached with your model like this, 
`$lastMedia = $yourModel->getMedia()->last()`
This will return the last media item in the media collection

ssheduardo:
> @ssheduardo, after uploading media using `toMediaCollection()` method this return the uploaded media model, which you can use, to access the media information.
> 
> Also, you can retrieve the **last** uploaded media attached with your model like this, `$lastMedia = $yourModel->getMedia()->last()` This will return the last media item in the media collection

Thanks for your information, yesterday I was try this and work nice.

aabadawy:
> > @ssheduardo, after uploading media using `toMediaCollection()` method this return the uploaded media model, which you can use, to access the media information.
> > Also, you can retrieve the **last** uploaded media attached with your model like this, `$lastMedia = $yourModel->getMedia()->last()` This will return the last media item in the media collection
> 
> Thanks for your information, yesterday I was try this and work nice.

You are welcome @ssheduardo, Happy to hear that 😄

richt222:
@freekmurze, just adding - I think this is a good contribution that would make a lot of sense. Given there's a "getFirst", a "getLast" counterpart seems a logical addition.

Retrieving the last added media would serve a lot of common use cases, as @aabadawy describes. Probably even more so than getFirst.

Might be worth considering again.

PR #3376, opened 2023-09-15
Avoid creating temporary directory until needed

cosmastech:
In performing conversions, a temporary directory is created and then the original file is copied to the temp directory. It's possible that this is never needed (since there may be no new conversions to create). I've created a simple wrapper that handles creating the temporary directory and copying the file only when first requested.

Not sure about tests here. Let me know if you have any ideas on what you would like to see on that front.


freekmurze:
Thank you! I'm going to pass on this for now to keep the package as simple as it can be.

PR #3389, opened 2023-09-25
Support drivers containing gcs or s3 in driver name

LukeAbell:
We have multiple gcs drivers (Ex. `gcs-private`, `gcs-public`).

This PR detects GCS as long as it's in the driver name.

### DIFF
--- a/src/MediaCollections/Filesystem.php
+++ b/src/MediaCollections/Filesystem.php
@@ -340,7 +340,7 @@ class Filesystem
             ? $media->conversions_disk
             : $media->disk;
 
-        if (! in_array($diskDriverName, ['s3', 'gcs'], true)) {
+        if (! Str::contains($diskDriverName, ['s3', 'gcs'], true)) {
             $this->filesystem->disk($diskName)->makeDirectory($directory);
         }
 


src/MediaCollections/Filesystem.php
```diff
@@ -340,7 +340,7 @@ public function getMediaDirectory(Media $media, ?string $type = null): string
             ? $media->conversions_disk
             : $media->disk;
 
-        if (! in_array($diskDriverName, ['s3', 'gcs'], true)) {
+        if (! Str::contains($diskDriverName, ['s3', 'gcs'], true)) {
             $this->filesystem->disk($diskName)->makeDirectory($directory);
```
grEvenX, line 344: We ran into an issue with this approach with detecting what actions to perform based on the driver.
Why? We are using Sentry and they create a wrapper for the storage drivers, see https://github.com/getsentry/sentry-laravel/issues/775 for the details about the issue originally reported to Sentry.

The question is: Is this logic really needed at all? From my experience the Filestorage drivers handles the creation of a directory _if needed_. Meaning if you try to use a local storage driver and you try to write to the path, it will try to create the directory, and if you use S3, then it will just skip the step to make a directory.

cc @freekmurze @LukeAbell

LukeAbell: No idea if the code is required. I would also assume you could just remove that line entirely.

grEvenX: @LukeAbell I created a PR for it here: https://github.com/spatie/laravel-medialibrary/pull/3397

PR #3395, opened 2023-09-27
Adding example to docs to pass translation keys from Vue I18n in (pro) components

jeffreyvanhees:
The package vue-i18n (https://vue-i18n.intlify.dev/) offers a $tm-function to pass all keys at once as an array from a translation file. Maybe it's good to adapt this in the documentation.

See: https://vue-i18n.intlify.dev/api/composition.html#tm-key

`lang/media-library.php`
```php
<?php

return [
    'fileTypeNotAllowed' => 'Je moet een bestand van het type',
    'tooLarge' => 'Bestand te groot, maximaal',
    'tooSmall' => 'Bestand te klein, minimaal',
    'tryAgain' => 'probeer dit bestand opnieuw te uploaden',
    // ...
];
```

```vue
<MediaLibraryCollection
    :translations="$tm('media-library')"
/>
```

### DIFF
--- a/docs/handling-uploads-with-media-library-pro/handling-uploads-with-vue.md
+++ b/docs/handling-uploads-with-media-library-pro/handling-uploads-with-vue.md
@@ -687,6 +687,14 @@ window.mediaLibraryTranslations = {
 };
 ```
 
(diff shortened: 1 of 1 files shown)


jeffreyvanhees:
Wow that was fast @freekmurze! Thanks for accepting :)

PR #3397, opened 2023-09-28
Remove unnecessary directory creation

grEvenX:
This PR suggests to remove the conditional directory creation logic in the `getMediaLibrary()` function.
Unless it's a very specific reason to do it, I think it would be better to avoid checking for a specific driver to run the logic for creating a directory or not (since that can cause unexpected issues). 
In our case we ran into an issue with the solution since another library was masking the real driver name (so it wasn't called anything with "s3" anymore), and since our S3 permission setup only allowed writing files, not creating an empty directory, the storing of the file caused an exception to be thrown.
See relevant comment at https://github.com/spatie/laravel-medialibrary/pull/3389/files#r1338270714.

According to [Flysystem](https://flysystem.thephpleague.com/docs/usage/filesystem-api/), the actual driver creates the required directory when attempting to write a file;
> When writing files, the directory you’re writing to will be created automatically if and when that is required in the filesystem you’re writing to. If your filesystem does not require directories to exist (like AWS S3), the directory is not created. This is a performance consideration. Of course, you can always create the directory yourself by using the createDirectory operation.

### DIFF
--- a/src/MediaCollections/Filesystem.php
+++ b/src/MediaCollections/Filesystem.php
@@ -332,18 +332,6 @@ class Filesystem
             $directory = $pathGenerator->getPathForResponsiveImages($media);
         }
 
-        $diskDriverName = in_array($type, ['conversions', 'responsiveImages'])
-            ? $media->getConversionsDiskDriverName()
-            : $media->getDiskDriverName();
-
-        $diskName = in_array($type, ['conversions', 'responsiveImages'])
-            ? $media->conversions_disk
-            : $media->disk;
-
-        if (! Str::contains($diskDriverName, ['s3', 'gcs'], true)) {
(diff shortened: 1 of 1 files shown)


freekmurze:
How have you tested that the package works without this logic?

grEvenX:
@freekmurze I have tested with Local Storage and S3 storage on where I have used media conversions, but haven't tested all drivers Flysystem works with and I can't guarantee that it works on all drivers. If this would cause an issue or an error I believe a bug-report should be filed to Flysystem as they have documented this behaviour.

I've also reviewed all code calling the changed method and verified that all operations related to the directory are done through Flysystem and verified that the current test suite passes all tests after this change.

freekmurze:
Thank you!

PR #3399, opened 2023-09-30
added HEIC support

boryn:
HEIC is supported by ImageMagick for more than two years already:
- added HEIC support when using 'imagick' extension
- HEIC test


freekmurze:
Very nice, thanks!

dgillier:
Great news !

But HEIC file conversion are not working, my model class include the following.
(It is fine with JPEG,WEBP,PNG files)
```
        $this
            ->addMediaCollection('pictures')
            ->acceptsMimeTypes(['image/jpeg', 'image/png', 'image/webp','image/tiff','image/heic', 'image/heif']);
        $this->addMediaConversion('webp')
            ->performOnCollections('pictures')
            ->width(1024)  
            ->format('webp');
```
I have in my .env file:
`IMAGE_DRIVER=imagick`
and of course imagick and ImageMagick are installed :)

What I'm missing?
Thanks, Denis

boryn:
I bet you get some specific exception? Where exactly does it fail?

boryn:
Probably you have conversions happening in background? Try to set this `QUEUE_CONVERSIONS_BY_DEFAULT=false` in `.env` (or directly `'queue_conversions_by_default'` in `config/media-library.php`) so that the conversions are done without a queue and maybe now there would be some more info.

dgillier:
Yes in Background, will try that :)
Thanks

dgillier:
My Imagemagick version is not supporting webp (I'm converting to webP) :(((

Thanks for your help,
Denis

boryn:
And if you switch target format to jpeg, does it allow to have heic as the source image? For sure heic is supported since imagic 6.9.10. maybe earlier

andrezanna:
Hello, I tried, but for me Heic files fail as below

php artisan media-library:regenerate --ids=19113
 1/1 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%All done, but with some error messages:
Media id 19113: `Unable to read image from path (/tmp/GlidefTouty).`

boryn:
> Unable to read image from path (/tmp/GlidefTouty)
This tells more like problems with just reading a file, not converting. Does it work normally with .jpg files? What happens if you upload .heic file without the regenerate option?

boryn:
I can see the latest imagick version is 7.1.1 (https://imagemagick.org/index.php) so maybe it would be worth upgrading.

On the other hand, I would debug in the `Decoder.php:25` what is inside the `$path` variable and if it empty, I would trace it why it's empty.

andrezanna:
Ok, so I found my imagick extension was built without heic support, now i rebuilded and added it. 

Now regenerating and normal creation work without errors.

For future reference you should check if HEIC is enabled and supported by running
`php -r 'phpinfo();' | grep HEIC`

Here is also the guide that helped me regenerating it
https://eplt.medium.com/5-minutes-to-install-imagemagick-with-heic-support-on-ubuntu-18-04-digitalocean-fe2d09dcef1

Many thanks

PR #3403, opened 2023-10-05
Get url of the image for last media

rrakibul:
Like `getFirstMediaUrl`, I added `getLastMediaUrl` function as a convenience-method. This will pick the last item of the collection.


freekmurze:
I'm going to pass on this one for now to keep the package as simple as it can be.

PR #3404, opened 2023-10-08
Add rollback for migration

dave-nicholas:
Add rollback migration for media table.


freekmurze:
Thanks, we don't use rollbacks for any of our migrations.

PR #3419, opened 2023-10-23
Delete orphaned media in clean command

mbardelmeijer:
The use-case here is to automatically delete orphaned media objects, which also deletes them from the disk.

In database structure, it's common to use cascading deletes via foreign keys. When deleting database records like that, the `Media` object isn't automatically deleted due to its morphed relation. This delete action should be implemented by the developer to delete the related media objects in the `deleting` event, but this can be easily missed. No direct effects are visible on application level, but deleted storage objects remain indefinitely in the `medias` table, as well as on storage, which might infringe on GDPR requirements, as well as unnessesary storage used.

This PR adds logic to automatically delete orphaned models in the clean command. ~I've added it as an opt-out, but if a safer opt-in option (`--delete-orphaned` instead of `--skip-orphaned`)~ I've added it as an opt-in option, where `--delete-orphaned` needs to be included to delete orphaned models.

Happy to chat about other approaches if preferred :)

### DIFF
--- a/src/MediaCollections/Commands/CleanCommand.php
+++ b/src/MediaCollections/Commands/CleanCommand.php
@@ -23,8 +23,8 @@ class CleanCommand extends Command
     protected $signature = 'media-library:clean {modelType?} {collectionName?} {disk?}
     {--dry-run : List files that will be removed without removing them},
     {--force : Force the operation to run when in production},
-    {--rate-limit= : Limit the number of requests per second },
-    {--skip-orphaned : Do not remove orphaned media items},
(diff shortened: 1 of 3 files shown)


freekmurze:

freekmurze:
Could you also rebase with `main`. That should make the tests pass.

mbardelmeijer:
The tests have been added. I've changed it to an opt-in function to ensure in complex cases we aren't deleting media objects where not wanted (i.e. complex global scopes, etc).

freekmurze:
Thanks!

PR #3428, opened 2023-10-31
Don't include soft deleted models as orphaned media

patrickomeara:
This follows on from @mbardelmeijer's great work in #3419. 

When cascading deletes are used the media should definitely be removed from the database and storage as it's relationship no longer exists. However, when a model is soft deleted it has the option of being restored and the accompanying media needs to be available.


freekmurze:
Thanks!

PR #3431, opened 2023-11-02
Added missing public function signatures from InteractsWithMedia to HasMedia interface

korridor:
Fixes #3426

### DIFF
--- a/src/HasMedia.php
+++ b/src/HasMedia.php
@@ -2,16 +2,19 @@
 
 namespace Spatie\MediaLibrary;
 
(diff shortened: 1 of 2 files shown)


freekmurze:
Modifying interfaces is potentially a breaking change. I'll take a look again when creating the next major version of this package.

korridor:
@freekmurze Would it be ok if I create PHPdocs for those functions instead?

DanielGSoftware:
@freekmurze the next major version was 1 month after this issue, but it's not been addresse..

PR #3432, opened 2023-11-02
Added missing config option to php-cs-fixer

korridor:
Without this option, I get the following error on my local setup:

```
In Finder.php line 661:
                                                                                 
  You must call one of in() or append() methods before iterating over a Finder.  

```


freekmurze:
Thanks for your work on this.

We're going to leave cs-fixer behind in the next major version.

PR #3434, opened 2023-11-06
Suggest the installation of svgo through snap on Ubuntu

3ub41r:
I would like to suggest installing svgo through snap as an alternative to installing npm


freekmurze:
Thanks!

freekmurze:
Looks good to me! Any thoughts @AlexVanderbist ?

PR #3451, opened 2023-11-29
Use image v3

freekmurze:


cosmastech:
@freekmurze previously, using `\Storage::fake('local')` would prevent test files from being created. It seems like there's some kind of regression here, not sure if it's with Image v3 or this PR. After running tests now, we have files being created at `./storage/media-library/temp/bTicW6ce5DddWYO3ASVwthRi29688BmM/RLBed4mV4XYy6uAMZPOGykJEWdqd7QZAthumb.jpg`

Luckily it's not breaking our pipelines, just a slight annoyance 😆

freekmurze:
I'm sorry that the upgrade is causing (minor) troubles. The test suites of stuff still seems to work.

As always, I'm open for PR to fix things that go wrong on your end.

PR #3455, opened 2023-12-11
Improve defaultSanitizer

francoism90:
This should replace double spaces, symbols, etc., but keep Unicode chars + dots.

It should also trim the last and first dot, to be sure it cannot contain hidden (Linux), and possible extension conflicts.

Maybe `p{S}` could be added, this allows specials symbols and emoji.


francoism90:
Too many errors for now, will check later.

PR #3473, opened 2023-12-21
[v11] Remove ZipStream V2 support

erikn69:
https://github.com/spatie/laravel-medialibrary/blob/280e20b0bd5649cd89b7d72d84afade38e3cb8b9/composer.json#L35


freekmurze:
Thanks!

PR #3474, opened 2023-12-21
fix png transparent bug

ariaieboy:
There is an unnecessary format change when doing conversation that makes some weird bug when we use the GD driver.
This one line of code made me read the whole system of the image, image-optimizer, and laravel media library package 😊

the image input:
image

the result after converting to webp format with a white background before this PR:
image

and after this PR:
image

### DIFF
--- a/src/Conversions/Actions/PerformManipulationsAction.php
+++ b/src/Conversions/Actions/PerformManipulationsAction.php
@@ -31,8 +31,7 @@ class PerformManipulationsAction
         }
 
         $image = Image::useImageDriver(config('media-library.image_driver'))
-            ->loadFile($conversionTempFile)
-            ->format('jpg');
+            ->loadFile($conversionTempFile);
 
         try {
             $conversion->getManipulations()->apply($image);


freekmurze:
By default, the media library will make jpg conversions. If you don't want that behaviour, tack on `shouldKeepOriginalImageFormat()` to your conversion and the result will be a png (when your input is a png)

ariaieboy:
> By default, the media library will make jpg conversions. If you don't want that behaviour, tack on `shouldKeepOriginalImageFormat()` to your conversion and the result will be a png (when your input is a png)

The `keepOriginalImageFormat` method is not fixing my problem. 
The problem is that when you use the imagick driver the format change won't take place until you write the image. 
But in the GD driver when you call the `format` method the image actually changes to that format. 
Since the JPG does not support transparent background. it's either gonna generate a black background or in my case this weird background.

ariaieboy:
I think instead of changing this here we should change the GD driver `format` method.
It must save the format and apply it when we call the save method.

freekmurze:
> I think instead of changing this here we should change the GD driver format method.
It must save the format and apply it when we call the save method.

Yeah, that would be better I think. You're up for making that PR?

Issue #3475, opened 2023-12-22
[11.x] Incorrect PDF conversion results when used with GD driver


kduma:
Continuation of #3465.

With following configuration in [User.php](https://github.com/kduma-archive/test-laravel-medialibrary-failing-pdf/blob/b11b6afd2844de11a85c3682774a098245133099/app/Models/User.php#L20-L26) model:
```php
public function registerMediaConversions(Media $media = null): void
{
    $this->addMediaConversion('preview')
        ->fit(Fit::Crop, 300, 300)
        ->format('png')
        ->nonQueued();
}
```

Given the following test case: [ExampleTest.php](https://github.com/kduma-archive/test-laravel-medialibrary-failing-pdf/blob/3f91954f91721a77ed6321bcd84d53ac881d4aed/tests/Feature/ExampleTest.php#L17-L30)
```php
$user = \App\Models\User::factory()
    ->create();

$user
    ->addMedia(resource_path('dummy.pdf'))
    ->preservingOriginal()
    ->toMediaCollection();

$path = $user->getFirstMediaPath(conversionName: 'preview');

[$width, $height] = getimagesize($path);

$this->assertEquals(300, $width, 'The width of the conversion is not 300px.');
$this->assertEquals(300, $height, 'The height of the conversion is not 300px.');
```

The generated conversion is incorrect. Here are results from failed assertions:
```
1) The width of the conversion is not 300px.
Failed asserting that 1190 matches expected 300.
Expected :300
Actual   :1190

2) The height of the conversion is not 300px.
Failed asserting that 1684 matches expected 300.
Expected :300
Actual   :1684
```

Repository with test app: [kduma-archive/test-laravel-medialibrary-failing-pdf](https://github.com/kduma-archive/test-laravel-medialibrary-failing-pdf)

freekmurze:
Let me know when #3476 can be reviewed.

Thanks for you work on this.

kduma:
> Let me know when #3476 can be reviewed.
> 
> Thanks for you work on this.

@freekmurze Looks that [spatie/image#221](https://github.com/spatie/image/pull/221) PR fixes this.

Issue #3465, opened 2023-12-19
[11.x] Conversions crashes when generating previews of PDF file


kduma:
Given following test case: [ExampleTest.php](https://github.com/kduma-archive/test-laravel-medialibrary-failing-pdf/blob/ace05273de809004e459e93452faebe3f891f02d/tests/Feature/ExampleTest.php#L17-L23)
```php
$user = \App\Models\User::factory()
    ->create();

$user
    ->addMedia(resource_path('dummy.pdf'))
    ->preservingOriginal()
    ->toMediaCollection();
```

With following configuration in [User.php](https://github.com/kduma-archive/test-laravel-medialibrary-failing-pdf/blob/ace05273de809004e459e93452faebe3f891f02d/app/Models/User.php#L20-L26) model:
```php
public function registerMediaConversions(Media $media = null): void
{
    $this->addMediaConversion('preview')
        ->fit(Fit::Crop, 300, 300)
        ->format('png')
        ->nonQueued();
}
```

There is a crash:
```
Spatie\Image\Exceptions\UnsupportedImageFormat: Unsupported format `pdf`.
/Users/kduma/Developer/web/test-laravel-medialibrary-failing-pdf/vendor/spatie/image/src/Exceptions/UnsupportedImageFormat.php:11
/Users/kduma/Developer/web/test-laravel-medialibrary-failing-pdf/vendor/spatie/image/src/Drivers/Gd/GdDriver.php:153
/Users/kduma/Developer/web/test-laravel-medialibrary-failing-pdf/vendor/spatie/image/src/Image.php:85
/Users/kduma/Developer/web/test-laravel-medialibrary-failing-pdf/package-sources-laravel-medialibrary/src/Conversions/Actions/PerformManipulationsAction.php:38
/Users/kduma/Developer/web/test-laravel-medialibrary-failing-pdf/package-sources-laravel-medialibrary/src/Conversions/Actions/PerformConversionAction.php:30
/Users/kduma/Developer/web/test-laravel-medialibrary-failing-pdf/package-sources-laravel-medialibrary/src/Conversions/FileManipulator.php:71
/Users/kduma/Developer/web/test-laravel-medialibrary-failing-pdf/vendor/laravel/framework/src/Illuminate/Collections/Traits/EnumeratesValues.php:240
/Users/kduma/Developer/web/test-laravel-medialibrary-failing-pdf/package-sources-laravel-medialibrary/src/Conversions/FileManipulator.php:70
[...]

PR #3477, opened 2023-12-22
[11.x] Add test to check that pdf conversions are correct

kduma:
This is a test case for issues: #3475 and #3465 which were fixed in [spatie/image#221](https://github.com/spatie/image/pull/221) Pull Request.


kduma:
Looks like prefer-lowest tests are failing. 

Updating `spatie/image` constraint in `composer.json` should be considered:
```
"spatie/image": "^3.3.1"
```

freekmurze:
Thanks!

PR #3484, opened 2023-12-26
laravel-pint-action to major version

erikn69:
https://github.com/aglipanci/laravel-pint-action/releases/tag/v2
major version tag avoids "bumps" on every menor release


freekmurze:
Thank you!

PR #3497, opened 2024-01-02

madurapa:
Added missing reverse the migration


freekmurze:
Thanks, but passing on this one. We never perform down migrations.

madurapa:
I got the error `SQLSTATE[42S01]: Base table or view already exists: 1050 Table 'media' already exists` when I call `php artisan migrate:refresh`  and that's why I sent this PR

PR #3504, opened 2024-01-09
New 'getDownloadFilename' method for extensibility

chrispage1:
When downloading a file by utilising Responsable, the file naming at the point of download isn't very flexible.

I've added a method that allows extending the Media class, and in turn the download file name, rather than having to create my own download controller for a filament project I'm working on (for some reason in Filament the file names are by default obfuscated). I can then simply extend the functionality by overriding the method

```php
    public function getDownloadFilename(): string
    {
        return $this->name . '.' . pathinfo($this->file_name, PATHINFO_EXTENSION);
    }
```

In addition I've utilised Symfony's HeaderUtils class (as Laravel does) to generate the disposition, ensuring it's sanitised and always correctly formatted.

Thanks!

### DIFF
--- a/src/MediaCollections/Models/Media.php
+++ b/src/MediaCollections/Models/Media.php
@@ -14,6 +14,7 @@ use Illuminate\Database\Eloquent\Relations\MorphTo;
 use Illuminate\Mail\Attachment;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
+use Illuminate\Support\Str;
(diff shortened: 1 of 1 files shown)


freekmurze:
Seems like the testsuite is failing because of these changes. Could you take a look at that?

chrispage1:
> Seems like the testsuite is failing because of these changes. Could you take a look at that?

Sure, its changed the disposition structure slightly, I'll reconsider that element 👍

freekmurze:
Thanks!

PR #3505, opened 2024-01-10
Adds the HttpFacadeDownloader

peterfox:
# Changes

- Adds a new HttpFacadeDownloader implementing the Downloader interface.
- Adds a test for the new Downloader.

# Why

I find it difficult to write tests with the current downloader as it can't even be mocked by the container and requires a config change. I thought at a minimum this might be a small improvement as the Http facade is pretty good for mocking HTTP calls.

# Usage

The downloader is used as normal.

```php
    // config/media-library.php

    /*
     * When using the addMediaFromUrl method you may want to replace the default downloader.
     * This is particularly useful when the url of the image is behind a firewall and
     * need to add additional flags, possibly using curl.
     */
    'media_downloader' => Spatie\MediaLibrary\Downloaders\HttpFacadeDownloader::class,
```

This then makes it easier in tests to mock the download of files.

```php
$url = 'http://medialibrary.spatie.be/assets/images/mountain.jpg';
$yourModel
   ->addMediaFromUrl($url)
   ->toMediaCollection();
```

with a test like this:

```php
Http::fake([
    // Stub a response where the body will be the contents of the file
    'http://medialibrary.spatie.be/assets/images/mountain.jpg' => Http::response('::file::'),
]);

// Execute code for the test

// Then check that a request for the file was made
Http::assertSent(function (Request $request) {
    return $request->url() == 'http://medialibrary.spatie.be/assets/images/mountain.jpg';
});

// We may also assert that the contents of any files created
// will contain `::file::`
```

### DIFF
--- a/docs/advanced-usage/using-a-custom-media-downloader.md
+++ b/docs/advanced-usage/using-a-custom-media-downloader.md
@@ -46,3 +46,48 @@ class CustomDownloader implements Downloader {
 
 }
 ```
+
+## Using the Laravel Downloader
+
+You may configure the medialibrary config to use a downloader compatible more
+with Laravel that makes use of the built-in HTTP client. This is the quickest way
+to mock any requests made to external URLs.
+
+```php
+    // config/media-library.php
+
+    /*
+     * When using the addMediaFromUrl method you may want to replace the default downloader.
+     * This is particularly useful when the url of the image is behind a firewall and
+     * need to add additional flags, possibly using curl.
+     */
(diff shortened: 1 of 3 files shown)


freekmurze:
Could you document how this should be used?

peterfox:
@freekmurze apologies for the slow update. I've added some docs and another test that displays it working. Let me know if you need more.

freekmurze:
Thanks!

PR #3509, opened 2024-01-15
Update preparing-your-model.md - manipulations class has been replaced with `Fit` enum

bomshteyn:
It seems the manipulations class has been replaced with `Fit` enum


freekmurze:
Thanks!

PR #3520, opened 2024-01-27
Ability to use different width calculator per conversion using Responsive Images

jhorie:
If you would like to use different width calculators for different conversions for responsive images. You use the method withWidthCalculator(WidthCalculator $widthCalculator)  to provide a width calculator for this specific conversion. If none is given, the default width calculator from the config file is used.


freekmurze:
Thanks!

PR #3522, opened 2024-01-29
update docs to align with the latest changes in medialibrary-pro

lukasleitsch:
The dragula library was replaced with Sortable. The PR updates the docs to the latest changes in medialibrary-pro.


bram-pkg:
Seems Vue 2 support was also removed

freekmurze:
Thanks!

PR #3523, opened 2024-02-01
Add count media functionality

salamwaddah:
I would like to add `countMedia` as a fluent alternative to
- `count($this->getMedia($collectionName, $filters))`
- `$this->getMedia($collectionName)->count();`

Here's my use case in blade
```php

// before
@if($model->getMedia()->count() > 1)
@endif

// after
@if($model->countMedia() > 1)
@endif
```

A similar PR #1582 closed 5 years ago but I wish this is reconsidered

### DIFF
--- a/src/InteractsWithMedia.php
+++ b/src/InteractsWithMedia.php
@@ -241,11 +241,19 @@ trait InteractsWithMedia
     }
 
     /*
+     * Count media in the given collection.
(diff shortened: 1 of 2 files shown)


freekmurze:
Thanks, but I'm going to pass on this to keep the package simple.

PR #3530, opened 2024-02-07
fix: add missing down in migration file

eznix86:


Nielsvanpach:
We don't use down migrations in our packages or projects.

PR #3537, opened 2024-02-14
Support for passing an array of classes as extra attributes

pascalbaljet:
This PR adds support for passing an array of classes when rendering a `Media` Model:

```php
Media::first()->img('thumb', ['class' => [
    'rounded',
    'border' => true,
    'border-black' => false,
]]),
```

Under the hood, it uses Laravel's `Arr::toCssClasses` [helper method](https://laravel.com/docs/10.x/helpers#method-array-to-css-classes).


freekmurze:
Very nice! Could you also put an example of this in the docs?

freekmurze:
Very nice, thanks!

PR #3539, opened 2024-02-14
Add Laravel 11 Support

Jubeki:
I also removed the dev-dependency `doctrine/dbal` because it seems it is no longer required. (It also hindered the installation of Laravel 11)


Nielsvanpach:
Thanks!

PR #3541, opened 2024-02-15
Refactored how download file name is determined within MediaStream

chrispage1:
In addition to PR #3504, I have added the `getDownloadFilename()` method to the MediaStream download.

This allows custom file names to be utilised within the MediaStream zip generation, while not impacting number appends, etc. I've also written a test to cover this.

Thanks to @justmetalcake for pointing out this issue!

Cheers.


freekmurze:
Thanks!

PR #3543, opened 2024-02-16
Leave the ability to override media model in InteractsWithMedia trait

illambo:
This PR add the (little) ability to override media model in trait (like MediaRepository), this can facilitate some cases of applications with package used on multiple databases (without the need to modify the configuration at runtime).


freekmurze:
Thanks!

PR #3544, opened 2024-02-17
Update defining-conversions.md

misaf:
According to the definition of the method in the HasMedia interface, Media should be nullable in the registerMediaConversions method. However, in the documentation, it is represented as 'should not be null.' I have corrected them.


freekmurze:
Thanks!

PR #3547, opened 2024-02-17
FileAdder - Resolve media model from subject and reinforce HasMedia contract

illambo:
This PR follow up the pull #3543, just try resolve media model from subject in FileAdder and reinforce the HasMedia contract.


freekmurze:
Thanks!

PR #3548, opened 2024-02-19
Clarify that class is resolved using service container

bram-pkg:
Title speaks for itself. It's probably good to mention the classes defined in the config are resolved through the service container.


freekmurze:
Thanks!

PR #3549, opened 2024-02-20
Fix DivisionByZeroError in getHumanReadable

ziming:
After updating to the latest media library I got this Divide by Zero error

Think this commit broke it https://github.com/spatie/laravel-medialibrary/commit/81876c36c09d717e99e7c07cad8dd2775e25504e


freekmurze:
Could you add a test for this, so we don't get regressions in the future?

ziming:
@freekmurze added!

freekmurze:
Thanks!

PR #3551, opened 2024-02-21
Fix handling of dots in conversion names for clean command

sebastianpopp:
The clean command had an issue with conversions that contain a dot because their status is saved in a multidimensional array in the database. The clean command expected a one-dimensional array.


freekmurze:
Thanks!

PR #3553, opened 2024-02-23
Add test to check if conversion orientation is correct

ITSYV:
Test if the conversions get the right orientation when the original image has exif orientation data
https://github.com/spatie/laravel-medialibrary/discussions/3546


timvandijck:
Hey @ITSYV, thank you for the test!

I added a different test to the spatie/image package (which is used by medialibrary under the hood) and fixed the issue. However it doesn't seem to make your test pass.

Would you mind checking if you still encounter problems with the updated package? 

https://github.com/spatie/image/blob/main/tests/OrientationTest.php

ITSYV:
Thanks for the quick fix @timvandijck !
Seems to be fixed on my local machine. Going to deploy tonight and check on production. 
Not sure why my test is failing though

Issue #3567, opened 2024-03-17
Publishing the views is not working or not documented


omarherri:
To customizing the rendered HTML i tried to publish the view as mentioned in the documentation using:

`php artisan vendor:publish --provider="Spatie\MediaLibrary\MediaLibraryServiceProvider" --tag="views"`

But this returned _No publishable resources for tag [views]._

patinthehat:
@omarherri Have you tried running 
```sh
php artisan vendor:publish --provider="Spatie\MediaLibrary\MediaLibraryServiceProvider" --tag="medialibrary-views"
```
This follows the same format as the other publish commands, even though the docs indicate "views" as the correct tag.  Please let me know if you still run into an issue, and I'll dig in and get to the bottom of it! :+1:

omarherri:
@patinthehat, I tried your solution, but I still received the same message: _"No publishable resources for tag [medialibrary-views]."_ 

So, I looked up the package and noticed that the views are declared under the 'media-library' namespace. Therefore, I modified the command to:

`php artisan vendor:publish --provider="Spatie\MediaLibrary\MediaLibraryServiceProvider" --tag="media-library-views"`
 
 And it worked for me! Thank you for pointing me in the right direction.

patinthehat:
@omarherri thank you for the update, I'll update the docs accordingly! :+1:

PR #3569, opened 2024-03-18
Update docs with correct tag for publishing package views

patinthehat:
This PR updates the docs to include the correct tag when publishing the views for the package, see #3567.

ping @freekmurze


freekmurze:
Thanks!

PR #3576, opened 2024-03-27

cleargoal:
add function down() to migration with a purpose to developer may have this when publishing the package assets

### DIFF
--- a/database/migrations/create_media_table.php.stub
+++ b/database/migrations/create_media_table.php.stub
@@ -29,4 +29,12 @@ return new class extends Migration
             $table->nullableTimestamps();
         });
     }
+
+    /**
+     * Reverse the migrations.
+     */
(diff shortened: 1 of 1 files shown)


freekmurze:
We don't use down migrations in any package / project

cleargoal:
> We don't use down migrations in any package / project

it's a pity,
this caused me some inconvenience

farishrf:
@cleargoal perhaps the goal of not including a `down method` is to mitigate the risk of unintended data loss in production environments if `php artisan migrate:rollback` were to be run by mistake.

cleargoal:
> @cleargoal perhaps the goal of not including a `down method` is to mitigate the risk of unintended data loss in production environments if `php artisan migrate:rollback` were to be run by mistake.

Good reason.
Thanks!

PR #3578, opened 2024-03-30

farishrf:
Clean up the migration stub by removing unneeded argument.
image
No breaking change as we require a minimum of Laravel 10.x. Just a small cleanup.


freekmurze:
Thanks!

Issue #3582, opened 2024-04-11
Error when using withManipulations in combination with enumerations


tkempf:
When adding media with medialibrary v11 using withManipulations, the BackedEnum BorderType gets transformed to its scalar string value before calling the manipulation itself. 
This leads to the following type error:
 Spatie\Image\Image::border(): Argument #2 ($type) must be of type Spatie\Image\Enums\BorderType, string given

The code used is as follows:
`
    $this->addMedia($params->favicon->file)
        ->preservingOriginal()
        ->withManipulations([
            'bgrectangle' => [
                'background' => ['color'=>$favIconbgcolor],
                'border' =>  ['width' => $bordersize,'type' => BorderType::Expand, 'color'=>$bordercolor]
            ]
        ])
        ->toMediaCollection('favicon');
`

freekmurze:
It's accept a PR that fixes this.

tkempf:
Ok, the problem occurs on FileAdder.php line 331 during
    $media->manipulations = $this->manipulations;
the manipulation get cast to an array as requested in Models/Media.php on Line 74
    protected $casts = [
        'manipulations' => 'array',
        'custom_properties' => 'array',
        'generated_conversions' => 'array',
        'responsive_images' => 'array',
    ];

As for now, neither do i know the internals of media-libray nor, how you could change that cast to leave the enums alone, i don't feel capable of fixing it the right way ...
For my purpose i just need a border, so i made it work for me now by recreating the Enum in Conversions/Manipulations 53

    public function apply(ImageDriver $image): void
    {
        foreach ($this->manipulations as $manipulationName => $parameters) {
            switch($manipulationName){
                case 'border':
                    $parameters['type'] = BorderType::from($parameters['type']);
                    break;
            }
            $image->$manipulationName(...$parameters);
        }
    }

pls. don't get me wrong, it's not that i'm not willing to provide a PR, but this is no code, i would dare to offer as such!
Alas perhaps someone who has a better knwoledge of media-libary or functionality of custom casts in laravel can use this as entrance point to fix this error.

PR #3584, opened 2024-04-16
recreate BackedEnums when applying Manipulations

tkempf:
Fixes Issue #3582

### DIFF
--- a/src/Conversions/Manipulations.php
+++ b/src/Conversions/Manipulations.php
@@ -60,13 +60,34 @@ class Manipulations
     public function apply(ImageDriver $image): void
     {
         foreach ($this->manipulations as $manipulationName => $parameters) {
-            (isset($parameters['type'])) && $parameters['type'] = BorderType::from($parameters['type']);
-            (isset($parameters['position'])) && $parameters['position'] = AlignPosition::from($parameters['position']);
-            (isset($parameters['colorFormat'])) && $parameters['colorFormat'] = ColorFormat::from($parameters['colorFormat']);
-            (isset($parameters['constraints'])) && $parameters['constraints'] = Constraint::from($parameters['constraints']);
-            (isset($parameters['position'])) && $parameters['position'] = CropPosition::from($parameters['position']);
-            (isset($parameters['fit'])) && $parameters['fit'] = Fit::from($parameters['fit']);
-            (isset($parameters['flip'])) && $parameters['flip'] = FlipDirection::from($parameters['flip']);
+            switch($manipulationName){
+                case 'border':
+                    (isset($parameters['type'])) && $parameters['type'] = BorderType::from($parameters['type']) ;
+                    break;
+                case 'watermark':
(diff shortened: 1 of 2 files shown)


src/Conversions/Manipulations.php
```diff
@@ -52,6 +60,34 @@ public function isEmpty(): bool
     public function apply(ImageDriver $image): void
     {
         foreach ($this->manipulations as $manipulationName => $parameters) {
+            switch($manipulationName){
```
Nielsvanpach, line 63: Could you use a `match` operator instead?

src/Conversions/Manipulations.php
```diff
@@ -52,6 +60,34 @@ public function isEmpty(): bool
...
+                case 'border':
+                    (isset($parameters['type'])) && $parameters['type'] = BorderType::from($parameters['type']) ;
```
Nielsvanpach, line 65:
```suggestion
                    (isset($parameters['type'])) && $parameters['type'] = BorderType::from($parameters['type']);
```

Nielsvanpach (commented):
Could you add a test case to make sure this fix solves the problem?

src/Conversions/Manipulations.php
tkempf: What do you think about using neither match nor switch. Just recreate all Enums if the parameter is present independent of thy type of manipulation
```suggestion
    (isset($parameters['type'])) && $parameters['type'] = BorderType::from($parameters['type']) ;
    (isset($parameters['fit'])) && $parameters['fit'] = Fit::from($parameters['fit']);
    (isset($parameters['position'])) && $parameters['position'] = AlignPosition::from($parameters['position']);
    (isset($parameters['colorFormat'])) && $parameters['colorFormat'] = ColorFormat::from($parameters['colorFormat']);
    (isset($parameters['constraints'])) && $parameters['constraints'] = Constraint::from($parameters['constraints']);
    (isset($parameters['position'])) && $parameters['position'] = CropPosition::from($parameters['position']);
    (isset($parameters['fit'])) && $parameters['fit'] = Fit::from($parameters['fit']);
    (isset($parameters['flip'])) && $parameters['flip'] = FlipDirection::from($parameters['flip']);
```

freekmurze: I think a match we be more clear in this case

MrMeshok:
There is also valid usage without named parameters
```php
$media->manipulations = [
    'default' => [
        'border' => [
            1,
            BorderType::Expand,
            '000000'
        ]
    ]
];
```
I see 2 solutions there:
1. Try to find enum from parameters array by named parameter key or numeric position e.g. `BorderType::from($parameters['type'] ?? $parameters[1])`
2. Use reflection and check ImageDriver method parameters before calling it

Obviously reflection would be better, but there performance concerns, so it needs benchmarking.

I could take over this issue as this also need for my use case

timvandijck:
@tkempf thank you for your work!

@MrMeshok I'm going to already merge this PR, but if you have some more improvements on this topic I would be happy to look at a new PR from you.

PR #3599, opened 2024-04-30
Add `down` DB Migration to drop media table

ollyollyollyltd:
This package publishes a migration to create the `media` table, but it only has an `up` method.

This means that if the migration is rolled back, and then migrated again, the migration will fail as the `media` table already exists.

This pull request adds a simple `down` method to the migration to drop the `media` table.

### DIFF
--- a/database/migrations/create_media_table.php.stub
+++ b/database/migrations/create_media_table.php.stub
@@ -29,4 +29,10 @@ return new class extends Migration
             $table->nullableTimestamps();
         });
     }
+
+    public function down(): void
+    {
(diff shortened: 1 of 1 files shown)


Nielsvanpach:
We don't use down migrations in any package / project.

99.9% of the time you are not gonna use a down migration and you’re probably never testing them. Just write an up migration if you want to drop a column again. 👍

ollyollyollyltd:
Okay, fair enough.

Kind regards,

Olly Cross
OllyOllyOlly Ltd

PR #3600, opened 2024-04-30
Add getLastMedia, getLastMediaUrl, getLastTemporaryUrl, getLastMediaPath methods

kidii:
After that, it would not only be getFirstXY, but also getLastXY methods.

### DIFF
--- a/docs/basic-usage/retrieving-media.md
+++ b/docs/basic-usage/retrieving-media.md
@@ -34,6 +34,12 @@ Since retrieving the first media and the URL for the first media for an object i
 $media = $yourModel->getFirstMedia();
 $url = $yourModel->getFirstMediaUrl();
 ```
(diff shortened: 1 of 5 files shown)


aabadawy:
Nice Work!, Check this PR also https://github.com/spatie/laravel-medialibrary/pull/3371

kidii:
Oh, I didn't searched for it. :(

Let's see what happens now.😬

aabadawy:
> Oh, I didn't searched for it. :(
> 
> Let's see what happens now.😬

I still see the idea of adding getLastMedia is so helpful 🤝 , as many cases need to fetch the last media, like CV,Profile picture, invoice, ...etc. let's see @freekmurze opinion.

freekmurze:
I'm going to hold off this one for now to keep the package simple.

PR #3614, opened 2024-05-18
fix: Add drop functionality to media table migration on rollback

shojibflamon:
Hi @freekmurze,

The package is fantastic and saves a lot of time on media handling. However, I encountered an issue when rolling back and re-running migrations. It caused an exception: **General error: 1 table "media" already exists.** This indicates the table isn't dropped on rollback.

This PR adds the drop table functionality during rollback to resolve the issue.

Thanks

### DIFF
--- a/database/migrations/create_media_table.php.stub
+++ b/database/migrations/create_media_table.php.stub
@@ -29,4 +29,9 @@ return new class extends Migration
             $table->nullableTimestamps();
         });
     }
+
+    public function down(): void
+        {
+            Schema::dropIfExists('media');
+        }
 };


Nielsvanpach:
We don't use down migrations in any package / project.

99.9% of the time you are not gonna use a down migration and you’re probably never testing them. Just write an up migration if you want to drop a column again. 👍

shojibflamon:
Hi @Nielsvanpach,
I don't think so that down migration is not useful. In development section we have to do the rollback & fresh/refresh migration functionality several times. Specially in TDD approach or some other works. But without down migration it will be a exception to migration again in the same database. So, what do you think about this? Manually remove the table & do the migration again & again?

image

Thanks.

PR #3618, opened 2024-05-22
Do not limit the maximum execution time if the queue connection is "sync"

VGirol:
When regenerating the conversions and if the queue connection is set to "sync", the jobs are not handled "asynchronously". So the execution time can be very long depending on the number of medias to handle. And in some cases, the script reached the maximum execution time defined in php.ini and exit with an error code.

This PR disable the maximum execution time if the queue connection is "sync".

### DIFF
--- a/src/Conversions/Commands/RegenerateCommand.php
+++ b/src/Conversions/Commands/RegenerateCommand.php
@@ -38,13 +38,17 @@ class RegenerateCommand extends Command
 
         $this->fileManipulator = $fileManipulator;
 
-        if (! $this->confirmToProceed()) {
+        if (!$this->confirmToProceed()) {
             return;
         }
 
         $mediaFiles = $this->getMediaToBeRegenerated();
 
         $progressBar = $this->output->createProgressBar($mediaFiles->count());
+
+        if (config('media-library.queue_connection_name') == 'sync') {
+            set_time_limit(0);
+        }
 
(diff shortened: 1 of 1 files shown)


src/Conversions/Commands/RegenerateCommand.php
```diff
@@ -38,14 +38,18 @@ public function handle(MediaRepository $mediaRepository, FileManipulator $fileMa
...
-        if (! $this->confirmToProceed()) {
+        if (!$this->confirmToProceed()) {
             return;
         }
 
         $mediaFiles = $this->getMediaToBeRegenerated();
 
         $progressBar = $this->output->createProgressBar($mediaFiles->count());
 
+        if (config('media-library.queue_connection_name') == 'sync') {
```
freekmurze, line 49: could you make this a strict check?

freekmurze (changes requested):
Thanks!

Could you remove all code style changes?

src/Conversions/Commands/RegenerateCommand.php
VGirol: Done

VGirol:
> Thanks!
> 
> Could you remove all code style changes?

I'm sorry. I didn't see that code style had changed.

freekmurze:
Thank you!

PR #3619, opened 2024-05-26
Unwanted deletion fixes

inalto:
I tried to regenerate 350K+ images conversions  with responsive images, i spotted some cases that lead to wrong deletions on some generated images that have a part of the name included in other filenames, so the Str::contains should be more specific and i fixed the issue borrowing the addPropertiesToFileName pattern to check "{$fileName}___{$conversionName}";
Now all the unwanted missing conversion that appeared randomly are gone away.


freekmurze:
Thanks!

PR #3620, opened 2024-05-28
Added when() method to FileAdder

nikrovir:
My pull request adds syntactic sugar like the one used in eloquent builder to allow the when() method to be used.

Example:

```php
$model->addMedia($file)
    ->preservingOriginal(false)
    ->when(in_array($file->getMimeType(), $availableMimeTypes), function (FileAdder $fileAdder) {
        $fileAdder->withCustomProperties([
            'property' => 'value',
        ]);
    })->toMediaCollection();
```

Instead:

```php
$media = $model->addMedia($file)->preservingOriginal(false);

if (in_array($file->getMimeType(), $availableMimeTypes)) {
    $media->withCustomProperties([
        'property' => 'value',
    ]);
}

$media->toMediaCollection();
```

### DIFF
--- a/src/MediaCollections/FileAdder.php
+++ b/src/MediaCollections/FileAdder.php
@@ -541,4 +541,15 @@ class FileAdder
             ? $file.'.'.$extension
             : $file;
     }
+
+    public function when(bool $value, callable $callback = null, callable $default = null): self
+    {
+        if ($value === true) {
+            return $callback($this, $value) ?? $this;
+        } elseif ($default !== null) {
+            return $default($this, $value) ?? $this;
+        }
+
+        return $this;
+    }
 }


freekmurze:
Could you also add a test for this and add an example to the readme?

src/MediaCollections/FileAdder.php
```diff
@@ -541,4 +541,15 @@ protected function appendExtension(string $file, ?string $extension): string
             ? $file.'.'.$extension
             : $file;
     }
+
+    public function when(bool $value, callable $callback = null, callable $default = null): self
+    {
+        if ($value === true) {
+            return $callback($this, $value) ?? $this;
+        } elseif ($default !== null) {
```
freekmurze, line 549: Let's use a simple `if` here instead of `elseif`

freekmurze:
Closing due to inactivity.

PR #3622, opened 2024-05-28
Add missing return types

stevebauman:
This PR adds some missing return types to methods that previously did not have any, where other methods in the same file do.


freekmurze:
Thank you, Steve!

PR #3623, opened 2024-05-29
Update 'queue_name' configuration to use environment variable

christeyerl:
This PR updates the ﻿`queue_name` configuration in ﻿`spatie/laravel-medialibrary` to make use of the ﻿`MEDIA_QUEUE` environment variable. By leveraging environment variables, we enhance flexibility and simplify configuration management without the need to publish the configuration file every time a value change is required. This change ensures seamless integration with environmental settings while maintaining the ability to customize the queue name as needed.


freekmurze:
Thanks!

Issue #3554, opened 2024-02-27
Media collections are lazy loaded with `preventLazyLoading()` enabled


sebastiaanluca:
This concerns https://github.com/spatie/laravel-medialibrary/pull/2784

I noticed in our old admin panel that we had a lot of lazy loaded media relations. We didn't get a warning about it (for 2 years 😄), even with `Model::preventLazyLoading();` enabled globally. The cause for this is that doing `loadMissing()` on each object in a loop doesn't trigger the exception even though it's technically an n+1 issue.

The solution to this is to always eager load relations manually in the controller, not in models or traits. Removing this would be a breaking change, but are you open to it? I expect a lot of package users have tons of n+1 media issues but aren't aware, unless you manually inspect the debug bar queries for each page you use media in collections.

freekmurze:
I would be open for a PR 👍

freekmurze:
This has been fixed in the latest version.

PR #3627, opened 2024-06-06
Make media lazy loading configurable

alidevweb:
### References

Related issue: https://github.com/spatie/laravel-medialibrary/issues/3554

### Story

In our legacy admin we noticed a lot of lazy loaded media relations that causes an n+1 issue silently without throwing an exception even with `Model::preventLazyLoading()` enabled globally.

The related issue referenced above gives more details.

### Changes

- Add new config parameter `force_lazy_loading` to allow enable/disable media lazy loading. Enabled by default, so it will not require any changes for those happy with the current behavior.
- Update the `InteractsWithMedia` trait to load media based on `force_lazy_loading` value.
- Update the docs to show the new config parameter.
- Add new test to check that media still lazy loaded even with `Model::preventLazyLoading()` enabled globally, but `force_lazy_loading` still on it's default value.
- Add new test to check that it throws a `LazyLoadingViolationException` exception when `Model::preventLazyLoading()` enabled globally and `force_lazy_loading` set to `true`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ phpunit.xml
 .php-cs-fixer.cache
 phpstan.neon
 tests/Support/temp/
+.idea/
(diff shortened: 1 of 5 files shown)


alidevweb:
Hi Spatie team, any feedback on this PR?
Thanks

freekmurze:
Thank you!

PR #3630, opened 2024-06-07
Fix wrong deletions

inalto:
In case of custom paths setup to save the images in the same subdirectory ex. username/post_id/photos/somefile.ext when a file is removed from the media directory it wipes all the directory content.

This commit fixes this problem only deleting the proper images (and related conversions and responsive images) and delete the directory only if is empty.

### DIFF
--- a/src/MediaCollections/Filesystem.php
+++ b/src/MediaCollections/Filesystem.php
@@ -242,7 +242,7 @@ class Filesystem
 
         $responsiveImagePaths = array_filter(
             $allFilePaths,
-            fn (string $path) => Str::contains($path, $conversionName)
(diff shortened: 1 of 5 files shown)


freekmurze:
Hi, could you also add a test to prove that this works?

inalto:
yes, I write one

inalto:
hello, i added some test and spotted some bugs on DefaultFileRemover, now i look at failing tests

inalto:
Problem in case of custom-named media is fixed now. And all the test are passing, included new ones.

freekmurze:
Thank you very much!

PR #3633, opened 2024-06-11
[11.x] Support spatie/pdf-to-image v3

Jubeki:
With the release of spatie/pdf-to-image v3, the spatie media library is no longer compatible with the current implementation.

This PR makes a breaking change by needing v3 instead of v2 and using the corresponding new methods.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -10,6 +10,7 @@ jobs:
       matrix:
         php: [8.3, 8.2]
         laravel: [10.*, 11.*]
+        pdf: [^2.2, ^3.0]
         dependency-version: [prefer-lowest, prefer-stable]
         include:
           - laravel: 10.*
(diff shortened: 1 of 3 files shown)


freekmurze:
I think creating a new major simply for updating the dependency might be a much.

Could you allow for both v2 and v3 to be installed? In code, we should check which version is installed (you can do that with the `InstalledVersions` class from Composer, and then call the right methods for that version. 

This way we can support v3 without making breaking changes.

Jubeki:
@freekmurze yeah, will do that!

Jubeki:
@freekmurze Do you want a new test workflow for testing v2 and v3 or do you want it with a matrix like I did it now?

freekmurze:
Like now is fine. Prefer-lowest will pull in v2, prefer-stable v3. That's good enough for me!

freekmurze:
Thank you!

PR #3635, opened 2024-06-13
Update composer.json

xnekv03:
php require 8.3 or 8.2


freekmurze:
The current code is equivalent.

PR #3638, opened 2024-06-14
Add composer/semver dependency

flightsupport:
Add missing dependency which was introduced with PR Support spatie/pdf-to-image v3 #3633


freekmurze:
Thanks!

PR #3645, opened 2024-06-17
Update preparing-your-model.md

KaanTanis:
compatible with HasMedia::registerMediaConversions(Media|null $media = null)


freekmurze:
Thanks!

PR #3648, opened 2024-06-21
add nullcheck for conversions_disk

DirtyRacer1337:
should fix crash if there is no conversions caused of PR #3630


freekmurze:
Thanks!

PR #3657, opened 2024-06-28
Don't pull all media models from DB when performing cleanup

ImJustToNy:
Hi!
I've noticed that when you do a clean-up, it pulls out all models from the database, just to get the IDs, which is a bit wasteful. This is a small change that changes this behaviour so that it pulls out only IDs.


hellomohsinhello (approved):

freekmurze:
Thanks!

PR #3658, opened 2024-07-02
Added callback to manipulate FileAdder when performing a copy

chrispage1:
This feature adds the ability to customise the FileAdder options when performing a copy action.

In my case, the requirement for this is to set `properties` against the media item, but it gives the user access to all of the FileAdder methods. Example:

```php
$media->copy(
    $oldMedia,
    fileAdderCallback: fn ($adder) => $adder->withProperties([
        'duplicate_of' => $oldMedia->getKey(),
    ])
);
```

Thanks!


freekmurze:
Thank you!

PR #3661, opened 2024-07-07
Check for enum instances when applying manipulations

dennisvandalen:
It seems there is a bug introduced when passing paramaters that are already enum to image manipulations.  https://github.com/spatie/laravel-medialibrary/discussions/3626

The existing code for image manipulations was using a match expression to transform parameters based on the manipulation name. However, the match expression does not perform assignments in the way expected, leading to an error when attempting to convert parameters to enum types. Specifically, the following error was encountered:

```txt
TypeError: Spatie\Image\Enums\Fit::from(): Argument #1 ($value) must be of type string|int, Spatie\Image\Enums\Fit given in /var/www/html/vendor/spatie/laravel-medialibrary/src/Conversions/Manipulations.php:69
```

The root cause was that some parameters were already instances of the enum classes, and attempting to reassign them caused errors.

**Solution Overview**

To address this issue, the match expression was replaced with a switch statement. This allows for more explicit checks and transformations of parameters, ensuring that parameters are only converted if they are not already instances of the corresponding enum classes. The solution involves:

1. Using a switch statement to handle each manipulation case.
2. Checking if the parameter is not already an instance of the enum class before converting it.

For readability reasons I've changed the match to a switch statement.

I've also removed the `pickColor` handling as it is not not to be a valid image manipulation operation.

### DIFF
--- a/src/Conversions/Manipulations.php
+++ b/src/Conversions/Manipulations.php
@@ -59,17 +59,53 @@ class Manipulations
     public function apply(ImageDriver $image): void
     {
         foreach ($this->manipulations as $manipulationName => $parameters) {
-            match ($manipulationName) {
-                'border' => (isset($parameters['type'])) && $parameters['type'] = BorderType::from($parameters['type']),
-                'watermark' => (isset($parameters['fit'])) && $parameters['fit'] = Fit::from($parameters['fit']),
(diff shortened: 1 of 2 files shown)


freekmurze:
Thanks!

PR #3664, opened 2024-07-09
Fix deletion of files without extension

cbaconnier:
Tests are failing in Laravel apps when using fake images generated by  `UploadedFile::fake()->image('image.jpg')` as the file will be generated internally by `tmpfile()` and will be named _(E.g.)_  `phpGoZkJ9`

Currently, it looks if the path contains the filename, including the dot: `phpGoZkJ9.`

I changed it to match the full filename after the forward slash `/`
From my understanding, the `$path` will always have a forward slash (`/`) before the filename. 

If there's a case where doesn't have a forward slash or uses double backslashes `\\` (because Windows),I will happily update my PR to include corresponding scenarios and tests.

Thanks!

### DIFF
--- a/src/Support/FileRemover/DefaultFileRemover.php
+++ b/src/Support/FileRemover/DefaultFileRemover.php
@@ -37,7 +37,7 @@ class DefaultFileRemover implements FileRemover
                     $imagePaths = array_filter(
                         $allFilePaths,
                         function (string $path) use ($media) {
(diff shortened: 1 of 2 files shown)


cbaconnier:
I'm wondering why is there a foreach and a check without the extension though. 
Can a file have multiple extension without being itself a conversion or a responsive image?

freekmurze:
Thanks!

Issue #3669, opened 2024-07-17
Memory exhausted options


patrickomeara:
I have around 700_000 media rows and whenever I run the regenerate command it just hits the memory limit straight away.

I have written my own regenerate command that uses a lazy collection that works well, however I assume others are hitting memory limits as well.

I would like to propose a solution:
* Allow MediaRepository to be swapped out via config
* Ship a LazyMediaRepository that user can opt-in to

Does anyone at Spatie or any users have any feedback, objections or other ideas before I start hacking on this?

freekmurze:
Handled with #3670

PR #3670, opened 2024-07-18
Use Lazy Collection for artisan commands

patrickomeara:
Following on from #3669...

The artisan commands are the only place that the `getBy*()` and `getOrphans*()` methods are used within the package. 

As `MediaRepository` is not yet configurable the only impacted users for this change would be those that replace it in the container, or those that resolve it and expect Eloquent Collections from the`getBy*()` and `getOrphans*()` methods.

Alternatively, we could introduce an interface to bind to and a `repository_class` configuration option, and ship a `CursorMediaRepository` that returns `LazyCollections` from the `getBy*()` and `getOrphans*()` methods. I initially started down this path however, I can't think of a reason that the artisan commands shouldn't always use `LazyCollection`.

If it's decided that the latter option is better, we should take the opportunity of a major release to create a `MediaCollections\MediaRepositories` namespace.

I'm open to feedback.

### DIFF
--- a/src/Conversions/Commands/RegenerateCommand.php
+++ b/src/Conversions/Commands/RegenerateCommand.php
@@ -6,7 +6,7 @@ use Exception;
 use Illuminate\Console\Command;
 use Illuminate\Console\ConfirmableTrait;
 use Illuminate\Support\Arr;
-use Illuminate\Support\Collection;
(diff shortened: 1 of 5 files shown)


freekmurze:
For now, I don't mind those queries are defined multiple times. I'll re-evaluate this when creating a new major version.

PR #3676, opened 2024-07-24
Added ability to programatically specify custom queue on FileAdder

chrispage1:
We've run into an issue where we have thousands of conversions being generated, it blocks up the queue for conversions that should be considered a higher priority.

This small addition allows us to programatically tap into the FileAdder and set a custom queue depending on the job we're performing.

So we can have two queues, one called `conversions` and one called `high_priority_conversions` for example.


freekmurze:
Thanks!

PR #3681, opened 2024-07-25
Adjust sanitization of PHP file extensions in `FileAdder.php`

rmdavi:
I recently noticed sporadically failing tests in my application after the [addition of PHP extensions ](https://github.com/spatie/laravel-medialibrary/commit/e663609647536ec799f3f5477cd232d83f1685fa)to the `defaultSanitizer()` function of [FileAdder.php](https://github.com/spatie/laravel-medialibrary/blob/main/src/MediaCollections/FileAdder.php). 

Digging into the issue, it turns out the `tempnam()` call in `addMediaFromBase64()` of [InteractsWithMedia.php](https://github.com/spatie/laravel-medialibrary/blob/main/src/InteractsWithMedia.php) was occasionally ending the file name string with `PHp` which was blocked by the PHP extension check.

This PR proposes to add the `.` to all `$phpExtensions` in the list to resolve the issue.

### DIFF
--- a/src/MediaCollections/FileAdder.php
+++ b/src/MediaCollections/FileAdder.php
@@ -405,7 +405,7 @@ class FileAdder
         $sanitizedFileName = str_replace(['#', '/', '\\', ' '], '-', $sanitizedFileName);
 
         $phpExtensions = [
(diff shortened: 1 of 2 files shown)


freekmurze:
Nice, thank you!

PR #3682, opened 2024-07-27
Skip path logic when not needed

patrickomeara:
This makes things a bit quicker when dealing with lots of files


freekmurze:
Thanks, but I prefer the current code.

PR #3688, opened 2024-08-08
Correctly remove media folders after removing all media files

StyxUA:
If the media model has conversions and/or responsive images, then `DefaultFileRemover` will not remove the media folder upon removing all files.

The reason of this issue lies in order of removing folders. It tries to remove media folder first, then conversions folder, and then responsive images folder (removing empty directories if there are no files left). Although, conversions/responsive images folders reside inside media folder, thus preventing it from being removed.

This PR fixes this issue by changing the order of removing these folders, moving removing media folder to the last, after all other folders have been removed.

Existing tests weren't catching this issue because they used models without conversions/responsive images. New tests have been added to solve this.


freekmurze:
Thanks!

PR #3696, opened 2024-08-20
Add support for callables in hasMedia() filters

### LINKED ISSUE #3526: hasMedia accept only array
```
hasMedia(string $collectionName = 'default', array $filters = [])
```

Should accept array|callable

crossiatlas:
**Description of Changes**
I encountered a scenario where I expected `$model->hasMedia('my-collection', $filters)` to accept a closure for `$filters`, but it's typehinted to only support an array.

This PR updates the typehint to match `getMedia()`

In the meantime I worked around this via:
```php
$model->getMedia('my-collection', fn (Media $media) => isset($media->custom_properties['something'])))->isNotEmpty()
```

**Changes Made**
* Update `hasMedia()` to support both `array` and `callable` filters, as per `getMedia()`
* Update test coverage

Resolves #3526 

(NOTE: This is my first contribution to this repository, I believe I've met the requirements but feel free to let me know if there's anything I'm missing and I'll update my PR. Thanks!)

### DIFF
--- a/src/InteractsWithMedia.php
+++ b/src/InteractsWithMedia.php
@@ -242,7 +242,7 @@ trait InteractsWithMedia
     /*
      * Determine if there is media in the given collection.
      */
-    public function hasMedia(string $collectionName = 'default', array $filters = []): bool
+    public function hasMedia(string $collectionName = 'default', array|callable $filters = []): bool
     {
         return count($this->getMedia($collectionName, $filters)) ? true : false;
     }
--- a/tests/Feature/InteractsWithMedia/HasMediaTest.php
+++ b/tests/Feature/InteractsWithMedia/HasMediaTest.php
@@ -1,5 +1,7 @@
 <?php
 
+use Spatie\MediaLibrary\MediaCollections\Models\Media;
+
 it('returns false for an empty collection', function () {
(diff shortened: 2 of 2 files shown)


crossiatlas:
@freekmurze Absolutely!

It looks like the `hasMedia()` method isn't documented at all currently (I found it through autocomplete in my IDE), so I assume I'd just need to add this to the [Retrieving Media](https://spatie.be/docs/laravel-medialibrary/v11/basic-usage/retrieving-media) docs?

While I'm at it, `public function getFirstMedia(string $collectionName = 'default', $filters = []): ?Media` isn't typehinted at all, but it calls `getMedia()` anyway. So for consistency I will update it to:
```php
public function getFirstMedia(string $collectionName = 'default', array|callable $filters = []): ?Media
```

I don't think that change will require additional test coverage, `getFirstMedia()` already has tests for filters (array and callable). The only difference a user may experience is the `TypeError` they get by passing an invalid `$filters` type would come from `getFirstMedia()` instead of `getMedia()`. Same result, just one level higher.

As far as I can see those are the only methods dependent on `getMedia()` so we shouldn't need to change any others.

timvandijck:
Looks good to me. Thanks for the contribution!

PR #3698, opened 2024-08-22
Added config variable to disable dispatching jobs after commit

chrispage1:
There are some use cases where the PR #3687 that ensures jobs are dispatched after database transactions may cause undesirable behaviour, particularly in tests which are run within a database transaction for example.

This PR introduces a new configuration variable, which allows disabling this behaviour.


freekmurze:
Thanks!

PR #3699, opened 2024-08-24
Add Support to `ULID` column type

devajmeireles:
Based on several studies and personal tests, we can prove the effectiveness of `ulid` columns instead of `uuid` when talking about a large volume of data.

A package like this tends to persist a thousand of data in the database since all the media of an entire application is persisted in a table. Some public studies point to abstracts where ULID is advantageous in these cases.

> "ULID is often considered superior to UUID when ordering and efficiency are important. It is lexicographically orderable, which makes it easier to organize data in chronological order, something that UUIDs cannot do efficiently. In addition, ULID is more compact, at 26 characters, compared to UUID's 36 characters, making it easier to read and transmit. Embedding a timestamp in ULID means that IDs generated at close points in time will be ordered in a natural way, which can improve database performance by minimizing index fragmentation. While UUIDs are widely used and supported, ULID offers specific advantages in terms of temporal ordering and storage efficiency, making it a preferred choice in many systems that prioritize these characteristics."

A good read is here: https://medium.com/@sammaingi5/uuid-vs-ulid-how-ulid-improves-write-speeds-d16b23505458

---

This PR is an attempt to implement a configuration that allows ULID to be used instead of UUID.

### DIFF
--- a/config/media-library.php
+++ b/config/media-library.php
@@ -15,7 +15,7 @@ return [
     'max_file_size' => 1024 * 1024 * 10, // 10MB
 
     /*
-     * Determiner if the media table should use ULID instead of UUID.
+     * Determine if the media table should use ULID instead of UUID.
      */
     'use_ulid_column' => env('MEDIA_USE_ULID_COLUMN', false),
 
--- a/config/media-library.php
+++ b/config/media-library.php
@@ -15,6 +15,11 @@ return [
     'max_file_size' => 1024 * 1024 * 10, // 10MB
 
     /*
+     * Determiner if the media table should use ULID instead of UUID.
+     */
+    'ulid' => env('MEDIA_ULID', true),
+
+    /*
      * This queue connection will be used to generate derived and responsive images.
      * Leave empty to use the default queue connection.
      */
--- a/config/media-library.php
+++ b/config/media-library.php
@@ -17,7 +17,7 @@ return [
     /*
      * Determiner if the media table should use ULID instead of UUID.
      */
-    'ulid' => env('MEDIA_ULID', false),
+    'use_ulid_column' => env('MEDIA_USE_ULID_COLUMN', false),
 
     /*
      * This queue connection will be used to generate derived and responsive images.
--- a/config/media-library.php
+++ b/config/media-library.php
@@ -17,7 +17,7 @@ return [
     /*
      * Determiner if the media table should use ULID instead of UUID.
      */
-    'ulid' => env('MEDIA_ULID', true),
+    'ulid' => env('MEDIA_ULID', false),
 
     /*
      * This queue connection will be used to generate derived and responsive images.
--- a/docs/installation-setup.md
+++ b/docs/installation-setup.md
@@ -51,6 +51,11 @@ return [
      * Adding a larger file will result in an exception.
      */
     'max_file_size' => 1024 * 1024 * 10,
+
+    /*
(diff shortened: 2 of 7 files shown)


devajmeireles:
Hey, @freekmurze ! I'd like to know your opinion on this before I continue with the PR :D There doesn't seem to be any breaking changes, but I'd like your opinion on whether I should move forward because it's something interesting in your eyes.

patinthehat:
@devajmeireles please feel free to continue with this PR. I'm happy to review it once it's ready :+1: Thank you!

devajmeireles:
> @devajmeireles please feel free to continue with this PR. I'm happy to review it once it's ready 👍 Thank you!

Hey, @patinthehat ! Thanks for your feedback. I think the PR is ready for review. Feel free to review it and point out anything misaligned or incorrect so I can adjust it.

src/MediaCollections/Models/Media.php
```diff
@@ -66,6 +68,7 @@ class Media extends Model implements Attachable, Htmlable, Responsable
     use CustomMediaProperties;
     use HasUuid;
     use IsSorted;
+    use DetermineUniqueIdentifierColumn;
```
patinthehat, line 71: nitpick: please add this in alpha order

patinthehat (changes requested):
Please update the docs to explain the new configuration setting/environment variable. Other than that and one other minor change, looks good!

devajmeireles:
Hey, @patinthehat ! I applied all your suggestions. I also tried searching for `UUID` (or variations, like: `uuid`) in the docs, to write something special for someone who wants to use ULID instead of UUID, but the docs don't mention anything related to that, so I just updated the `installation.md` docs with the new configuration value.

config/media-library.php
```diff
@@ -14,6 +14,11 @@
      */
     'max_file_size' => 1024 * 1024 * 10, // 10MB
 
+    /*
+     * Determiner if the media table should use ULID instead of UUID.
```
patinthehat, line 18: nitpick: Typo in "Determine" :smile:

docs/installation-setup.md
```diff
@@ -51,6 +51,11 @@ return [
      * Adding a larger file will result in an exception.
      */
     'max_file_size' => 1024 * 1024 * 10,
+
+    /*
+     * Determiner if the media table should use ULID instead of UUID.
```
patinthehat, line 56: nitpick: typo

patinthehat (changes requested):
Thanks for the updates! Would you mind adding an additional markdown file in "advanced usage" that explains the configuration option and what it does when `true` and `false`?

config/media-library.php
devajmeireles, line 18:
```suggestion
     * Determine if the media table should use ULID instead of UUID.
```

docs/installation-setup.md
devajmeireles, line 56:
```suggestion
     * Determine if the media table should use ULID instead of UUID.
```

timvandijck:
Hi @devajmeireles I'm sorry it took a while to come back on this PR. 
We had a little discussion in our team and agreed that we will not be merging the PR. 

While we acknowledge the benefits of ULID  we believe it's better to keep simplicity in mind and not offer too many options. Users can still extend the Media model to modify the current implementation.

PR #3704, opened 2024-09-01
fix: make media-library:clean use of media prefix

ImJustToNy:
Hi!
I wanted to use `media-library:clean` command, and I've noticed that it ignores the `media-library.prefix` config value. This PR fixes that.


freekmurze:
Thanks!

PR #3707, opened 2024-09-10
Fix renaming files with custom model

esadewater:
This PR attempts to fix file renaming when using a custom model with a primary key name other than 'id'.

We use a custom model that changes the primary key name from `id` to `media_id`:
```php
<?php

namespace App\Models;

use Illuminate\Database\Eloquent\Concerns\HasUuids;
use Spatie\MediaLibrary\MediaCollections\Models\Media as BaseMedia;

class Media extends BaseMedia
{
    use HasUuids;

    protected $primaryKey = 'media_id';
}
```

We want to rename a file as described in the docs ([Retrieving media](https://spatie.be/docs/laravel-medialibrary/v11/basic-usage/retrieving-media)):

```php
$mediaItems[0]->file_name = 'newFileName.jpg';
$mediaItems[0]->save();
```

When renaming file `src/MediaCollections/Filesystem.php` tries to find a media model using the `id` attribute instead of the actual primary key. This PR fixes this bug by using `->getKey()` in this specific place, which works in both cases.

### DIFF
--- a/src/MediaCollections/Filesystem.php
+++ b/src/MediaCollections/Filesystem.php
@@ -298,7 +298,7 @@ class Filesystem
 
     protected function renameConversionFiles(Media $media): void
     {
-        $mediaWithOldFileName = config('media-library.media_model')::find($media->id);
+        $mediaWithOldFileName = config('media-library.media_model')::find($media->getKey());
         $mediaWithOldFileName->file_name = $mediaWithOldFileName->getOriginal('file_name');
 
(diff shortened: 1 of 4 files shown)


patinthehat:
@esadewater would you mind adding a unit test that makes sure both `id` and a custom `id` value continue to work as expected? Thanks!

timvandijck:
Thanks for the PR!

PR #3715, opened 2024-10-20
Correctly search media conversion by name

StyxUA:
This PR fixes `ConversionCollection`'s behavior in finding conversion by name.

If a subject model has multiple collections, and multiple conversions with the same name, but applied to different collections, then `ConversionCollection` incorrectly determines the conversion which should be applied to the media.

Example:
```php
{
    $this->addMediaCollection('avatar')->acceptsMimeTypes(['image/png'])->singleFile();
    $this->addMediaCollection('signature')->acceptsMimeTypes(['image/jpeg'])->singleFile();
}

{
    $this
        ->addMediaConversion('preview')
        ->fit(Fit::Crop, 50, 50)
        ->performOnCollections('avatar')
        ->format('png');

    $this
        ->addMediaConversion('preview')
        ->fit(Fit::Crop, 300, 100)
        ->performOnCollections('signature')
        ->format('jpeg');
}

```

Now, whereas conversion files are generated correctly, their URLs and paths are wrong, because `ConversionCollection` doesn't honor media collection, getting just the first conversion which name matches:

```php
$media = $model->getFirstMedia('signature');

ConversionCollection::createForMedia($media)
    ->getByName('preview') // <-- issue occurs here, first conversion with this name is fetched
    ->getManipulations()
    ->toArray();
```

Result:
```
[
  "format" => [
    "png",
  ],
  "fit" => [
    Spatie\Image\Enums\Fit {#4033
      +name: "Crop",
      +value: "crop",
    },
    50,
    50,
  ],
]
```

Generated URL is wrong too, because internally the fetched conversion is used to determine file extension:
```php
$media->getFirstMediaUrl('signature', 'preview');
```
Result:
```
https://.../media/2/c/test-preview.png <-- should be .jpeg
```

This PR fixes this behavior. Specifically, `ConversionCollection::getByName()` function first filters conversions by media's collection, and only then searches by conversion name.

Tests included.

### DIFF
--- a/src/Conversions/ConversionCollection.php
+++ b/src/Conversions/ConversionCollection.php
@@ -38,9 +38,11 @@ class ConversionCollection extends Collection
 
     public function getByName(string $name): Conversion
     {
-        $conversion = $this->first(fn (Conversion $conversion) => $conversion->getName() === $name);
+        $conversion = $this
+            ->getConversions($this->media->collection_name)
+            ->first(fn (Conversion $conversion) => $conversion->getName() === $name);
 
-        if (! $conversion) {
+        if (!$conversion) {
             throw InvalidConversion::unknownName($name);
         }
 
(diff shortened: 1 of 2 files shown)


timvandijck:
Thanks for the PR, looks like great work!

PR #3717, opened 2024-10-23
FileNamer should be used while removing conversions and/or responsive images

### LINKED ISSUE #3702: Cant delete files after using a customer FileNamer
Hello @Spatie, great package you have here.

Its probably documented somewhere but I am having issues deleting obsolete files when I add new ones.
I defined a collection to keep only 3 files.
```php
    {
        $this->addMediaCollection('cover')
            ->onlyKeepLatest(3)
            ->useDisk('s3')
            ->acceptsFile(function (File $file) {
                return Str::of($file->mimeType)->contains('image');
            });
    }
```

I then created a custom file namer, I do not want my image and its conversions to share a name. A user can reverse engineer and I am trying to avoid that.
```php
class FileNamer extends SpatieFileNamer
{
    public function conversionFileName(string $fileName, Conversion $conversion): string {
        $strippedFileName = pathinfo($fileName, PATHINFO_FILENAME);

        return md5($strippedFileName)."-{$conversion->getName()}";
    }

    public function originalFileName(string $fileName): string
    {
        $extLength = strlen(pathinfo($fileName, PATHINFO_EXTENSION));

        $baseName = substr($fileName, 0, strlen($fileName) - ($extLength ? $extLength + 1 : 0));

        return md5($baseName);
    }
}
```

Basically I am md5ing the original file name and the resultant file name so they cant be similar.
Examples are shown below. 
image
image

The furthest I could go investigating was upto these lines below when deleting the conversions, still to inspect the other removers.
https://github.com/spatie/laravel-medialibrary/blob/548626199039f41d561fb01674f49a3f4534244e/src/Support/FileRemover/DefaultFileRemover.php#L68-L79

According to L72, only deletable conversations are files with a naming pattern like `filename-conversionname` which is not whats not set on my new file namer.

One can write a new `file_remover_class` or wire an extra media delete observer but cant we have this natively in the package.

Thanks

StyxUA:
Fixes #3702

### DIFF
--- a/src/MediaCollections/Filesystem.php
+++ b/src/MediaCollections/Filesystem.php
@@ -11,6 +11,7 @@ use Spatie\MediaLibrary\MediaCollections\Events\MediaHasBeenAddedEvent;
 use Spatie\MediaLibrary\MediaCollections\Exceptions\DiskCannotBeAccessed;
 use Spatie\MediaLibrary\MediaCollections\Models\Media;
 use Spatie\MediaLibrary\Support\File;
+use Spatie\MediaLibrary\Support\FileNamer\FileNamer;
 use Spatie\MediaLibrary\Support\FileRemover\FileRemoverFactory;
 use Spatie\MediaLibrary\Support\PathGenerator\PathGeneratorFactory;
 use Spatie\MediaLibrary\Support\RemoteFile;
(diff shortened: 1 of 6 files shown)


timvandijck:
Again, thanks for the PR! 🙌

PR #3721, opened 2024-11-02
Don't copy the file if no conversion is needed

patrickomeara:
I'm processing over 700k files with media library now and this sticks out as an easy optimisation.


timvandijck:
Looks like a good improvement, thanks!

PR #3722, opened 2024-11-04
Add undocumented `storeConversionsOnDisk()` function

djl997:
Add the `storeConversionsOnDisk` function to the documentation of registering mediaconversions. Not sure this is the right place since it is also very much related to registering mediaCollections? In that case I think the best place is `docs/working-with-media-collections/defining-media-collections.md#registering-media-conversions`.


timvandijck:
Thanks!

PR #3743, opened 2024-12-09
Fix to broken tests for Responsive Media Conversions.

aSeriousDeveloper:
Fix for my broken tests from #3742 .

I have double checked that these tests work correctly on my end. There shouldn't be any change in underlying functionality.

I apologise for not properly & thoroughly checking my previous PR w/rt the tests themselves.


freekmurze:
Thank you!

PR #3759, opened 2025-01-02
Fix confusing documentation for Media::setNewOrder method

### LINKED ISSUE #3758: Misleading documentation for Media::setNewOrder method
The [documentation](https://github.com/spatie/laravel-medialibrary/blob/0060425ee4082580a3071f7d377510517905dff6/docs/advanced-usage/ordering-media.md?plain=1#L12) states:

	“the record with the first id in the array will get order 11, the record with the second id will get order 2, …”

This is incorrect and confusing.
The actual behavior, as per the [code](https://github.com/spatie/laravel-medialibrary/blob/0060425ee4082580a3071f7d377510517905dff6/src/MediaCollections/Models/Concerns/IsSorted.php#L29), assigns the first record the startOrder (default is 1), the second record gets startOrder + 1, and so on.

rottifant:
fixes #3758

### DIFF
--- a/docs/advanced-usage/ordering-media.md
+++ b/docs/advanced-usage/ordering-media.md
@@ -10,9 +10,10 @@ You can easily reorder a list of media by calling  ̀Media::setNewOrder`:
 ```php
  /**
   * This function reorders the records: the record with the first id in the array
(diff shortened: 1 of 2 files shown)


freekmurze:
Thanks!

PR #3777, opened 2025-02-04
Improve return types when using a custom Media model

sforward:
This PR improves several return types when the Media model is extended (using the media_model config). With these generic types PHPStan can correctly resolve the custom Media model type.


freekmurze:
Thanks!

christeyerl:
So, from now I have to add generics befere each use of that trait?
image

```php
/** @use InteractsWithMedia<Media> */
use InteractsWithMedia;
```

sforward:
I will submit another PR to add a default template value. Sorry for the inconvenience.

PR #3779, opened 2025-02-05
Add default types for template

sforward:
This is an improvement of https://github.com/spatie/laravel-medialibrary/pull/3777, so it is no longer required to set the generic value.


freekmurze:
Thanks!

PR #3780, opened 2025-02-09
Add @return $this annotation to methods

lorenzolosa:
#3777 and #3779 added generics to `InteractsWithMedia` and `FileAdder`, which improves PHPStan recognition of return types. However, PHPStan is still unable to follow through chaining; for instance, it will interpret the return value of:
```
$this->addMedia($file)
    ->usingFileName($filename)
    ->toMediaCollection('collection');
```
as the base `Media` class, even if an extended class has been specified.

This PR fixes that by listing the return type of a bunch of methods as `$this`. These include methods in the `FileAdder` class, needed for the example above; but I have also included several others (mostly with generics, where I believe it's most useful, but not only).

### DIFF
--- a/src/Conversions/ConversionCollection.php
+++ b/src/Conversions/ConversionCollection.php
@@ -23,6 +23,9 @@ class ConversionCollection extends Collection
         return (new static)->setMedia($media);
     }
 
+    /**
+     * @return $this
+     */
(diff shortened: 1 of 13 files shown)


freekmurze:
Thanks!

Issue #3665, opened 2024-07-12
Wrong order of manipulations


Jackardios:
The documentation says that when using `withManipulations` ["The package will regenerate all files (conversions) using the saved manipulation as the first manipulation when creating each derived image."](https://spatie.be/docs/laravel-medialibrary/v11/advanced-usage/storing-media-specific-manipulations) But in fact this is not the case, all manipulations that are added through `withManipulations` will be executed last. 

For example:
```php
class SomeModel extends Model implements HasMedia
{
    use InteractsWithMedia;

    public function registerMediaCollections(): void
    {
        $this->addMediaCollection('image')->singleFile();
    }

    public function registerMediaConversions(Media $media = null): void
    {
        $this->addMediaConversion('cropped')
            ->fit(Fit::Crop, 300, 300)
            ->performOnCollections('image');
    }
}

$someModel = SomeModel::create();
$addedMedia = $someModel->addMedia($mediaFile)
  ->storingConversionsOnDisk(config('media-library.disk_name'))
  ->withManipulations([
      'cropped' => [
          'manualCrop' => [500, 300, 200, 150],
      ],
  ])
  ->toMediaCollection('image');

$conversions = Spatie\MediaLibrary\Conversions\ConversionCollection::createForMedia($addedMedia)
  ->filter(fn (Spatie\MediaLibrary\Conversions\Conversion $conversion) => $conversion->shouldBePerformedOn($addedMedia->collection_name));

dd($conversions->first()->getManipulations()->toArray());
```

```
[
  "optimize" => [
    Spatie\ImageOptimizer\OptimizerChain {#7285},
  ],
  "format" => [
    "jpg",
  ],
  "fit" => [
    Spatie\Image\Enums\Fit {#7532
      +name: "Crop",
      +value: "crop",
    },
    300,
    300,
  ],
  "manualCrop" => [500, 300, 200, 150],
]
```

I'm guessing this is due to an incorrect array merge in `addAsFirstManipulations` method: 
https://github.com/spatie/laravel-medialibrary/blob/4ebb5642323aa7399cc6a5924b8245adc1862678/src/Conversions/Conversion.php#L140
Instead it should be:
```php
$allManipulations = array_merge($newManipulations, Arr::except($currentManipulations, array_keys($newManipulations)));
```

If this is indeed a bug, I can create a PR to fix it

freekmurze:
I'm be open for a PR that fixes this. Make sure to include tests so this won't break again in the future.

PR #3783, opened 2025-02-11
fix: media manipulations should be applied before media conversions

JavoByte:
This issue was originally opened here #3665 .

This PR fixes it.

It was not originally caught because test assertions `assertEquals` doesn't check ordering for associative arrays. 

I added tests to assert ordering of array keys. 

Also added logic to make `*` manipulations to be applied before anything else. Otherwise 

```
    $this->media->manipulations = [
        'thumb' => ['greyscale' => [], 'height' => [10]],
        '*' => ['manualCrop' => [600, 400, 0, 0]],
    ];

```

would apply `greyscale` and `height` before anything else. The `manualCrop` defined in `*` should be applied before `*`.

### DIFF
--- a/src/Conversions/Conversion.php
+++ b/src/Conversions/Conversion.php
@@ -3,6 +3,7 @@
 namespace Spatie\MediaLibrary\Conversions;
 
 use Illuminate\Support\Arr;
+use Illuminate\Support\Facades\Log;
 use Illuminate\Support\Traits\Conditionable;
 use Spatie\ImageOptimizer\OptimizerChainFactory;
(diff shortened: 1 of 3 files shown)


freekmurze:
It seems the tests are failing. Could you take a look?

JavoByte:
Currently this is a strange case of "it works on my machine".

I am trying to run tests replicating pipelines environment. I hope to reproduce that way. This might take a bit longer.

JavoByte:
I rebased and ran all of the tests with `act` instead of my local php installation. All of the tests were successful.

freekmurze:
Thanks!

PR #3794, opened 2025-02-26
Add responsive images temporary url support

enricodelazzari:


freekmurze:
Could you also add tests and update the docs?

PR #3795, opened 2025-03-11
Dynamically check instance of temporary upload model

sebastiaanluca:
This PR makes the `instanceof` check for the temporary upload model dynamic by retrieving the model from the config instead of hardcoding it. This allows using a model that does not extend the base class.

### DIFF
--- a/src/MediaCollections/FileAdder.php
+++ b/src/MediaCollections/FileAdder.php
@@ -123,7 +123,7 @@ class FileAdder
             return $this;
         }
 
-        if ($file instanceof TemporaryUpload) {
+        if ($file instanceof (config('media-library.temporary_upload_model'))) {
             return $this;
         }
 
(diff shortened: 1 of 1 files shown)


freekmurze:
Nice, thanks!

ziming:
This is causing issues for me in prod because I had that config value to null (to avoid phpstan or ide helper telling me I don't have that model as I am not using media library pro)

sylouuu:
> This is causing issues for me in prod because I had that config value set to null (to avoid phpstan or ide helper telling me I don't have that model as I am not using media library pro)

Same here, we have `null` for `temporary_upload_model` config so we have errors in our tests suite. Could you handle `null` value please?

Bests

sebastiaanluca:
Added a check in https://github.com/spatie/laravel-medialibrary/pull/3801.

PR #3798, opened 2025-03-25
Add conversion parameter to toResponse and toInlineResponse

miccehedin:
Also add convenience methods toAvailableResponse and toAvailableInlineResponse

This way you can stream converted media, useful when the url:s are not suitable for public access.


freekmurze (commented):
Could you also add docs for this?

freekmurze:
Could you make it so that if the conversion does not exists, that an exception is thrown?

PR #3802, opened 2025-04-01
Add check for missing files referenced by media model

miccehedin:
Optionally delete media models where the file is missing.

This will identify the problem in case the file has been removed from storage by another means.


freekmurze:
To keep the package simple, I'm going to pass on this for now.

PR #3818, opened 2025-05-16
Add support for LastMedia functions

bAngerman:
Getting the last media item, which is also the most recent media item, is also useful when a collection is used for storing multiple versions of a single file.

### DIFF
--- a/src/InteractsWithMedia.php
+++ b/src/InteractsWithMedia.php
@@ -12,6 +12,7 @@ use Illuminate\Support\Facades\Validator;
 use Illuminate\Support\Str;
 use Spatie\MediaLibrary\Conversions\Conversion;
 use Spatie\MediaLibrary\Downloaders\DefaultDownloader;
+use Spatie\MediaLibrary\Enums\CollectionPosition;
 use Spatie\MediaLibrary\MediaCollections\Events\CollectionHasBeenClearedEvent;
 use Spatie\MediaLibrary\MediaCollections\Exceptions\InvalidBase64Data;
 use Spatie\MediaLibrary\MediaCollections\Exceptions\InvalidUrl;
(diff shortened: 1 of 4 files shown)


src/InteractsWithMedia.php
```diff
@@ -289,24 +289,32 @@ public function getMediaModel(): string
         return config('media-library.media_model');
     }
 
+    private function getMediaItem(string $collectionName, $filters, string $firstOrLast)
```
freekmurze, line 292: Make this one protected, and move this under the last function in this file that uses this.

src/InteractsWithMedia.php
```diff
@@ -289,24 +289,32 @@ public function getMediaModel(): string
...
-    public function getFirstMediaUrl(string $collectionName = 'default', string $conversionName = ''): string
+    public function getLastMedia(string $collectionName = 'default', $filters = []): ?Media
     {
-        $media = $this->getFirstMedia($collectionName);
+        return $this->getMediaItem($collectionName, $filters, 'last');        
+    }
+
+    private function getMediaItemUrl(string $collectionName, string $conversionName, string $firstOrLast): string
+    {
+        $media = $firstOrLast === 'first' ? $this->getFirstMedia($collectionName) : $this->getLastMedia($collectionName);
```
freekmurze, line 317: Reformat this so `?` and `:` go on their own line.

src/InteractsWithMedia.php
freekmurze, line 292: Instead of a string, introduce a new enum called `CollectionPosition`, with `first` and `last` cases.

freekmurze (changes requested):
Thanks for your work on this.

bAngerman:
Thanks for taking a look @freekmurze 

I've made those changes, let me know if additional changes are needed?

freekmurze:
Thanks!

Issue #3733, opened 2024-11-27
Octane file upload issue


uvarats:
I am using spatie/laravel-medialibrary for model files upload and Octane + RoadRunner as a server. Files are uploaded via the Backpack admin panel (i dont think, that this is really important).

I faced some issue, when uploading image for new model (when model does not exists in db). After several model creates with file upload, there is error, that looks like:

<details>
    <summary>Stacktrace</summary

[2024-11-27 15:43:31] local.ERROR: fopen(/app/storage/app/public/backpack/temp/1480184708144619212-ZIxE.jpg): Failed to open stream: No such file or directory {"exception":"[object] (ErrorException(code: 0): fopen(/app/storage/app/public/backpack/temp/1480184708144619212-ZIxE.jpg): Failed to open stream: No such file or directory at /app/vendor/spatie/laravel-medialibrary/src/MediaCollections/Filesystem.php:138)
[stacktrace]
#0 /app/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php(256): Illuminate\\Foundation\\Bootstrap\\HandleExceptions->handleError(2, 'fopen(/app/stor...', '/app/vendor/spa...', 138)
#1 [internal function]: Illuminate\\Foundation\\Bootstrap\\HandleExceptions->Illuminate\\Foundation\\Bootstrap\\{closure}(2, 'fopen(/app/stor...', '/app/vendor/spa...', 138)
#2 /app/vendor/spatie/laravel-medialibrary/src/MediaCollections/Filesystem.php(138): fopen('/app/storage/ap...', 'r')
#3 /app/vendor/spatie/laravel-medialibrary/src/MediaCollections/Filesystem.php(29): Spatie\\MediaLibrary\\MediaCollections\\Filesystem->copyToMediaLibrary('/app/storage/ap...', Object(App\\Models\\Media), NULL, '148018470814461...')
#4 /app/vendor/spatie/laravel-medialibrary/src/MediaCollections/FileAdder.php(470): Spatie\\MediaLibrary\\MediaCollections\\Filesystem->add('/app/storage/ap...', Object(App\\Models\\Media), '148018470814461...')
[...]

PR #3826, opened 2025-06-18
Update FileAdder.php

imitronov:
Fix: Prevent model caching issue in Roadrunner event handler

This commit addresses an issue where the model was being cached by Roadrunner when processing the `created` event. This caused incorrect behavior when adding media to the model.

The fix replaces `$model` with `$this->subject` within the `processUnattachedMedia` callback to ensure the correct model instance is used.


cinsort:
@imitronov better assign PR to issue https://github.com/spatie/laravel-medialibrary/issues/3733

PR #3832, opened 2025-08-04
Feat: set default expiration minutes when no value passes in args for temp url

chinmaypurav:
Most of the time I work with temporary URLs when working with private media. It is very cumbersome to set the expiration time for every `$model->getFirstTemporaryUrl(now()->addMinutes(5))` 

To overcome this, I am submitting this PR, where, if no arg is passed, a default time is set, can be configured from config.

This want I can do

```php
// Before
$model->getFirstTemporaryUrl(now()->addMinutes(5));

// After
$model->getFirstTemporaryUrl();
```

### DIFF
--- a/config/media-library.php
+++ b/config/media-library.php
@@ -215,6 +215,12 @@ return [
      */
     'media_downloader_ssl' => env('MEDIA_DOWNLOADER_SSL', true),
 
+    /*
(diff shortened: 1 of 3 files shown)


freekmurze:
Thanks!

PR #3834, opened 2025-08-05
Add native `copyObject` API support for R2 disks

kha333n:
This PR adds optional support for native copyObject when using R2-compatible S3 disks.

- Introduced `r2_disk` config to define is that R2 disk and use `copyObject`.
- If the disk supports getClient and getBucket, the library will use native S3 copyObject.
- Falls back to default copy method if unavailable.

This improves performance and avoids unnecessary file downloads/uploads.

### DIFF
--- a/config/media-library.php
+++ b/config/media-library.php
@@ -11,9 +11,7 @@ return [
     /*
      * If using R2 (Cloudflare), set your disk key below to true. Enables optimized copy behavior.
      */
-    'r2_disks' => [
-        'r2' => true, // disk name => true
-    ],
+    'r2_disk' => false,
 
     /*
      * The maximum file size of an item in bytes.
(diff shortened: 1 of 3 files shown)


kha333n:
If you have some better idea in mind, you are welcomed to guide me and i'll update it. @freekmurze

freekmurze:
Maybe instead of adding a global option in our config file, let's make it work by letting people adding a configuration option at the configured disk itself, so we can read there.

A bit similar of how we ask people to ask to add dump configuration on the db connecting itself: https://spatie.be/docs/laravel-backup/v9/installation-and-setup#content-skipssl-in-mysqlmariadb-database-connection

kha333n:
@freekmurze That's a good point. I'll update it.

And thanks for the guidance.

PR #3846, opened 2025-08-30
Adding support for `.m4v` video format

nexxai:
I have a use case where I am downloading `.m4v` files from a remote server and need to then generate thumbnails of them.

ffmpeg and ffprobe (which are used directly by this package) both already support this video format so it's just a matter of adding their extension and MIME-type to the package to allow it to see them as acceptable video files.


freekmurze:
If we get all the test passing, I'd consider merging this in.

PR #3862, opened 2025-10-24
Add timeout and threads configuration for FFMpeg

francoism90:
Configuration enhancements:

* Added support for a configurable FFmpeg timeout using the `media-library.ffmpeg_timeout` setting, with a default value of 900 seconds.
* Added support for configurable FFmpeg thread count using the `media-library.ffmpeg_threads` setting, with a default value of 0 (auto).

I receive a lot of `ffmpeg 300 timeout process` errors, because my system sometimes cannot handle the amount of work, and needs more time. See https://github.com/PHP-FFMpeg/PHP-FFMpeg#ffmpeg how these options are defined.

### DIFF
--- a/config/media-library.php
+++ b/config/media-library.php
@@ -193,6 +193,18 @@ return [
     'ffprobe_path' => env('FFPROBE_PATH', '/usr/bin/ffprobe'),
 
     /*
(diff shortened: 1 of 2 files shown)


freekmurze:
Thanks!

PR #3869, opened 2025-11-06
fix: duration should be parsed as float

francoism90:
I notice the parsed duration may be a string, e.g. `"10.50"`, instead of `10.50`.

I also introduced Number class, which is introduced two years ago: https://laravel-news.com/laravel-number-utility-class - if this is unwanted, I can switch to a PHP alternative instead.

This PR also adds https://github.com/spatie/laravel-medialibrary/pull/3846.


freekmurze:
Could you resolve the conflicts?

PR #3877, opened 2025-11-14
Add `getAvailablePathRelativeToRoot()` method

adriaanzon:
Adds a method similar to `getAvailablePath()` that calls `getPathRelativeToRoot()` instead of `getPath()`.

This allows code like the following, in cases where you use the `local` filesystem driver in your development environment and the `s3` driver in production, as `getPath()` returns the absolute path which you cannot pass to `Storage::get()`.

```php
Storage::disk($media->disk)->get(
    $media->getAvailablePathRelativeToRoot(['conversion'])
)
```

### DIFF
--- a/src/MediaCollections/Models/Media.php
+++ b/src/MediaCollections/Models/Media.php
@@ -175,6 +175,19 @@ class Media extends Model implements Attachable, Htmlable, Responsable
         return $this->getPath();
     }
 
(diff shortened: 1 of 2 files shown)


freekmurze:
Some of the tests are failing, could you have a look?

adriaanzon:
This is an issue in Testbench: https://github.com/orchestral/testbench-core/issues/368

PR #3895, opened 2025-12-12
Fix PHPDoc property name to use snake_case

michael-tscholl:
Updated the @property-read annotation for human readable size from $humanReadableSize to $human_readable_size to align with Laravel's attribute naming conventions. Added tests to verify the attribute is accessible and returns the correct format.


freekmurze:
Thanks!

PR #3927, opened 2026-04-05
Fix awkward phrasing: 'want to not move, but copy' -> clearer wording

tareq-halaby:
The sentence 'If you want to not move, but copy, the original file you can call' was awkward and grammatically poor. Rewritten as 'If you want to copy the original file rather than move it, you can call' for clarity.


freekmurze:
Thanks!

PR #3929, opened 2026-04-07
Add onlyKeepFirst() method to MediaCollection

netjordanlee:
- Add onlyKeepFirst() method to MediaCollection to allow rejecting new additions once the collection size limit is reached, while keeping the first N items in the collection.
- Update the onlyKeepLatest() method to set the keepFirst property to false, ensuring that existing behavior is maintained unless onlyKeepFirst() is explicitly called.
- Update FileAdder to check for the keepFirst property when adding files to a collection with a size limit.
- Add a new exception FileCannotBeAddedToFullCollection to handle cases where a file cannot be added to a collection that has reached its size limit with keepFirst enabled.
- Update tests to cover the new functionality of onlyKeepFirst() in MediaCollection and the corresponding behavior in FileAdder.

### DIFF
--- a/src/MediaCollections/Exceptions/FileCannotBeAddedToFullCollection.php
+++ b/src/MediaCollections/Exceptions/FileCannotBeAddedToFullCollection.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace Spatie\MediaLibrary\MediaCollections\Exceptions;
+
+use Spatie\MediaLibrary\HasMedia;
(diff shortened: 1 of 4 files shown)


netjordanlee:
I have created this primarily for my personal use case but I figured it may be useful for others.

I wanted to limit the size of the media collection but I didn't want to cycle out the oldest images as this causes potentially undesirable data loss.

freekmurze:
To keep the package simple, I'm going to pass on this one for now. Thanks!

PR #3934, opened 2026-04-22
fix: skip URL encoding for S3 disk to prevent double encoding

### LINKED ISSUE #3933: Fix double encoding of percent signs in filenames when using S3 disk
## Problem
The fix in #3930 introduced a regression for S3 disks.

When a filename contains `%` (e.g. `IMG_5405%20copy.jpg`), the URL comes out 
triple-encoded on S3: IMG_5405%252520copy.jpg

This happens because the S3 SDK already encodes URLs internally, so our `rawurlencode()` on top causes double encoding.

Local disk is fine, only S3 is affected.

## Proposed Fix
Skip encoding in `getUrlEncodedPathRelativeToRoot()` when the disk driver is `s3`.

moemadeldin:
Fixes #3933

### DIFF
--- a/src/Support/UrlGenerator/BaseUrlGenerator.php
+++ b/src/Support/UrlGenerator/BaseUrlGenerator.php
@@ -73,6 +73,12 @@ abstract class BaseUrlGenerator implements UrlGenerator
 
     protected function getUrlEncodedPathRelativeToRoot(): string
     {
+        $driver = config("filesystems.disks.{$this->getDiskName()}.driver");
+        
+        if ($driver === 's3') {
+            return $this->getPathRelativeToRoot();
+        }
+        
         $segments = explode('/', $this->getPathRelativeToRoot());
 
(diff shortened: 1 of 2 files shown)


src/Support/UrlGenerator/BaseUrlGenerator.php
```diff
@@ -73,6 +73,12 @@ protected function getDisk(): Filesystem
 
     protected function getUrlEncodedPathRelativeToRoot(): string
     {
+        $driver = config("filesystems.disks.{$this->getDiskName()}.driver");
+        
+        if ($driver === 's3') {
```
francoism90: Hmm, what if you did a local check instead? Laravel does support additional fs drivers, I don't know if they also should be skipped?

moemadeldin: switched to checking for `local` instead.

freekmurze:
Thanks!

PR #3936, opened 2026-05-05
Update Retrieving converted images documentation to include getAvailableTemporaryUrl

Orrison:
Updates the documentation on the "Retrieving converted images" page to include details on the `getAvailableTemporaryUrl` method, which is not currently documented.


freekmurze:
Thanks!

PR #3938, opened 2026-05-20
Add opt-in conversions_disk_name config for default conversions disk

patrickweh:
## What

Adds a new optional `conversions_disk_name` setting to `config/media-library.php`. When set, conversions and responsive images are stored on that disk by default — without requiring `->storingConversionsOnDisk(...)` at every upload site or a `storeConversionsOnDisk(...)` declaration on every media collection.

```php
// config/media-library.php
'conversions_disk_name' => env('MEDIA_CONVERSIONS_DISK', null),
```

## Why

When originals live on a remote disk (S3, Hetzner Object Storage, …) it is almost always a mistake to also keep the thumbnails/responsive images there: every page that renders a list of thumbnails then turns into a fan-out of remote `GET`s, which is both slow and (for paid object storage) expensive. The recommended setup is "originals remote, derivatives local," but the framework currently has no central way to express that — every code path that adds media must remember to opt in via `storingConversionsOnDisk(...)`, and any path that uses `->addMedia(...)->toMediaCollection(...)` without that call silently inherits the originals' disk.

Apps that wire in third-party media (Nextcloud importers, mail attachments, IMAP sync, etc.) tend to skip the helper and write directly via `addMedia`, so the divergence is easy to miss until many thousands of conversions have ended up on the wrong disk.

## Behavior

The lookup order in `FileAdder::determineConversionsDiskName()` becomes:

1. Per-call `->storingConversionsOnDisk(...)` (unchanged, highest priority)
2. Per-collection `storeConversionsOnDisk(...)` (unchanged)
3. **New:** `config('media-library.conversions_disk_name')` when not empty
4. Fall back to the originals' disk (unchanged, current default)

## Backward compatibility

- Default value is `null`, which falls through to the original behavior — no change for existing installs.
- All higher-precedence mechanisms (per-call and per-collection overrides) keep winning.
- No migrations, no signature changes, no deprecations.

## Tests

The existing `tests/Feature/FileAdder/ConversionsDiskTest.php` is extended with four new cases covering: opt-in via config, the `null` fallback, per-call override precedence, and per-collection override precedence. The existing four cases keep passing unchanged.

```
Tests:    8 passed (21 assertions)
```

## Docs

Updated `docs/installation-setup.md` and `docs/advanced-usage/working-with-multiple-filesystems.md` with the new setting and a short rationale.

### DIFF
--- a/config/media-library.php
+++ b/config/media-library.php
@@ -35,6 +35,17 @@ return [
     'disk_name' => env('MEDIA_DISK', 'public'),
 
     /*
+     * The disk on which to store conversions (thumbnails, etc.) and responsive images
+     * when no disk is specified explicitly on the media collection or via
+     * `storingConversionsOnDisk()`. When left null, conversions are stored on the
+     * same disk as the original media — preserving previous behavior.
+     *
+     * This is useful when the originals live on a remote disk (e.g. S3) but the
+     * generated derivatives should stay local for faster access and lower egress.
+     */
+    'conversions_disk_name' => env('MEDIA_CONVERSIONS_DISK', null),
+
+    /*
      * The maximum file size of an item in bytes.
      * Adding a larger file will result in an exception.
      */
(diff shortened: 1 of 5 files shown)


freekmurze:
Thanks!

PR #3941, opened 2026-06-05
Add security note for addMediaFromUrl

freekmurze:
This adds a short security note to the `addMediaFromUrl` API docs.

We received a low quality CVE submission claiming `addMediaFromUrl` is a remote code execution vulnerability. It is not. The method fetches whatever is hosted at the given URL, exactly like `file_get_contents` or `curl`, and "RCE" only occurs if an application feeds it unvalidated user input, stores the result inside the web root, and configures the server to execute scripts from that directory. Those are application and deployment responsibilities, not defects in the package. The method already validates the URL scheme and supports MIME type allowlisting.

The note documents the defense in depth practices for applications that pass user supplied URLs:

* pass allowed MIME types to restrict accepted content
* validate the URL before use (SSRF)
* store media outside the web root or where script execution is disabled

Docs only, no behavior change.

### DIFF
--- a/docs/api/adding-files.md
+++ b/docs/api/adding-files.md
@@ -49,6 +49,18 @@ This method only accepts URLs that start with `http://` or `https://`
 public function addMediaFromUrl(string $url)
 ```
 
+#### A note on security
+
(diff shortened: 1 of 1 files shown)


freekmurze:
Superseded by the SSRF security note already on main. Closing this redundant version.

PR #3943, opened 2026-06-15
Attribute-based media configuration (new major version)

freekmurze:

This is the first PR toward a new major version of Media Library. The new major focuses on more ergonomic configuration and better control over what happens after you add a file. Breaking changes are on the table.

This PR ships the first headline feature (attribute-based configuration). A second feature (derivative completion callbacks) is in development in #3944 and previewed below. **Feedback and feature ideas are very welcome** (see the bottom of this description).

## Feature 1: Attribute-based configuration (in this PR)

Declare media collections and conversions with PHP attributes, as an alternative to (and coexisting with) `registerMediaCollections()` / `registerMediaConversions()`.

```php
use Spatie\Image\Enums\Fit;
use Spatie\MediaLibrary\Attributes\MediaCollection;
use Spatie\MediaLibrary\Attributes\MediaConversion;

#[MediaCollection(name: 'avatar', singleFile: true, fallbackUrl: '/default-avatar.png')]
#[MediaConversion(name: 'thumb', collections: ['avatar'], width: 150, height: 150, fit: Fit::Crop, format: 'webp')]
class User extends Model implements HasMedia
{
    use InteractsWithMedia;
}
```

### More examples

Both attributes are repeatable, so declare as many as you need.

A conversion with no `collections` applies to every collection on the model.

```php
#[MediaCollection(name: 'images')]
#[MediaCollection(name: 'banners')]
#[MediaConversion(name: 'optimized', width: 2000, format: 'webp', quality: 80)]
class Page extends Model implements HasMedia
{
    use InteractsWithMedia;
}
```

Collections can carry their own constraints, and conversions can opt into responsive images or run outside the queue.

```php
#[MediaCollection(name: 'documents', onlyKeepLatest: 10, acceptsMimeTypes: ['application/pdf'])]
#[MediaCollection(name: 'photos', disk: 'media', conversionsDisk: 'media', responsiveImages: true)]
#[MediaConversion(name: 'hero', collections: ['photos'], width: 1600, responsiveImages: true, queued: false)]
class Article extends Model implements HasMedia
{
    use InteractsWithMedia;
}
```

Attributes and methods combine. The static configuration lives in attributes, while the dynamic tail (the full spatie/image manipulation set, per-media conversions, `acceptsFile()` closures) stays in methods. A method declaration overrides a same-named attribute.

```php
#[MediaCollection(name: 'gallery')]
#[MediaConversion(name: 'thumb', collections: ['gallery'], width: 150, height: 150)]
class Product extends Model implements HasMedia
{
    use InteractsWithMedia;

    {
        $this->addMediaConversion('watermarked')
            ->watermark(storage_path('logo.png'));
    }
}
```

### How it works

- **Repeatable, class-level attributes.** `#[MediaCollection]` and `#[MediaConversion]` carry plain config data. A `MediaAttributeResolver` reads them via reflection (cached per model class) and turns them into the same builder objects the methods already produce.
- **Coexistence and precedence.** Attributes and the existing methods can be combined. Attributes resolve first, and a collection or conversion declared in a method overrides a same-named attribute one.
- **The dynamic tail stays in methods.** The attribute surface covers geometry, format, and the collection-level toggles (the common case).
- **Validation.** Clear exceptions for a conversion referencing an unknown collection and for duplicate collection names.
- **Argument surface.** `#[MediaCollection]`: `name`, `singleFile`, `onlyKeepLatest`, `acceptsMimeTypes`, `disk`, `conversionsDisk`, `fallbackUrl`, `fallbackPath`, `responsiveImages`. `#[MediaConversion]`: `name`, `collections`, `width`, `height`, `fit`, `format`, `quality`, `queued`, `responsiveImages`, `keepOriginalImageFormat`.
- Docs: `docs/basic-usage/configuring-with-attributes.md`. Full suite green.

## Feature 2: Derivative completion callbacks, `then()` / `catch()` (#3944)

Run code after a media item's conversions and responsive images finish generating, with failure handling. This mirrors the `then` / `catch` ergonomics of `spatie/laravel-pdf`, adapted to how Media Library generates derivatives.

```php

        // runs after the derivatives are generated
    })
        // runs if derivative generation fails
    })
```

Using `then()` runs the derivatives on the queue and fires the callback when they finish. With the `sync` queue driver it runs inline. `toMediaCollection()` still returns the `Media` synchronously. See #3944.

## Feedback welcome

The new major version is taking shape and nothing is locked in yet. If you have feature ideas, API feedback, or use cases the current design does not cover, please comment here or open a discussion. We would love your input on what the next version of Media Library should include.

### DIFF
--- a/docs/basic-usage/configuring-with-attributes.md
+++ b/docs/basic-usage/configuring-with-attributes.md
@@ -1,0 +1,80 @@
+---
+title: Configuring with attributes
+weight: 4
+---
+
+Media collections and conversions can be declared directly on your model using PHP attributes, instead of (or alongside) the `registerMediaCollections()` and `registerMediaConversions()` methods.
+
+```php
+use Spatie\Image\Enums\Fit;
+use Spatie\MediaLibrary\Attributes\MediaCollection;
+use Spatie\MediaLibrary\Attributes\MediaConversion;
+use Spatie\MediaLibrary\HasMedia;
+use Spatie\MediaLibrary\InteractsWithMedia;
+
+#[MediaCollection(name: 'avatar', singleFile: true, fallbackUrl: '/default-avatar.png')]
+#[MediaCollection(name: 'downloads')]
+#[MediaConversion(name: 'thumb', collections: ['avatar'], width: 150, height: 150, fit: Fit::Crop, format: 'webp')]
+#[MediaConversion(name: 'preview', width: 500)]
+class User extends Model implements HasMedia
+{
+    use InteractsWithMedia;
+}
+```
+
+Both attributes are repeatable, so you can declare as many collections and conversions as you need.
+
+## MediaCollection attribute
+
+The `#[MediaCollection]` attribute accepts the following arguments.
+
+| Argument | Type | Description |
+| --- | --- | --- |
+| `name` | `string` | The name of the collection. Required. |
+| `singleFile` | `bool` | Keep only the latest file in the collection. |
+| `onlyKeepLatest` | `?int` | Keep only the latest N files in the collection. Takes precedence over `singleFile`. |
+| `acceptsMimeTypes` | `array` | Restrict the collection to these mime types. |
+| `disk` | `?string` | The disk the collection stores its files on. |
+| `conversionsDisk` | `?string` | The disk the collection stores its conversions on. |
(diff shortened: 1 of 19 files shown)


freekmurze:
Superseded by #3945, which combines this work onto a single next-major-version branch (`next-major`).

PR #3944, opened 2026-06-15
Run code after conversions with then() / catch() (new major version)

freekmurze:

This is the companion PR to #3943, the second headline feature of the new major version of Media Library. **Feedback and feature ideas are very welcome** (see the bottom).

## Run code after a media item's derivatives are generated

You can now register a callback that runs after a media item's derivatives (its conversions and responsive images) have been generated, plus a callback for failures. This mirrors the `then` / `catch` ergonomics of `spatie/laravel-pdf`, adapted to how Media Library generates derivatives.

```php
use Throwable;

        // Runs after the derivatives have been generated.
    })
        // Runs if derivative generation fails.
    })
```

The `then` callback receives the `Media`. The `catch` callback receives the `Throwable`.

## How it works

- `then()` / `catch()` store the callbacks as `SerializableClosure`s on the `FileAdder`.
- When set, all of the media's derivative work (the conversions job, and the responsive images job when enabled) plus a final callback job are dispatched as one `Bus::chain(...)->catch(...)`. The `then` closure runs in the final job, the chain `catch` runs the `catch` closure.
- A `Bus::chain` is used rather than a batch, so no `job_batches` migration is required.

## Behavior

- **Derivatives run on the queue.** Using `then()` runs the derivatives on the queue and fires the callback when they finish, even for conversions that would otherwise run inline. With the `sync` queue driver they run inline and the callback still fires.
- **`toMediaCollection()` still returns the `Media` synchronously.** The callbacks are additive.
- **No derivatives:** if a media item produces no conversions and no responsive images, `then` fires immediately with the media.
- **Failure handling:** on a real queue, a failing derivative is handled in the worker and your `catch` runs there. On the `sync` driver, `catch` still runs, but the underlying exception also propagates out of `toMediaCollection()` (this is how the sync queue behaves).
- **Serialization constraint:** callbacks are serialized to run on the queue, so they cannot capture non serializable state (resolve services inside the callback).

## Notes

- Adds `laravel/serializable-closure` to require (already present transitively via the framework).
- Docs added at `docs/advanced-usage/running-code-after-conversions.md`.
- Full suite green.

## Feedback welcome

The new major version is taking shape and nothing is locked in yet. If you have feature ideas, API feedback, or use cases this design does not cover, please comment here or open a discussion. We would love your input on the next version of Media Library.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,6 +33,7 @@
         "illuminate/database": "^10.2|^11.0|^12.0|^13.0",
         "illuminate/pipeline": "^10.2|^11.0|^12.0|^13.0",
         "illuminate/support": "^10.2|^11.0|^12.0|^13.0",
+        "laravel/serializable-closure": "^1.3|^2.0",
         "maennchen/zipstream-php": "^3.1",
         "spatie/image": "^3.3.2",
         "spatie/laravel-package-tools": "^1.16.1",
--- a/docs/advanced-usage/running-code-after-conversions.md
+++ b/docs/advanced-usage/running-code-after-conversions.md
@@ -33,9 +33,11 @@ With the `sync` queue driver the derivatives run inline, and the callback still 
 
 If a media item has no conversions and no responsive images, the `then` callback fires immediately with the media.
 
-## Failure handling on the sync driver
+## Failure handling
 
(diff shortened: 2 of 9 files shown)


freekmurze:
Superseded by #3945, which combines this work onto a single next-major-version branch (`next-major`).

PR #3945, opened 2026-06-15
The next major version of Media Library

freekmurze:

This PR collects the work for the next major version of Media Library on a single branch. It focuses on more ergonomic configuration and better control over what happens after you add a file. Breaking changes are on the table. It supersedes #3943 and #3944, which are combined here.

**Feedback and feature ideas are very welcome** (see the bottom). Nothing is locked in yet.

## Feature 1: Attribute-based configuration

Declare media collections and conversions with PHP attributes, as an alternative to (and coexisting with) `registerMediaCollections()` / `registerMediaConversions()`.

```php
use Spatie\Image\Enums\Fit;
use Spatie\MediaLibrary\Attributes\MediaCollection;
use Spatie\MediaLibrary\Attributes\MediaConversion;

#[MediaCollection(name: 'avatar', singleFile: true, fallbackUrl: '/default-avatar.png')]
#[MediaConversion(name: 'thumb', collections: ['avatar'], width: 150, height: 150, fit: Fit::Crop, format: 'webp')]
class User extends Model implements HasMedia
{
    use InteractsWithMedia;
}
```

Both attributes are repeatable. A conversion with no `collections` applies to every collection on the model.

```php
#[MediaCollection(name: 'documents', onlyKeepLatest: 10, acceptsMimeTypes: ['application/pdf'])]
#[MediaCollection(name: 'photos', disk: 'media', conversionsDisk: 'media', responsiveImages: true)]
#[MediaConversion(name: 'hero', collections: ['photos'], width: 1600, responsiveImages: true, queued: false)]
class Article extends Model implements HasMedia
{
    use InteractsWithMedia;
}
```

Attributes and methods combine. The static configuration lives in attributes, the dynamic tail (the full spatie/image manipulation set, per-media conversions, `acceptsFile()` closures) stays in methods, and a method declaration overrides a same-named attribute.

```php
#[MediaCollection(name: 'gallery')]
#[MediaConversion(name: 'thumb', collections: ['gallery'], width: 150, height: 150)]
class Product extends Model implements HasMedia
{
    use InteractsWithMedia;

    {
        $this->addMediaConversion('watermarked')
            ->watermark(storage_path('logo.png'));
    }
}
```

- A `MediaAttributeResolver` reads the attributes via reflection (cached per model class) and turns them into the same builder objects the methods already produce.
- Attributes resolve first, then the methods run on top.
- Clear exceptions for a conversion referencing an unknown collection and for duplicate collection names.
- Argument surface. `#[MediaCollection]`: `name`, `singleFile`, `onlyKeepLatest`, `acceptsMimeTypes`, `disk`, `conversionsDisk`, `fallbackUrl`, `fallbackPath`, `responsiveImages`. `#[MediaConversion]`: `name`, `collections`, `width`, `height`, `fit`, `format`, `quality`, `queued`, `responsiveImages`, `keepOriginalImageFormat`.
- Docs: `docs/basic-usage/configuring-with-attributes.md`.

## Feature 2: Run code after conversions with `then()` / `catch()`

Run code after a media item's conversions and responsive images finish generating, with failure handling. This mirrors the `then` / `catch` ergonomics of `spatie/laravel-pdf`, adapted to how Media Library generates derivatives.

```php
use Throwable;

        // runs after the derivatives are generated
    })
        // runs if derivative generation fails
    })
```

- The `then` callback receives the `Media`, the `catch` callback receives the `Throwable`.
- Using `then()` runs the derivatives on the queue and fires the callback when they finish. With the `sync` queue driver it runs inline. `toMediaCollection()` still returns the `Media` synchronously.
- The derivatives run inside a single orchestrator job with a plain `try/catch`, so `then` and `catch` behave deterministically across Laravel versions and queue drivers.
- If a media item produces no derivatives, `then` fires immediately.
- Callbacks are serialized to run on the queue, so they cannot capture non serializable state.
- Adds `laravel/serializable-closure` to require. Docs: `docs/advanced-usage/running-code-after-conversions.md`.

## Feature 3: Enum collection names

Anywhere a collection name is accepted you can now pass a string-backed enum instead of a string. This covers the attributes, `addMediaCollection()`, `toMediaCollection()` (and variants), the `getMedia()` / `getFirstMedia()` / URL / path family, `clearMediaCollection()`, and `performOnCollections()`. The enum is resolved to its value, so existing string usage is unchanged.

```php
enum MediaCollectionName: string
{
    case Avatar = 'avatar';
}

#[MediaCollection(name: MediaCollectionName::Avatar, singleFile: true)]
#[MediaConversion(name: 'thumb', collections: [MediaCollectionName::Avatar], width: 150, height: 150)]
class User extends Model implements HasMedia
{
    use InteractsWithMedia;
}
```

```php
$user->addMedia($request->file('avatar'))->toMediaCollection(MediaCollectionName::Avatar);

$user->getFirstMediaUrl(MediaCollectionName::Avatar, 'thumb');
```

Implemented via a small `Spatie\MediaLibrary\Support\CollectionName` helper. Requested in discussion #3833. Docs: `docs/basic-usage/configuring-with-attributes.md`.

## Status

Full suite green (447 passed), PHPStan clean.

## Feedback welcome

If you have feature ideas, API feedback, or use cases the current design does not cover, please comment here or open a discussion. We would love your input on what the next major version of Media Library should include.


rcerljenko:
Hi there! It's maybe worth mentioning that Laravel now has integrated image processing. Maybe the next version can make use of it and drop some dependencies along the way => https://laravel-news.com/a-practical-guide-to-laravels-first-party-image-processing

PR #3948, opened 2026-06-29
Fix media-library:clean ignoring custom PathGenerator for orphaned directories

nikolauspschuetz:
## Fix: `media-library:clean` ignores a custom `PathGenerator` when removing orphaned directories

Fixes the bug reported in #3809.

### Problem

`media-library:clean`'s orphaned-directory cleanup (`deleteOrphanedDirectories()`) assumed the **default** `{prefix}/{mediaId}` layout. It listed the top-level directories on the disk, kept only the ones whose names were numeric (`->filter(fn ($d) => is_numeric($d))`), and compared them against the set of media IDs.

With a custom `PathGenerator` (UUID-based, hash-based, nested, etc.) the on-disk directory names are **not** numeric media IDs, so:

- every directory was filtered out by `is_numeric()`, and
- orphaned-directory cleanup silently became a no-op.

This is inconsistent with the sibling halves of the same command: `deleteConversionFilesForDeprecatedConversions()` already resolves paths via `PathGeneratorFactory::create($media)->getPathForConversions($media)`. Only the orphaned-directory half was still ID-based.

### Root cause

`src/MediaCollections/Commands/CleanCommand.php` — `deleteOrphanedDirectories()`:

- `$mediaIdSet = $this->mediaRepository->allIds()->flip();` — assumes dirs are named with media IDs.
- `->filter(fn (string $directory) => is_numeric($directory))` — drops every non-numeric (custom-layout) directory.

### Fix

Derive the set of *in-use* top-level directories from the live media through the configured path generator(s), instead of assuming numeric IDs:

```php
$usedDirectories = $this->mediaRepository->all()
    ->map(fn (Media $media) => $this->getTopLevelDirectory(
        PathGeneratorFactory::create($media)->getPath($media), $prefix
    ))
    ->filter()
    ->collect()
    ->unique()
    ->flip();
```

On-disk top-level directories that are **not** in that set are treated as orphaned and removed. This mirrors how the rest of the command resolves paths and makes orphaned-directory cleanup correct for any path generator — the default numeric layout still works (covered by a regression test). `prefix`, `--dry-run`, and `--rate-limit` behaviour are preserved.

### Behavioural note

The previous `is_numeric()` filter also acted as an incidental guard against deleting *foreign* directories on a shared disk. The new logic deletes any top-level directory under the configured prefix that no media maps to. This matches the documented expectation that media-library owns its disk/prefix (and is consistent with the conversion-cleanup half, which already deletes per resolved path). Using a dedicated disk or a `media-library.prefix` remains the recommended setup.

### Tests

Added to `tests/Conversions/Commands/CleanCommandTest.php`:

- **`it('can clean orphaned directories when using a custom path generator')`** — registers `CustomPathGenerator` (hash-based, non-numeric layout), creates a live media plus a leftover orphan directory in the custom layout, runs `media-library:clean`, asserts the orphan dir is removed and the live media dir is kept. **Fails on `main`** (orphan never matched, so never deleted), passes with the fix.
- **`it('keeps cleaning orphaned directories for the default numeric path generator')`** — regression guard for the default `{id}` layout.

Test results (PHP 8.3, `vendor/bin/pest tests/Conversions/Commands/CleanCommandTest.php`):

Tests:    19 passed (121 assertions)
```

Verified test-first: with the source fix reverted (test only), `it can clean orphaned directories when using a custom path generator` **fails** ("orphan directory … does not exist" assertion fails — the dir is never removed); with the fix applied it passes. Full suite: 414 passed / 23 skipped (the only failures in this sandbox are pre-existing and unrelated — they require the `imagick` PHP extension, which isn't installed here). `phpstan analyse` on the changed file: no errors.

### AI assistance

This change was developed with AI assistance (Claude Code, Anthropic, Opus 4.x). The AI did the codebase investigation, drafted the fix and the failing-first tests, and ran the suite. I (the contributor) set the direction, verified the root cause against the live source on `main` and `next-major`, reviewed the diff and behavioural trade-off, and confirmed the test fails without the fix and passes with it. The commit carries an `Assisted-by:` trailer.

Note on the CHANGELOG: not hand-edited, since this repo auto-generates `CHANGELOG.md` on release via `update-changelog.yml` (changelog-updater-action).

cc @freekmurze — this is the orphaned-directory half of #3809; on inspection the same `is_numeric`/ID-based code is still present on both `main` and `next-major`, so it didn't get the `PathGeneratorFactory` treatment that the conversion-cleanup half has. Happy to adjust the approach (e.g. re-add a guard) if you'd prefer.


freekmurze:
Thanks for the clear report and fix, @nikolauspschuetz! This is merged and released in 11.23.7.

PR #3950, opened 2026-06-30
Fix double URL encoding for getTemporaryUrl() with non-ASCII filenames on local disk

### LINKED ISSUE #3949: [Bug] Double URL encoding causes 404 for getTemporaryUrl() with non-ASCII filenames on local disk
## Bug Description
When using `$media->getTemporaryUrl()` on a `local` disk, files with non-ASCII characters (e.g., Cyrillic, Arabic, special symbols) return a 404 error when accessed. This occurs because the file path undergoes **double URL encoding**.

### The Root Cause
1. `DefaultUrlGenerator::getTemporaryUrl()` calls `$this->getUrlEncodedPathRelativeToRoot()`.
2. For the `local` driver, `getUrlEncodedPathRelativeToRoot()` applies `rawurlencode()` to the path segments.
3. This already-encoded path is then passed to Laravel's `$disk->temporaryUrl()`.
4. However, Laravel's built-in `LocalFilesystemAdapter::temporaryUrl()` *also* applies `rawurlencode($path)` internally.
5. The result is double-encoded (e.g., `%D0%9F` becomes `%25D0%259F`), causing Laravel's route handler to fail `Storage::exists()` after decoding the route parameter only once.

## Steps to Reproduce
1. Upload a file with a Cyrillic name, e.g., `test-document-кирилиця.pdf` to a collection using the `local` disk.
2. Generate a temporary URL: `$url = $media->getTemporaryUrl(now()->addMinutes(5));`
3. Attempt to fetch the URL - it returns `404 Not Found`.

## Minimal Reproducer / Proof
Here is a test that proves pure Laravel works correctly on its own, but the Spatie pipeline results in a double-encoded parameter.

```php
use Illuminate\Http\Request;
use Illuminate\Support\Facades\Route;
use Illuminate\Support\Facades\Storage;

it('proves spatie double encodes local temporary urls compared to pure laravel', function () {
    $filename = 'test-document-кирилиця.pdf';
    
    // 1. PURE LARAVEL (Works correctly)
    Storage::disk('local')->put($filename, 'content');
    $laravelUrl = Storage::disk('local')->temporaryUrl($filename, now()->addMinutes(5));
    
    // Simulate Laravel's route resolution
    $laravelRouteParam = Route::getRoutes()
        ->match(Request::create($laravelUrl))
        ->parameter('path');
        
    // Laravel correctly decodes back to the raw filename
    expect($laravelRouteParam)->toBe($filename); 

    // 2. SPATIE MEDIA LIBRARY (Fails due to double encoding)
    $media = YourModel::first()->addMediaFromString('content')                                                                                                                                                                                                                                                       
      ->usingFileName($filename)                                                                                                                                                                                                                                                                                   
      ->toMediaCollection('default', 'local');
        
    $spatieUrl = $media->getTemporaryUrl(now()->addMinutes(5));
    
    $spatieRouteParam = Route::getRoutes()
        ->match(Request::create($spatieUrl))
        ->parameter('path');
        
    // Fails! $spatieRouteParam is '%D0%BA%D0...pdf' instead of 'test-document-кирилиця.pdf'
    // Because it was encoded twice, it was only decoded once by the router.
    expect($spatieRouteParam)->not->toBe($filename); 
});
```

## Expected Behavior
The path should be encoded exactly once. Since Laravel handles encoding internally for `local` disk temporary URLs, Spatie should pass the raw path to `$disk->temporaryUrl()`.

## Environment
- **Laravel:** 11.x / 12.x / 13.x
- **Media Library:** 11.x

maeandrew:
# Fix double URL encoding for temporary URLs on local disk

## Description
Fixes #3949

### The Problem
Calling `$media->getTemporaryUrl()` for files with non-ASCII characters (like Cyrillic or special symbols) on a `local` disk results in a 404 error. 

This occurs due to double URL-encoding: 
1. `DefaultUrlGenerator::getTemporaryUrl()` calls `$this->getUrlEncodedPathRelativeToRoot()`, which encodes the path for `local` drivers.
2. This encoded path is passed to `$disk->temporaryUrl()`.
3. Laravel's built-in `LocalFilesystemAdapter::temporaryUrl()` encodes the path a second time internally (`rawurlencode($path)`).

Because the path is encoded twice, Laravel's route parser only decodes it once. `Storage::exists()` is then called with a still-encoded string instead of the raw filename, resulting in a `404 Not Found`.

### The Fix
Changed `DefaultUrlGenerator::getTemporaryUrl()` to pass `$this->getPathRelativeToRoot()` instead of `$this->getUrlEncodedPathRelativeToRoot()`.

**File changed:** `src/Support/UrlGenerator/DefaultUrlGenerator.php`
```diff
    public function getTemporaryUrl(DateTimeInterface $expiration, array $options = []): string
    {
-       return $this->getDisk()->temporaryUrl($this->getUrlEncodedPathRelativeToRoot(), $expiration, $options);
+       return $this->getDisk()->temporaryUrl($this->getPathRelativeToRoot(), $expiration, $options);
    }
```

### Why this is safe for all drivers
This change only affects the `local` driver. 

If we look at `getUrlEncodedPathRelativeToRoot()`, we can see it already skips encoding for non-local drivers:
```php
    {
        $driver = config("filesystems.disks.{$this->getDiskName()}.driver");

        // 👇 For S3, MinIO, etc. it ALREADY returns the raw path!
        if ($driver !== 'local') {
            return $this->getPathRelativeToRoot();
        }
        
        // ... local driver encoding logic ...
    }
```
Because non-local drivers (like S3) were already receiving the raw `getPathRelativeToRoot()` via the early return, their behavior remains 100% unchanged. This PR simply aligns the `local` driver to also pass the raw path, which Laravel's `LocalFilesystemAdapter` explicitly expects.

### DIFF
--- a/src/Support/UrlGenerator/DefaultUrlGenerator.php
+++ b/src/Support/UrlGenerator/DefaultUrlGenerator.php
@@ -16,7 +16,7 @@ class DefaultUrlGenerator extends BaseUrlGenerator
 
     public function getTemporaryUrl(DateTimeInterface $expiration, array $options = []): string
     {
-        return $this->getDisk()->temporaryUrl($this->getUrlEncodedPathRelativeToRoot(), $expiration, $options);
+        return $this->getDisk()->temporaryUrl($this->getPathRelativeToRoot(), $expiration, $options);
     }
 
     public function getBaseMediaDirectoryUrl(): string
--- a/tests/Feature/Media/GetUrlTest.php
+++ b/tests/Feature/Media/GetUrlTest.php
@@ -1,6 +1,7 @@
 <?php
 
 use Carbon\Carbon;
+use Illuminate\Support\Facades\Storage;
 use Spatie\MediaLibrary\MediaCollections\Exceptions\InvalidConversion;
 
 it('can get an url of an original item', function () {
@@ -45,4 +46,22 @@ it('throws an exception when trying to get a temporary url on local disk', funct
     $this->expectException(RuntimeException::class);
 
     $media->getTemporaryUrl(Carbon::now()->addMinutes(5));
+});
+
+it('passes the raw path to the disk when generating a temporary url on local disk', function () {
+    $media = $this->testModel
+        ->addMedia($this->getTestJpg())
+        ->usingFileName('tést.jpg')
+        ->toMediaCollection();
+
+    $disk = Mockery::mock(Storage::disk('public'))->makePartial();
+    $disk->shouldReceive('temporaryUrl')
+        ->once()
+        ->withArgs(fn (string $path) => $path === "{$media->id}/tést.jpg")
+        ->andReturn('https://example.com/temporary-url');
+
+    Storage::set('public', $disk);
+
+    expect($media->getTemporaryUrl(Carbon::now()->addMinutes(5)))
+        ->toEqual('https://example.com/temporary-url');
 });


freekmurze:
Thanks for the fix, @maeandrew! I added a regression test and shipped this in [11.23.2](https://github.com/spatie/laravel-medialibrary/releases/tag/11.23.2).

PR #3951, opened 2026-07-09
test: remove real sleep(1) from regenerate command tests

nikolauspschuetz:
Four tests in `RegenerateCommandTest` call `sleep(1)` so that a regenerated conversion ends up with a strictly greater `filemtime()` than the original. That is needed only because `filemtime()` has one-second granularity — regenerating within the same wall-clock second produces an equal mtime and fails `toBeGreaterThan`.

Rather than waiting a real second each time, this backdates the original conversion before capturing its mtime:

```php
touch($derivedImage, time() - 5);
$createdAt = filemtime($derivedImage);
```

so the freshly regenerated file is deterministically newer without any sleeping.

**No behaviour change** — the same assertions run, and the "only-missing" tests still assert the untouched conversion keeps its original mtime (`toBe`). Only the missing/regenerated file is backdated.

Verified locally with `vendor/bin/pest tests/Conversions/Commands/RegenerateCommandTest.php`:
- before: **15 passed, 56 assertions — 7.09s**
- after: **15 passed, 56 assertions — 3.24s** (~3.9s faster)

### AI assistance

This change was drafted with Claude Code (Anthropic, Opus 4.x): identifying the fixed-sleep pattern, replacing it with `touch()`-based backdating, and verifying the suite locally before/after. I reviewed the diff and the timing results. Disclosed per my disclose-by-default practice; also noted in the commit trailer (`Assisted-by:`).

### DIFF
--- a/tests/Conversions/Commands/RegenerateCommandTest.php
+++ b/tests/Conversions/Commands/RegenerateCommandTest.php
@@ -8,13 +8,14 @@ it('can regenerate all files', function () {
     $media = $this->testModelWithConversion->addMedia($this->getTestFilesDirectory('test.jpg'))->toMediaCollection('images');
 
     $derivedImage = $this->getMediaDirectory("{$media->id}/conversions/test-thumb.jpg");
+    // Backdate the existing conversion so the regenerated file is guaranteed a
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks!

PR #3958, opened 2026-08-01
Fix media-library:clean when conversions live on a separate disk

### LINKED ISSUE #3957: media-library:clean ignores the conversions disk
When conversions are stored on a disk other than the original (`storingConversionsOnDisk()` or `media-library.conversions_disk_name`), `media-library:clean` operates on the wrong disk in two places.

**Deprecated conversion files are never cleaned**

`CleanCommand::deleteConversionFilesForDeprecatedConversions()` lists and deletes conversion files on `$media->disk`:

```php
$currentFilePaths = $this->fileSystem->disk($media->disk)->files($conversionPath);
// ...
$this->fileSystem->disk($media->disk)->delete($currentFilePath);
```

The files live on `$media->conversions_disk`, so the listing comes back empty and deprecated conversions accumulate forever. If the original disk happens to hold files under the same conversion path, they are the ones considered for deletion instead.

**Orphaned directories are only swept on one disk**

`CleanCommand::deleteOrphanedDirectories()` resolves a single disk:

```php
$diskName = $this->argument('disk') ?: config('media-library.disk_name');
```

Directories left behind on the conversions disk after media rows are deleted are never removed.

**Reproducing**

1. Attach media with `->storingConversionsOnDisk('secondMediaDisk')`.
2. Drop a stale file into the conversions directory on `secondMediaDisk`.
3. Run `media-library:clean` — the stale file is still there.
4. Delete the media row, run `media-library:clean` again — the directory on `secondMediaDisk` is still there.

Both reproduce on `main`.

PR with a fix and tests: #3958

AdzerKI:
Fixes #3957.

When conversions are stored on a disk other than the original, `media-library:clean` worked on the wrong disk in two places.

**Deprecated conversions**

`deleteConversionFilesForDeprecatedConversions()` listed and deleted conversion files on `$media->disk`. They live on `$media->conversions_disk`, so the listing came back empty and deprecated conversions were never cleaned. It now resolves `$media->conversions_disk ?: $media->disk` and uses that disk for both the listing and the delete, so the original disk is left untouched.

**Orphaned directories**

`deleteOrphanedDirectories()` only ever swept one disk, so directories left on the conversions disk survived. It now sweeps every disk actually referenced by media rows (new `MediaRepository::allDiskNames()`, reading both `disk` and `conversions_disk`) plus `media-library.disk_name` and `media-library.conversions_disk_name`. Each disk is validated before the sweep, so a missing disk still raises `DiskDoesNotExist`. Passing the `disk` argument keeps the previous behaviour and restricts the sweep to that one disk. The info line now names the disk the directory was found on.

**Tests**

Four cases in `CleanCommandTest`: a deprecated conversion on a separate conversions disk is removed; a stray file on the original disk is not touched while doing so; an orphaned directory on the conversions disk is removed while a live one is kept; the `disk` argument still limits the sweep to the given disk.

The first and third fail on `main` and pass with this change. Full suite is green.

### DIFF
--- a/src/MediaCollections/Commands/CleanCommand.php
+++ b/src/MediaCollections/Commands/CleanCommand.php
@@ -146,14 +146,15 @@ class CleanCommand extends Command
         $conversionFilePaths = ConversionCollection::createForMedia($media)->getConversionsFiles($media->collection_name);
 
         $conversionPath = PathGeneratorFactory::create($media)->getPathForConversions($media);
-        $currentFilePaths = $this->fileSystem->disk($media->disk)->files($conversionPath);
+        $conversionsDisk = $media->conversions_disk ?: $media->disk;
+        $currentFilePaths = $this->fileSystem->disk($conversionsDisk)->files($conversionPath);
 
         collect($currentFilePaths)
             ->reject(fn (string $currentFilePath) => $conversionFilePaths->contains(basename($currentFilePath)))
             ->reject(fn (string $currentFilePath) => $media->file_name === basename($currentFilePath))
-            ->each(function (string $currentFilePath) use ($media) {
+            ->each(function (string $currentFilePath) use ($media, $conversionsDisk) {
                 if (! $this->isDryRun) {
-                    $this->fileSystem->disk($media->disk)->delete($currentFilePath);
+                    $this->fileSystem->disk($conversionsDisk)->delete($currentFilePath);
 
                     $this->markConversionAsRemoved($media, $currentFilePath);
                 }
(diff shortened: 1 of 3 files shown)


freekmurze:
Thank you @AdzerKI! Great issue write-up and a clean fix with proper reproducing tests.

I pushed two follow-ups before merging:

- `markConversionAsRemoved()` used `Conversion::create($name)` to compute the expected file name, which builds a *default* conversion. For a conversion registered with `->format('webp')` that resolved to `test-thumb.jpg`, so a live conversion could still lose its generated flag. It now looks up the registered conversion from the `ConversionCollection`. Added a regression test for the custom-format case.
- The "will not touch the original disk" test placed the stray file outside the conversions path, so it passed on `main` too. Moved it into `{id}/conversions` so it actually covers the bug you described.

Released as [11.23.4](https://github.com/spatie/laravel-medialibrary/releases/tag/11.23.4).

PR #3960, opened 2026-08-06
docs: fix HasMedia interface import in custom image generator

latent-9:
The custom image generator example imports the `HasMedia` interface as `Spatie\MediaLibrary\Interfaces\HasMedia`, but there is no `Interfaces` namespace. The interface is `Spatie\MediaLibrary\HasMedia` (`src/HasMedia.php`), which is how it is imported throughout the rest of the docs. The example's `class News extends Model implements HasMedia` needs this corrected import to run.


freekmurze:
Thanks!

PR #3961, opened 2026-08-07
Decode remote url path before stripping directory in addMediaFromUrl

bunlongheng:
addMediaFromUrl in src/InteractsWithMedia.php runs basename() on the raw url path and only urldecodes afterwards. that ordering defeats basename, since percent-encoded separators survive the strip and only turn into real slashes once decoded. a url like https://host/uploads%2F..%2F..%2Fevil.jpg comes out as uploads/../../evil.jpg instead of just evil.jpg, so the directory portion basename is supposed to remove leaks into the derived filename.

the default sanitizer still swaps out the slashes, so this is mostly hardening rather than a break under the shipped config, but anyone using a custom sanitizer via sanitizingFileName() that keeps separators would inherit the traversal. decoding first, then calling basename, gives you the last path segment like it was meant to.

added a test covering the encoded-separator case.

### DIFF
--- a/src/InteractsWithMedia.php
+++ b/src/InteractsWithMedia.php
@@ -159,8 +159,7 @@ trait InteractsWithMedia
         $temporaryFile = (new $downloader)->getTempFile($url);
         $this->guardAgainstInvalidMimeType($temporaryFile, $allowedMimeTypes);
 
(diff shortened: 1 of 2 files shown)


freekmurze:
Thank you! Released in 11.23.5.

PR #3964, opened 2026-09-14
fix: lowercase extension check in keepOriginalImageFormat

### LINKED ISSUE #3963: keepOriginalImageFormat() ignores uppercase file extensions, saves JPEG bytes with a PNG filename
Found this while checking conversion output for files uploaded with uppercase extensions..

`keepOriginalImageFormat()` set on a conversion, uploading test.PNG produces a conversion file still named .PNG, but the actual image content is JPEG.

The check in `PerformManipulationsAction::execute()` is `in_array($media->extension, $supportedFormats)`, and `$supportedFormats` is all lowercase, so an uppercase extension fails the check and the conversion falls through to the default `->format('jpg')` . Meanwhile the file naming logic `(Conversion::getConversionFile())` derives the extension separately and keeps the original casing, so you end up with a filename and its actual content disagreeing, which is a quiet form of data corruption.

BTW: Lowercase extensions are working fine, this only affects uppercase ones.

moemadeldin:
Fixes #3963


freekmurze:
Thanks for tracking this down and fixing it, @moemadeldin! Nice catch on the filename and content getting out of sync, and the test that checks the actual mime type is a good touch. This is now released in [11.23.8](https://github.com/spatie/laravel-medialibrary/releases/tag/11.23.8).

